PR #17, opened 2016-08-16
Dependency integration in the repo

numist:
Getting started took a lot longer than expected in part thanks to the lack of a reasonable package management system in the Arduino ecosystem.

An easy fix to that is to move the firmware sources a level deeper in the repo (where they can be kept pure, which will please `arduino-builder`), and include the dependencies as submodules alongside the firmware.

In the future, there could be some benefit to using /libraries as a place where firmware code could be factored as well, such as `Storage` and `LEDControl`. This would keep the core sketch code compact and accessible to people hacking on their keyboards.


obra:
Thanks. Applied. (With an added commit adding minimal doc to the README)

PR #131, opened 2017-05-02
add header with rgb struct definition

wez:
refs: https://github.com/keyboardio/Kaleidoscope-Focus/issues/3

### DIFF
--- a/src/cRGB.h
+++ b/src/cRGB.h
@@ -1,0 +1,6 @@
+#pragma once
+struct cRGB {
+    uint8_t b;
+    uint8_t g;
+    uint8_t r;
+};


algernon:
There's one big issue with this, and why I suggested moving the struct to the hardware library: the order of bytes is not always `BGR`. The [Shortcut](https://github.com/Shortcutgg/Kaleidoscope-Hardware-Shortcut/blob/a6d07d8b3f5b3f5bec690e521ff3685f952fd6f3/src/Kaleidoscope-Hardware-Shortcut.h#L29-L33) for example uses `GRB`.

If the struct is always `BGR`, we'd have to do a lot of massaging in the hardware libraries of those keyboards that use another order.

What would make sense, is a small no-op library.

wez:
Ah, ok, so new game plan is to add a Kaleidoscope-ColorFormat library that defines this struct, and provide an #ifdef that allows selecting GRB vs BGR as the internal ordering?   The hardware header could define an appropriate value such that the color format header is evaluated the right way

obra:
Out of curiosity, how does FastLED deal? They're often considered to be the best practice.

algernon:
I'd leave the struct in the hardware headers (or KeyboardioScanner in its case), and provide a `Kaleidoscope-Dummy-CRGB` library, that has it in any order, and provides a few functions that are expected from the hardware libraries, that do nothing.

Hardware libs that have RGBs can then implement their own thing (and we can lift out common stuff once there is a need to do so), and libs that don't, can pull in the dummy cRGB lib.

algernon:
It would be lovely to use FastLED, too. I gave it a try when porting to Shortcut, but didn't manage to make it work. (It compiled, but LEDs never lit up)

FastLED uses some templating magic to decide the byte order - that's a lot better than ifdefs... if we don't go the FastLED route, then a library with a bit of template sugar coating would be neat. Better than both the dummy lib idea, or the ifdeffing, in my opinion.

PR #132, opened 2017-05-06
EventDispatcher interface

wez:
The goal is to be able to support:

* PluggableUSB hosts
* Other USB implementations (such as Teensy 3.x)
* BLE on nRF52
* BLE via Adafruit SPIFRIEND boards
* Be able to direct keypresses to some combination of the above,
  depending on the policy in the firmware.  For example, the
  most desirable default would be to use USB when available,
  and fall back to BLE if available.

To that end, this diff adds:

* `PtrArray` - a reusable implementation of the same approach
  used for recording hooks, this is a bit like std::array or
  std::vector, but with the append/replace semantics from
  the existing hook registration.  A later commit could
  switch the hooks over to PtrArray.

* `EventDispatcher` - a virtual base class with methods
  relating to pressing different kinds of keys.  There
  is a global registry of `EventDispatcher`'s maintained
  using a `PtrArray`.  Each `EventDispatcher` registers
  itself during construction.

`PtrArray::apply` allows walking over the list of dispatchers and
applying a lambda.  If you're worried about the overhead of this, the
compiler should optimize this to the equivalent for loop and direct
function call.

The `EventDispatcher` has the concept of a `connectionMask` which
is essentially a bitfield that indicates the combination of USB
and BLE connections that are available.   Each dispatcher will
report its nature via the `queryConnectionTypes` method.

The loop driver then has the option to apply the policy mentioned
above.  I haven't provided the code for that portion, because
I haven't actually implemented that yet, and it can be done
in a separate smaller commit.  The key concept here is that
we have the option to transform the *available* `connectionMask`
into a *desired* `connectionMask`.

Each of the key handling related methods is passed the *desired*
`connectionMask` along with the information about the key press.
The implementation will NOP if the desired `connectionMask` mask
doesn't match its nature.

On Registration:

The intent is that the hardware class will simply embed an
instance of the appropriate `EventDispatcher`(s) as member fields.
This will implicitly register them.

The `HIDEventDispatcher.h` header contains an implementation
with the same semantics as we had prior to this diff, so most
USB based boards will just need to add:

```
   HIDEventDispatcher dispatcher_;
```

as a field and be done.

In an attempt to minimize the number of changes in the other repos,
I've added a little bit of preprocessor stuff to allow setting
up a default HIDEventDispatcher instance.  The protocol for this
is that, until the hardware classes are updated, if the hardware
header doesn't define `HARDWARE_EVENT_DISPATCHER`, the
`DefaultEventDispatcher.cpp` class will instantiate one.  The
`Kaleidoscope::setup` method will explicitly reference that class
to ensure that the linker pulls it in.

Fan out to other repos:

The Macro and DualKey repos each need a small change to use the
new `send_keyboard_report` functon instead of trying to call
`Keyboard.sendReport` directly.

Mouse Keys:

We can either expand `EventDispatcher` to incorporate mouse
related methods, or we can add a similar `MouseDispatcher` class
that works in basically the same way.  Thoughts?


algernon:
> A later commit could switch the hooks over to PtrArray.

Oh, that would be lovely!

The rest of the changes - at least their description, I haven't looked at the code yet - sound good too.

The only thing I may change, is the `HARDWARE_EVENT_DISPATCHER`: I'd just update the small number of hardware boards at this point. None of them are in end-user hands yet, so breaking them for a night or so is in my opinion acceptable. OTOH, as the compat thing is already written, there is no real harm having it for the time being, either. No strong feelings on this, mind you.

wez:
For an example of a teensy hardware implementation that uses this interface, take a look at the files here:

https://github.com/wez/KaleidoscopeKeyboards/tree/master/one-key-teensy/hw/src

obra:
My gut feeling is to that we want a single HID Event Dispatcher, rather than adding equivalent dispatchers for mouse, joystick, etc

obra:
With the new kaleidoscope::hid facade, it SHOULD be the case that making this change can now be done without touching anything other than Kaleidoscope and hardware definitions

algernon:
Long overdue, but with all the mess trying to (optionally) support a boot keyboard, the features in this PR would be of incredible use. It's near the top of my TODO list to review this throughly.

Apologies for the extreme delay in getting here :(

algernon:
Having gone through this a few times, I am not convinced that the `kaleidoscope::hid` facade helps. Quite the contrary, it goes against what the event dispatcher is meant to do: separate the dispatch from the core, and delegate figuring out the dispatch to the hardware plugins. Mind you, the `kaleidoscope::hid` facade makes it a tiny bit easier to transition, so there's that.

I like the direction the original PR is going. If we could delegate routing policy to the end-user (with a sane default provided by the hardware), that would be grand.

(*/me goes back to updating this to current master*)

obra:
One of the issues with getting the original event dispatcher in was that it
was going to end up touching _lots_ of places in the code and plugins
because individual plugins were calling directly to the HID library. The
facade meant that there _should_ only be a single file that needed to be
touched.

algernon:
Yeah. I was thinking a bit backwards.

algernon:
A work in progress rebase of this is on the [f/eventdispatcher](https://github.com/keyboardio/Kaleidoscope/tree/f/eventdispatcher) branch. It works, but there's a bunch of stuff in `kaleidoscope::hid` that need to be moved to `EventDispatcher`. I also want to figure out a way to make it smaller (it adds about 1.5k PROGMEM and 58 bytes of RAM right now).

Not entirely sold on the way dispatch works, by the way. It feels expensive to call every dispatcher, and let them decide whether to dispatch or not. Ideally, I'd prefer a solution where we dispatch to one thing, and have a dynamic policy that can decide where to dispatch. This would mean less code (because most dispatchers won't have to check, just dispatch), better performance (one call instead of N). On the other hand, this would be less flexible, but perhaps that's ok.

I'll add the missing bits and pieces, and we can work from there.

algernon:
Started adding the rest of the wrappers, and... it just keeps growing, and growing. The way I had to implement `isModifierKeyActive` is also less than intuitive.

algernon:
In short, I'll try another way instead, borrowing ideas from this PR, but with a different approach.

algernon:
I'll try to summarize my opinion: what this PR tries to solve needs to be solved, but the more I work with it, the more I'm convinced this is not the best approach. It has some very, very sweet ideas (`PtrArray` is <3, though it could perhaps use a better name; the functionality is something I want to pull into Kaleidoscope), but to add everything that is currently in `kaleidoscope::hid` makes it *huge*. I'm not even halfway through, and am looking at 1.5k of PROGMEM added, and that's a lot, while not providing any benefit for the board I'm working on. Ideally, I'd want something that does what we do now, allow pulling in BootKeyboard too, allow fallback from USB to BLE, etc, but when doing just what we do now, it shouldn't increase our code size so drastically.

Furthermore, while running through handlers works great for event handlers and loop hooks, doing the same for dispatching feels too costy. In most cases, we want to route to one particular object at a time, and no more. In some cases, we may want to change that object, to route future events elsewhere. For this, the `PtrArray`-based setup is not the most efficient - or practical - method.

I'd propose a different route instead: what if we made a base class that has the same methods as the `kaleidoscope::hid` facade (lets call it `kaleidoscope::HID` for now)? We could then provide derived classes like `kaleidoscope::ArduinoHID` (which is pretty much what `kaleidoscope::hid` is now), `kaleidoscope::TeensyHID`, `kaleidoscope::BLEHID`, and `kaleidoscope::MultiHID<ArduinoHID, BLEHID>` (or something along these lines). The Hardware class would have a reference, or a pointer perhaps. It would also have two `setup()` functions: the argument-less one would set the Hardware's HID pointer to whatever is a sane default. Another variant would take an object, and use that as the pointer.

This way we could have a `KeyboardHardware.setup()` that sets up a HID dispatcher that starts as a boot keyboard, but turns NKRO when appropriate. If one is only interested in the NKRO bits, they'd use `KeyboardHardware.setup(&kaleidoscope::ArduinoNKRO())` or something similar. Because the argument-less is never called, it would get optimized out. We could change it anytime, and thus control routing.

I believe this would provide similar levels of customisability, at the fraction of the cost. Some things, like fallback may need a tiny bit more boilerplate, but hopefully not too much.

Does this make sense?

(In any case, I'll have a go at implementing what I tried to describe above, to validate my assumptions.)

algernon:
I think I had an epiphany. What if we shunned such granularity, and embraced a bit of code repetition? What if the hardware definition had the dispatcher hard-wired at compile time, but we'd be able to change that with a preprocessor flag? What if we added a few more boards to `boards.txt`, for a few variants? We could have `Keyboardio Model01 (Full HID)`, `Keyboardio Model01 (NKRO only)`, and so on. Behind the scenes, these would pull in a different dispatch layer, each with its own set of logic. This would allow us to have different dispatchers for different boards, with minimal impact on code size and efficiency. The cost of it is that we'd need a new board - or a compiler flag - to change the dispatcher layer. Since we can make this reasonably user friendly (people will have to select a board *anyway*), I think this is an acceptable compromise. And it is a lot less complicated, a lot cleaner, smaller, and more efficient than the alternatives I could think of.

algernon:
A small clarification to the above: it would solve the problem of being able to have different HID libraries. It would not - in itself - allow run-time event routing. But that could be built on top of it, and may not need to be part of the core firmware.

obra:
@algernon - Would it be possible to have the EventDispatcher be a kaleidoscope plugin instead, so you had to choose one in your sketch, but that it wasn't _tied_ to the hardware definition?

algernon:
I have my doubts about that... even if possible, it would require more RAM and at least a few instructions more per call. But what I worry about most is the compiler not having enough data to optimize things out, if the dispatcher is in a separate compilation & linking unit.

Nevertheless, I can give it a try. Mind you, I tried lifting out (parts of) `kaleidoscope::hid` once, and that was total failure. But that may have been due to a partial lift, and not a complete one.

obra:
I was thinking of maybe even something as gross as different libraries that implement the same class.

algernon:
Oh. Now I understand. We're thinking the same. What I planned - and still plan - is to have these dispatchers be libraries, and the hardware definitions will only differ in which one they pull in. Kinda like how we pull in the Hardware plugins right now.

Is this what you had in mind?

obra:
My objection is to abusing the hardware selector to pick software libraries.

algernon:
Mhm. I see. Will see what I can do, but I don't have high hopes at the moment.

obra:
Sure. I’m happy to talk this through on IRC.

algernon:
@obra just gave a terrific idea on IRC, which shares a lot of code with what I originally had in mind (so I don't have to rethink everything). So next time I have a few hours (~tomorrow), I'll get busy and dump it into code.

Thanks!

algernon:
I think that with #279, and the following #286 (and dependencies), this can be considered fixed, although in a very different way.

PR #139, opened 2017-07-22
Fix paths in README.md to match Makefile

cdisselkoen:
Following the instructions in README.md resulted in build errors for me, due to a disagreement between file paths in README.md and the ones specified in the Makefile.  I assume that the Makefile is correct and not README.md, but I can make the other pull request (changing the Makefile to match README.md) if needed.

It's always possible I'm completely missing something, too.  Did I do something wrong?


obra:
Thank you!

PR #185, opened 2017-09-22
Establish standardized format of cRGB for all hardwares

cdisselkoen:
This is pull request # 1/4 unifying the definition of `cRGB` across hardwares.  The other three pull requests are against `keyboardio/Kaleidoscope-Hardware-Model01`, `keyboardio/KeyboardioScanner`, and `shortcutgg/Kaleidoscope-Hardware-Shortcut`.  None of the four pull requests will build without the others, unfortunately.

The rationale for these pull requests is as follows.  First, a unified definition of `cRGB` is intuitive and elegant.  As part of the core hardware API, it also makes sense to define `cRGB` in the Kaleidoscope core.  

Second, this brings down a major barrier to implementing an actual abstract hardware base class (which is planned as a followup series of pull requests against these same repos; however, the issue of unifying `cRGB` seemed separate enough to warrant separate pull requests with their own discussion).  In fact, the unified definition of `cRGB` is declared by itself in a file called “Kaleidoscope-Hardware.h”, which is very forward-looking as eventually it will simply be declared as part of the abstract hardware base class.

Third, although it may seem at first glance that these pull requests may harm performance in terms of code size and/or execution time, this is actually not the case.  The additional function calls are inlined, and the resulting code generally just takes struct-copy operations that were already happening and changes them to copy struct members to different places.  The code is more verbose, but should compile down to the same number of operations to my knowledge.  To back up these assertions, we compare results of compiling a few sketches before and after these four pull requests are applied.  For the Model 01, we build the two sketches in `keyboardio/Kaleidoscope/examples`; the unified `cRGB` definition results in the exact same code size for `AppSwitcher`, and actually a 2-byte *improvement* for `Kaleidoscope`.  For the Shortcut, we build the sketch in `shortcutgg/Kaleidoscope-Hardware-Shortcut/examples`; again we end up with the exact same code size with or without these four pull requests.  I would imagine these results also indicate execution time hasn’t changed.

### DIFF
--- a/src/Kaleidoscope-Hardware.h
+++ b/src/Kaleidoscope-Hardware.h
@@ -1,0 +1,79 @@
+#pragma once
+
+typedef struct {
+  uint8_t r;
+  uint8_t g;
+  uint8_t b;
+} cRGB;
+
+typedef struct {
+  uint8_t g;
+  uint8_t r;
+  uint8_t b;
+} cGRB;
+
+static inline cGRB asCGRB(const cRGB& crgb) {
+  return (cGRB){crgb.g, crgb.r, crgb.b};
+}
+
+static inline cRGB asCRGB(const cGRB& cgrb) {
+  return (cRGB){cgrb.r, cgrb.g, cgrb.b};
+}
+
+typedef struct {
+  uint8_t r;
+  uint8_t b;
+  uint8_t g;
+} cRBG;
+
+static inline cRBG asCRBG(const cRGB& crgb) {
+  return (cRBG){crgb.r, crgb.b, crgb.g};
+}
+
+static inline cRGB asCRGB(const cRBG& crbg) {
+  return (cRGB){crbg.r, crbg.g, crbg.b};
+}
+
+typedef struct {
+  uint8_t g;
+  uint8_t b;
+  uint8_t r;
+} cGBR;
+
+static inline cGBR asCGBR(const cRGB& crgb) {
+  return (cGBR){crgb.g, crgb.b, crgb.r};
+}
+
+static inline cRGB asCRGB(const cGBR& cgbr) {
+  return (cRGB){cgbr.r, cgbr.g, cgbr.b};
+}
+
+typedef struct {
+  uint8_t b;
+  uint8_t r;
+  uint8_t g;
+} cBRG;
+
+static inline cBRG asCBRG(const cRGB& crgb) {
+  return (cBRG){crgb.b, crgb.r, crgb.g};
+}
+
+static inline cRGB asCRGB(const cBRG& cbrg) {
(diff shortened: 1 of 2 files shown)


algernon:
FWIW, there already is a mostly-finished hardware base class at [keyboardio/Kaleidoscope-Hardware](https://github.com/keyboardio/Kaleidoscope-Hardware). As for `cRGB`, the reason that wasn't in Kaleidoscope is twofold:

- The order of items is hardware dependant, and for some of the operations, it was faster to have them in the native order. But I see you kept the native order where it needs to be kept, yay!
- Not all keyboards have LEDs, it'd look weird to have a dummy one there.

The major reason was speed, though, which looks like you have addressed nicely.

cdisselkoen:
Yes, I'm aware of the mostly-finished base class, and actually designing specifically for it.  Following these pull requests, you can mostly just drop in that hardware base class and it works. 

Regarding keyboards without LEDs, the cRGB struct is part of the hardware API so it makes total sense that it would be defined for all keyboards, as the API should be fully defined for all keyboards.  It's understood for people using those keyboards that the set-LED hardware functions would just be nops (I assume).

obra:
The concept of these PRs makes me very happy.

As I was looking at the changes to KeyboardioScanner and Kaleidoscope-Hardware-Model01, I got a little twitchy about  the naming of 'Model01_cRGB ' and 'asM01cRGB'

Since there are only six possible LED configurations, I _think_ it'd make sense to define them in the core and then just have the hardware definition use the one it wants. (RGB, RBG, BRG, BGR, GBR,GRB) Does that make sense?

cdisselkoen:
Sure.  If we just define structs and inline functions, and they end up not being used, it shouldn't bloat the firmware any.

cdisselkoen:
Stupid question - what does the 'c' stand for?  i.e. should the new structs also have 'c' in their names?  `cRGB`, `cBRG`, `cGBR` etc?

obra:
I’m not actually sure. We got it from FastLED. 
https://github.com/FastLED/FastLED/wiki/Pixel-reference#crgb

cdisselkoen:
I'll bet it stands for "color".  I'll leave it in the others as well.  Note FastLED has a "cHSV", which implies the 'c' is generic across representations

cdisselkoen:
PR updated.

obra:
Thanks!

With the new changes, do we still need to include the change to include the Hardware base class in the same merge?

I don't object to finally getting the dep on -Hardware in there, but if it's not tied to the RGB change, I might make that change separately as I start to merge stuff

obra:
Based on discussion on IRC today, Craig and I agreed that this change isn't necessary. Closing the PR

PR #186, opened 2017-09-28
kaleidoscope-builder: Better error message for sketch-not-found

cdisselkoen:
Currently, if kaleidoscope-builder fails to find a sketch file, the
resulting error message is very confusing and unhelpful.  This commit
makes it more straightforward and helpful.


obra:
Thanks! Applied.

PR #190, opened 2017-10-03
Introduce two layer-switch helper functions

martinakcsilla:
This introduces `ToggleLayer(n)` and `MomentaryLayer(n)`, which make it easier to switch to layers higher than five, and allow one to use enum values in place of `n`, such as: `ToggleLayer(NUMPAD)`, `MomentaryLayer(FUNCTION)`.


obra:
I was just thinking about how this would be helpful earlier tonight. Thank you! I’ll review and merge in the morning.

martinakcsilla:
I'm good at mind reading. Didn't realise I can do that across continents! :rofl:

obra:
I'm thrilled to merge this PR now, though I'm looking for "Better" words for the macros that we can switch to before this gets into wide use.

"Toggle" implies turning on or off. Instead of "ToggleLayer", I'm thinking "SwitchToLayer(n)"

for activating a layer when held, the existing "Momentary" nomenclature I had isn't great. We should find a nice name with an action verb that feels parallel to "SwitchTo"

algernon:
Perhaps `ShiftToLayer`? Everyone knows how shift works - it must be held, thus `ShiftToLayer`.

(Borrowed from KLL, where `Shift` is momentary, `Lock` is toggle, and `Latch` is one-shot)

obra:
That makes a lot of sense.

PR #198, opened 2017-10-06
Add instruction for forked firmware repositories to README.md

shieldsd:


shieldsd:
Having looked a little longer it's probably the Model 01 Firmware that people want to fork.

PR #201, opened 2017-10-07
Add deprecation message for deprecated keyswitch_state functions

cdisselkoen:
Even more deprecation messages for things already deprecated.

Note, these have been deprecated since commit 3154f7c on Jul 24, 2017


cdisselkoen:
Good catch.  PR updated.

PR #202, opened 2017-10-08
Fix URL to USD HID documentation

Katee:
This URL seems to have suffered overzealous replacement. It also appears to have changed on usb.org. I think this new URL is for the correct documentation (relevant table that starts on page 53).


obra:
*sheepish* Thanks!

PR #204, opened 2017-10-08
Fix some markdown oddities (#203)

Jennigma:
Adding some corrections from @jamesnvc


Jennigma:
Wheee! I accepted a pull request into my branch, and then created a pull request to the origin/master. And I kind of know what all those words mean. :-P

PR #205, opened 2017-10-09
Allow injected events tied to physical keyswitches

cdisselkoen:
Currently, injected events *must* use a valid `mappedKey` in their call to `handleKeyswitchEvent()`, and the `row` and `col` parameters are ignored (at least in core).  The theory, I believe, is that injected events are never tied to a physical keyswitch, always to a mapped key identity.

This pull request asks, why can't we allow code to instead (at its option) inject events onto a given physical keyswitch, and have them resolved normally based on whatever layer is active.  This is at least partly a genuine question - is there a reason I'm not aware of that says we can't allow this?  I haven't tested this pull request other than to say it passes CI; I wouldn't even know what test cases would be relevant, or if I've broken anything.  Review from @algernon / @obra is requested.  I'm wondering if the check for the INJECTED flag that I've removed was there simply to avoid out-of-bounds accesses, and therefore is no longer necessary (since we guard against out-of-bounds separately now).  On the other hand, if there was another reason for the check, then by removing it I may have broken something.

I've at least preserved back-compatibility by ensuring that if any existing code does inject keypresses with a valid `mappedKey` *and* a valid `row` and `col`, the `mappedKey` wins and the `row` and `col` are ignored, just as they were before this commit.

### DIFF
--- a/src/key_events.cpp
+++ b/src/key_events.cpp
@@ -48,13 +48,18 @@ void handleKeyswitchEvent(Key mappedKey, byte row, byte col, uint8_t keyState) {
    * In particular, doing them for keypresses with out-of-bounds (row,col)
    *   would cause out-of-bounds array accesses in Layer.lookup(),
    *   Layer.updateLiveCompositeKeymap(), etc.
-   * Perhaps this check should instead be for INJECTED - I'm not clear on
-   *   whether keypresses with out-of-bounds (row,col) are the same as
-   *   keypresses with INJECTED set, or are a superset or a subset of
-   *   INJECTED keypresses.  In any case, the (row,col) test is *safe* in that
-   *   it avoids out-of-bounds accesses, at least in core.  (Can't promise
-   *   anything about event handlers - they may still receive out-of-bounds
-   *   (row,col), and handling that properly is on them.)
+   * Note that many INJECTED keypresses use the UNKNOWN_KEYSWITCH_LOCATION macro
+   *   which gives us row==255, col==255 here.  Therefore, it's legitimate that
+   *   we may have keypresses with out-of-bounds (row, col).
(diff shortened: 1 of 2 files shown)


algernon:
That check was there for two purposes I believe: one to prevent loops, and to allow plugins to override what a key at `row,col` does, while preserving the coordinates. Mostly the latter, I think.

For example, `DualUse` will try to preserve the coordinates, yet, overrides the `mappedKey`. Mind you, your code will preserve this behaviour too. As far as I see, it will work correctly for all existing cases, but I'd like to give this a more through try, because it has the potential to have far-reaching implications.

algernon:
Just a heads up, to say I have not forgotten about this, and I will get around to thoroughly test it soon, hopefully today.

Sorry for taking this long to even reply, things will be much faster from now!

cdisselkoen:
Sure.  The original motivation for this pull request (see cdisselkoen/Kaleidoscope-MacrosOnTheFly#1) is now moot, so I don't have any immediate application in mind for this feature.  However, the pull request still stands as a valid feature addition that someone may eventually have a use for.

gedankenexperimenter:
This change (or something functionally equivalent) is necessary to solve a compatibility issue between certain plugins. For example, the current behaviour prevents Qukeys from working properly with Leader keys (in the case of rollover from a Qukey to a Leader key, the leader key won't function).

obra:
Yup. I'm just waiting on Algernon's review :)

algernon:
Ooops. This fell off my radar again. After promising it'll be faster. Mea culpa!

This is now on top of my TODO list for tomorrow.

gedankenexperimenter:
No rush! I'm not aware of anyone waiting on this change in practice; I just don't want it forgotten entirely.

PR #208, opened 2017-10-11
Update Draft - pull from arley's branch to  jenigma's

obra:
I don't have write access, but through conversation with @Jennigma in the forum I've made a few changes above.

- I gave a first pass up to the Editing Keymaps part, then skimmed after there.
- I rearranged some sentences to try to fit more with the "voice" of the Keyboardio brand from its website, to the emails and tweets. It's a soft, friendly and inviting voice. Specifically I used a lot more "We", and tried to gracefully include lines like "uncompromising typists" and "heirloom-grade" (these distinctions left a powerful first impression on me). To be honest, I could spend A LOT more time doing that brand writing, it's so much fun. If there are parts that could be more "branded" let me know.
- I changed a few instances of "Keyboardio" to "Model 01", but perhaps not consistently, when I saw that "keyboardio" is lowercase in the Arduino board manager I wondered if I have over capitalized. If I thought a part of the software could apply to future Keyboardio-brand products I left it as "Keyboardio", and if it seemed very specific to this product I used "Model 01".
- I softened some parts, like removed the IDE part in the first paragraph, moving that to the Arduino introduction below.
- The only completely-removed line was about requiring cleverness and following instructions. I feel like those aren't required. The offers for support I've seen online suggest the non-clever would also get the help they need! Besides, the non-clever just use whatever flavour of rubber domes Dell is shipping this month amiright?! haha.

### DIFF
--- a/doc/DRAFT_Intro_Docs.md
+++ b/doc/DRAFT_Intro_Docs.md
@@ -2,17 +2,21 @@
 
 <h2>Introduction</h2>
 
-Welcome to your Keyboardio!
+Welcome to your Keyboardio Model 01!
 
 <h3>Audience</h3>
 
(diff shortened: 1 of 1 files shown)


arleym:
Thanks for sorting that!

PR #213, opened 2017-10-15
rename KaleidoscopePlugin.begin to KaleidosopePlugin.initialSetup

obra:
to better explain what it does.

This is part 2 of #196. I (jesse) broke it out because it ends up being...very noisy.

I'm not sure we need to make this change and wanted to split it out from the docs and other code changes.


obra:
@algernon agreed with me on IRC just now to punt this for now. closing.

PR #221, opened 2017-10-28
Specify full path to `stty`

genehack:
I install GNU coreutils on my Macs, to get GNU `ls`. It also installs a GNU-flavored `stty`, which wants `-F` instead of `-f`. Giving the full path to the OS `stty` avoids issues in this case, and should be consequence-free for folks that _don't_ have GNU coreutils installed.


obra:
Thanks. Applied.

PR #222, opened 2017-10-28
Don't flash the device unless we successfully detected the device

genehack:
Nicer error messages are nicer.

### DIFF
--- a/bin/kaleidoscope-builder
+++ b/bin/kaleidoscope-builder
@@ -212,6 +212,15 @@ clean () {
 }
 
 reset_device () {
+    if [ -z $DEVICE_PORT ]; then
+        echo "Couldn't autodetect the keyboard's serial port."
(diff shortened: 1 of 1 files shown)


obra:
@genehack Thanks for the PR. The intent is <3. 

If you're up for it, the error message doesn't quite match the tone we're shooting for with Kaleidoscope (though I know there's no way to have known that.) 

Maybe something like "Couldn't autodetect the keyboard's serial port. If you see this message and your keyboard is connected to your computer, it probably means that our serial port detection logic is buggy or incomplete. Please report this issue at https://github.com/keyboardio/Kaleidoscope"?

Or something else with that kind of sentiment.

genehack:
@obra updated the error message with your suggested wording (with a few line breaks to keep things within the expected width of a terminal).

obra:
Thanks! Applied.

PR #230, opened 2017-11-01
Rearranged and commented Layer.on() and Layer.off()

gedankenexperimenter:
The boolean wasOn was unnecessary, and there was no need to call
bitSet() (or bitClear(), in the case of Layer.off()) if the test
passed. Mostly, I just added a few explanatory comments.

**WARNING**: This change has not been tested on a keyboard yet.

### DIFF
--- a/src/layers.cpp
+++ b/src/layers.cpp
@@ -125,29 +125,37 @@ void Layer_::move(uint8_t layer) {
 }
 
 void Layer_::on(uint8_t layer) {
-  bool wasOn = isOn(layer);
+  // If the target layer was off, turn it on:
+  if (!isOn(layer)) {
+    // Turn on its bit in LayerState
+    bitSet(LayerState, layer);
 
-  bitSet(LayerState, layer);
-  if (layer > highestLayer)
-    updateHighestLayer();
+    // If the target layer is above the previous highest active layer,
+    // update highestLayer
+    if (layer > highestLayer)
+      updateHighestLayer();
 
-  /* If the layer did turn on, update the keymap cache. See layers.h for an
-   * explanation about the caches we have. */
-  if (!wasOn)
+    // Update the keymap cache (but not liveCompositeKeymap; that gets
+    // updated separately, when keys toggle on or off. See layers.h
     updateActiveLayers();
+  }
 }
 
 void Layer_::off(uint8_t layer) {
-  bool wasOn = isOn(layer);
+  // If the target layer was on, turn it off:
+  if (isOn(layer)) {
+    // Turn off its bit in LayerState
+    bitClear(LayerState, layer);
 
-  bitClear(LayerState, layer);
-  if (layer == highestLayer)
-    updateHighestLayer();
+    // If the target layer was the previous highest active layer,
+    // update highestLayer
+    if (layer == highestLayer)
+      updateHighestLayer();
 
-  /* If the layer did turn off, update the keymap cache. See layers.h for an
-   * explanation about the caches we have. */
-  if (wasOn)
+    // Update the keymap cache (but not liveCompositeKeymap; that gets
+    // updated separately, when keys toggle on or off. See layers.h
     updateActiveLayers();
+  }
 }
 
 boolean Layer_::isOn(uint8_t layer) {
--- a/src/layers.cpp
+++ b/src/layers.cpp
@@ -144,7 +144,7 @@ void Layer_::on(uint8_t layer) {
 
 void Layer_::off(uint8_t layer) {
   // If the target layer was already off, return
-  if (bitRead(LayerState, layer))
+  if (!bitRead(LayerState, layer))
     return;
 
   // Turn off its bit in LayerState
--- a/src/layers.cpp
+++ b/src/layers.cpp
@@ -146,7 +146,7 @@ void Layer_::off(uint8_t layer) {
   // If the target layer was already off, return
   if (bitRead(LayerState, layer))
     return;
-  
+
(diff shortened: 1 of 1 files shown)


gedankenexperimenter:
I also think that the call to `isOn(layer)` is unnecessary, given that we're already calling `bitSet(LayerState, layer)`; I think we might as well eliminate and function call overhead (if it hasn't been optimized out by the compiler) and just use `bitRead(LayerState, layer)` directly. I think it's actually clearer that way.

obra:
I'm a little hesitant about these changes.  In general, a lot of the function call overhead gets optimized out. (You can test this by changing just one thing and comparing the compiled size or using the 'make decompile' target and diffing.)

In general, I'd prefer straight doc commits to be separate from commits that change logic. (Adding the doc is a no-brainer, but the micro-optimization of a tricky bit of code is one of those things I or @algernon is going to need to think long and hard about. (At least until there's a good test suite)

gedankenexperimenter:
Sure. When I started to make the change, it was really more about clarity than optimization. The current version reads a bit, then sets that bit even if it was already set. And it uses two different methods for doing those things, so it was a bit confusing. It made more sense to me to reorganize it for better clarity, rather than to simply document the confusing code that I found there.

My motivation here is clarity of the code; any optimization that might take place is purely incidental.

algernon:
I think I like the new code organization better, but as @obra said, will need to check the compiled code too. One thing I would change is that I'd invert the test at the beginning and return in the true branch. Then we could have the main functionality of the function directly in the body, instead of on a branch. This likely doesn't change the compiled code, but would look better, in my opinion, for the human eye.

gedankenexperimenter:
I usually do those tests the way @algernon suggested. Not sure why I did it the other way; I guess I'm pretty rusty.

gedankenexperimenter:
I can't figure out what made that build fail. Travis-ci doesn't give me any useful log info or error message.

obra:
Travis usually shows a log. Something is up. I can't seem to rerun the
build, either.

gedankenexperimenter:
It took a very long time before that build actually got started, too. Hopefully it'll get sorted out soon. At least this change isn't really important.

gedankenexperimenter:
It was two trailing whitespace characters. Incidentally, `make astyle` had no effect on that.

gedankenexperimenter:
> I'm a little hesitant about these changes. In general, a lot of the function call overhead gets optimized out. (You can test this by changing just one thing and comparing the compiled size or using the 'make decompile' target and diffing.)

I finally got around to testing this out. The resulting program is 36 bytes smaller, according to `kaleidoscope-builder` (404792 vs 405220 bytes in the *.elf files).

src/layers.cpp
```diff
@@ -125,29 +125,39 @@ void Layer_::move(uint8_t layer) {
 }
 
 void Layer_::on(uint8_t layer) {
-  bool wasOn = isOn(layer);
+  // If the target layer was already on, return
+  if (bitRead(LayerState, layer))
```
algernon: Is there a reason for using `bitRead` here instead of `isOn`? I think the code would be easier to read if we used `isOn` here (and in `Layer_::off` too).

The compiler should be able to do the right thing even when using `isOn`.

algernon:
Other than the tiny change mentioned above, it's all good. Tested it on a keyboard too, no issues so far.

gedankenexperimenter:
Amusingly, I changed it fro `isOn()` to directly using `bitRead()` because I thought it was clearer (because of the next line, which calls `bitSet()/bitClear()`), not because I thought the function call would incur any overhead. I can see your point of view, though.

src/layers.cpp
gedankenexperimenter: I think it's easier to read the other way around (since the next line calls `bitSet()`), but I see your point of view.

obra:
"Amusingly, I changed it fro isOn() to directly using bitRead() because I thought it was clearer (because of the next line, which calls bitSet()/bitClear()), not because I thought the function call would incur any overhead. I can see your point of view, though."

That led me to be about to suggest that we should be adding a 'Layer::turnOn' function instead of the bitSet...until I realized that 'on' is supposed to be that function... It's not a blocker for merging this, but I wonder if we should be renaming 'Layer::on to Layer::turnOn and adding a deprecation warning for Layer::on. I'll open an issue for discussion.

obra:
Thanks!

obra:
@gedankenexperimenter - I pulled one of the doc comment patches into master on its own while I was working on fixing a merge conflict with this branch.

As I did that, I also cleaned up the commit history so that what's merged into master is a little easier to follow. Before I merge it in, I'd love it if you could check over https://github.com/keyboardio/Kaleidoscope/tree/f/layer-logic to see if it matches with what your -intent- was.

It's actually just a single commit now: https://github.com/keyboardio/Kaleidoscope/commit/dedb2724e1f676feb4c7e14d3ee56c2c18cf8989

gedankenexperimenter:
That looks good to me, but if @algernon still objects to calling `bitRead()` directly instead of `isOn()`, I consider my preference to be unimportant.

algernon:
Doesn't matter much. I find `isOn()` clearer, but this is not a strongly held opinion.

obra:
Oh. That was a merge error on my part. I'll fix it up.

obra:
Merged my branch, which should have had the fixes discussed here. (Including undoing my bitRead misedit)

gedankenexperimenter:
Thanks!

PR #238, opened 2017-11-06
Update key_events.cpp (IS_INTERNAL Handling)

ejinks4:
Reordered if chain in handleSyntheticKeyswitchEvent to fix a bug preventing some Consumer and System Control HID functions from being sent due to bit overload/collision with the IS_INTERNAL flag.

### DIFF
--- a/src/key_events.cpp
+++ b/src/key_events.cpp
@@ -8,9 +8,7 @@ static bool handleSyntheticKeyswitchEvent(Key mappedKey, uint8_t keyState) {
   if (!(mappedKey.flags & SYNTHETIC))
     return false;
 
-  if (mappedKey.flags & IS_INTERNAL) {
-    return false;
-  } else if (mappedKey.flags & IS_CONSUMER) {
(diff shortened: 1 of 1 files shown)


obra:
Based on the discussion elsewhere, I think we would be better off getting rid of “is_internal”

ejinks4:
That would be a cleaner method... Just submitted as this could be a stopgap until that gets figured out.

obra:
I do greatly appreciate the proposal. While I know that this wasn't an
intentional application of Cunningham's Law, it will probably function that
way ;)


(https://meta.wikimedia.org/wiki/Cunningham%27s_Law)

ejinks4:
Haha - have never seen that before.  Very good, sir.

obra:
I was wrong! I'm merging this for now, even though I know it's not the 'right' solution. But I'd like Mute to work correctly for MP2 keyboards and we're on deadline

ejinks4:
Glad to help... Even if it's just a temp fix.

PR #242, opened 2017-11-08
Support all 10bit HID Consumer code.

SjB:
By moving the  IS_CONSUMER flag to B00001000 instead of
B00000010 (swap with IS_INTERNAL) we can detect if the key is a
consumer key and strip out the flags and use the full 10bit to send to
the hid report. This enable us to use all the Consumer_* keys


obra:
Thanks for the PR! 

I'm a little bit uncomfortable with the bit shifts and magic constants. Do you have ideas about refactorings that could move them to macros?

obra:
(I'm happy to do some of the tweaking.)

What would you think about the macro being something more like CONSUMER_KEY() and automatically adding in IS_CONSUMER and SYNTHETIC? I think that'd probably be a bit easier to read and understand.

At the same time, is it safe to wrap _all_ consumer key definitions? 

If so, then we can make the macro naming simpler and more obvious. I'd also probably change the 'mappedKey.raw & 0x03FF' to use a macro rather than a magic constant.

A couple lines of comments about the _why_ of the macro would also not be wrong to help folks trying to understand this code in the future.

Thanks!

obra:
Thanks! Applied.

PR #244, opened 2017-11-12
Prevent reading past the end of the keymaps[] array

gedankenexperimenter:
If [pull request #35 for Model01-Firmware](https://github.com/keyboardio/Model01-Firmware/pull/35) is adopted, this change can prevent turning on a layer higher than the top layer defined in `keymaps[]`, thus preventing reading past the end of that array when a key is defined with `Key_KeymapNext_Momentary` on the top layer (or is visible through transparent keys from the top layer). This would address #243, I believe.

If this technique is used to fix that problem, the variable name in Model01-Firmware must match what's used here. See also [this discussion from the forum](https://community.keyboard.io/t/how-to-get-the-number-of-defined-layers/647).

### DIFF
--- a/examples/AppSwitcher/AppSwitcher.ino
+++ b/examples/AppSwitcher/AppSwitcher.ino
@@ -24,7 +24,7 @@
 #include "Macros.h"
 
 /* *INDENT-OFF* */
-const Key keymaps[][ROWS][COLS] PROGMEM = {
+CREATE_KEYMAP(
   [0] = KEYMAP_STACKED
   (
    Key_NoKey,    Key_1, Key_2, Key_3, Key_4, Key_5, Key_NoKey,
@@ -43,7 +43,7 @@ const Key keymaps[][ROWS][COLS] PROGMEM = {
    Key_RightShift, Key_RightAlt, Key_Spacebar, Key_RightControl,
    M(M_APPCANCEL)
    ),
-};
+	      )
 /* *INDENT-ON* */
 
 
@@ -66,6 +66,3 @@ void setup() {
 void loop() {
   Kaleidoscope.loop();
 }
-
-// This line should always be at the end of the sketch file
-#include <sketch-trailer.h>
--- a/examples/AppSwitcher/AppSwitcher.ino
+++ b/examples/AppSwitcher/AppSwitcher.ino
@@ -46,11 +46,6 @@ const Key keymaps[][ROWS][COLS] PROGMEM = {
 };
 /* *INDENT-ON* */
 
-// Calculate the number of layers defined in the "keymaps[]" array
-// above for use elsewhere in Kaleidoscope. Don't remove or modify
-// this line; it's used to prevent reading past the end of the array!
-const uint8_t LayerCount PROGMEM = sizeof(keymaps) / sizeof(*keymaps);
-
 
 const macro_t *macroAction(uint8_t macroIndex, uint8_t keyState) {
   switch (macroIndex) {
@@ -71,3 +66,6 @@ void setup() {
 void loop() {
   Kaleidoscope.loop();
 }
+
+// This line should always be at the end of the sketch file
+#include <sketch-trailer.h>
--- a/examples/AppSwitcher/AppSwitcher.ino
+++ b/examples/AppSwitcher/AppSwitcher.ino
@@ -46,6 +46,11 @@ const Key keymaps[][ROWS][COLS] PROGMEM = {
 };
 /* *INDENT-ON* */
 
+// Calculate the number of layers defined in the "keymaps[]" array
+// above for use elsewhere in Kaleidoscope. Don't remove or modify
+// this line; it's used to prevent reading past the end of the array!
+const uint8_t LayerCount PROGMEM = sizeof(keymaps) / sizeof(*keymaps);
+
 
 const macro_t *macroAction(uint8_t macroIndex, uint8_t keyState) {
   switch (macroIndex) {
--- a/examples/AppSwitcher/AppSwitcher.ino
+++ b/examples/AppSwitcher/AppSwitcher.ino
@@ -24,7 +24,7 @@
 #include "Macros.h"
 
 /* *INDENT-OFF* */
-CREATE_KEYMAP(
+KEYMAPS(
   [0] = KEYMAP_STACKED
   (
    Key_NoKey,    Key_1, Key_2, Key_3, Key_4, Key_5, Key_NoKey,
@@ -43,7 +43,7 @@ CREATE_KEYMAP(
    Key_RightShift, Key_RightAlt, Key_Spacebar, Key_RightControl,
    M(M_APPCANCEL)
    ),
-	      )
+)
 /* *INDENT-ON* */
 
 
--- a/examples/Kaleidoscope/Kaleidoscope.ino
+++ b/examples/Kaleidoscope/Kaleidoscope.ino
@@ -57,7 +57,7 @@
                                               Key_KeymapNext_Momentary,         Key_KeymapNext_Momentary \
 )
 
-CREATE_KEYMAP(
+KEYMAPS(
   QWERTY,
   GENERIC_FN2,
   NUMPAD
--- a/examples/Kaleidoscope/Kaleidoscope.ino
+++ b/examples/Kaleidoscope/Kaleidoscope.ino
@@ -65,11 +65,6 @@ const Key keymaps[][ROWS][COLS] PROGMEM = {
 
 };
 
-// Calculate the number of layers defined in the "keymaps[]" array
-// above for use elsewhere in Kaleidoscope. Don't remove or modify
-// this line; it's used to prevent reading past the end of the array!
-const uint8_t LayerCount PROGMEM = sizeof(keymaps) / sizeof(*keymaps);
-
 static kaleidoscope::LEDSolidColor solidRed(60, 0, 0);
 static kaleidoscope::LEDSolidColor solidOrange(60, 20, 0);
 static kaleidoscope::LEDSolidColor solidYellow(40, 35, 0);
@@ -109,3 +104,6 @@ void setup() {
 void loop() {
   Kaleidoscope.loop();
 }
+
+// This line should always be at the end of the sketch file
+#include <sketch-trailer.h>
--- a/examples/Kaleidoscope/Kaleidoscope.ino
+++ b/examples/Kaleidoscope/Kaleidoscope.ino
@@ -57,13 +57,12 @@
                                               Key_KeymapNext_Momentary,         Key_KeymapNext_Momentary \
 )
 
-const Key keymaps[][ROWS][COLS] PROGMEM = {
+CREATE_KEYMAP(
   QWERTY,
   GENERIC_FN2,
   NUMPAD
+)
 
-
-};
 
 static kaleidoscope::LEDSolidColor solidRed(60, 0, 0);
 static kaleidoscope::LEDSolidColor solidOrange(60, 20, 0);
@@ -104,6 +103,3 @@ void setup() {
 void loop() {
   Kaleidoscope.loop();
 }
-
-// This line should always be at the end of the sketch file
-#include <sketch-trailer.h>
--- a/examples/Kaleidoscope/Kaleidoscope.ino
+++ b/examples/Kaleidoscope/Kaleidoscope.ino
@@ -65,6 +65,11 @@ const Key keymaps[][ROWS][COLS] PROGMEM = {
 
 };
 
+// Calculate the number of layers defined in the "keymaps[]" array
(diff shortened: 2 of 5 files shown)


gedankenexperimenter:
Okay, I'm very rusty on the intricacies of the C++ compiler and linker. When I build this from `Model01-Firmware`, everything works fine, but when I build Kaleidoscope alone, it fails:

```
[merlin@brunelleschi:~/src/keyboardio/Arduino-Boards/libraries/Kaleidoscope]
$ make
BOARD_HARDWARE_PATH="/Users/merlin/Documents/Arduino//hardware" /Users/merlin/Documents/Arduino//hardware/keyboardio/avr/libraries/Kaleidoscope/bin//kaleidoscope-builder build-all
Building output/AppSwitcher/AppSwitcher (0.0.1-g9db5-dirty) ...
/var/folders/5j/1cp97c6d49v5xwhfqsfkn5w80000gn/T/tmp.DWJ2Y7OK/libraries/Kaleidoscope/Kaleidoscope.a(layers.cpp.o): In function `Layer_::on(unsigned char)':
/Users/merlin/Documents/Arduino/hardware/keyboardio/avr/libraries/Kaleidoscope/src/layers.cpp:181: undefined reference to `LayerCount'
exit status 1
make: *** [build-all] Error 1
```

How is it that the symbol `keymaps` can be used and doesn't cause an error, but the symbol `LayerCount` is an undefined reference? I'm completely stumped.

gedankenexperimenter:
After lots of trial and error, I figured it out; the example sketches need to be updated in order for the build to work.

gedankenexperimenter:
I'd certainly prefer that, too. I think using an `#include` instead of that `sizeof` in every sketch would be a slight improvement, but still not great. But unless someone can think of a better way to accomplish this without using a lot more memory, we may have to choose between the bug and this solution.

gedankenexperimenter:
One more point: making the number of defined layers available ought to make certain other things more efficient, as discussed elsewhere.

gedankenexperimenter:
I suppose the "correct" way to do it would be to store the keymap as a full-fledged object (`class Keymap`?), which is mostly a container for the array. I'm pretty sure it would come with more overhead, but it would be far more robust.

gedankenexperimenter:
Latest change puts the code back in Kaleidoscope, but requires the sketch file to `#include` a new header file: `sketch-trailer.h`.

algernon:
I'm not sure I like the `sketch-trailer.h` method. Feels a lot like magic, something very easy to forget, and if done so, we'd get cryptic error messages.

Perhaps another approach to consider would be the re-introduction of `Kaleidoscope.setup(KEYMAP_SIZE)`, which is much more explicit. Granted, we'd store the size in RAM this way, instead of PROGMEM, but that few bytes I think we can live with.

Of course, the best would be to find a way to figure out the size of the keymap at run-time. I have a few ideas about that, will need to look into them. One of those ideas is playing with the linker to arrange data in a certain way, another is to turn `keymaps` into a struct, and introduce a `KEYMAPS` macro, that also fills out the size. This way we wouldn't need to change `Kaleidoscope.setup`, just wrap the current keymap stuff in a `KEYMAPS` macro and change the type.

All of these are breaking changes, though. But then, so is the `sketch-trailer.h` method...

obra:
Indeed, solutions that add required bits of magic to the user's sketch are
close to "no go" for me.

I think I'd be ok with switching to defining keymaps with a function that
could do internal bookkeeping, especially if it didn't _break_ the current
way of doing things immediately.
ᐧ

algernon:
I'm not sure a function would work. First, the keymap is in PROGMEM, so we need to have it available at compile time. That pretty much rules out any function, unless we move the keymap to RAM (yikes).

Other than that, we could introduce a macro that would define the keymap in such a way that we get its size stored too. The size could be a weak symbol, and default to -1 (unspecified), for backwards compatibility.

To clear things up, this is roughly what I'm thinking of:

```c++
CREATE_KEYMAP(
[QWERTY] = KEYMAP_STACKED(...),
[FN] = KEYMAP_STACKED(...),
[NUMPAD] = KEYMAP_STACKED(...)
);
```

The macro would then define the `keymaps` symbol, and `keymap_size` too. Older sketches would continue to work, because we do not change the signature of `keymaps`, and a default value for `keymap_size` would mean it is not required in the sketch, either.

gedankenexperimenter:
I like the macro idea much better than my `sketch-trailer.h` method. But I definitely think it makes sense to store just a single byte for the number of defined layers, not the full size of the keymap. It's no more "magic" than the current system, and won't break any old sketches.

I still think the best method would be to use a `struct` (or something like a `std::array`) instead, but that does make things less backward-compatible. I'm also wondering if row & column indexing was really a good idea, rather than just using a single byte for `key_index` (or whatever name you like). The row & column way makes some sense for the Model01, but less for other layouts (e.g. Shortcut).

obra:
> 
> . I'm also wondering if row & column indexing was really a good idea, rather than just using a single byte for key_index (or whatever name you like). The row & column way makes some sense for the Model01, but less for other layouts (e.g. Shortcut).
> 
Row and Column maps very closely to the hardware implementation of just about every keyboard in existence even if those rows and columns aren’t visible from the layout of the keys.

gedankenexperimenter:
> The macro would then define the keymaps symbol, and keymap_size too. Older sketches would continue to work, because we do not change the signature of keymaps, and a default value for keymap_size would mean it is not required in the sketch, either.

How would this work?  Defining and initializing the variables is easy, but I can't figure out how to get `keymap_size` initialized with a default value in the case of an old sketch that doesn't use the macro.

gedankenexperimenter:
> Row and Column maps very closely to the hardware implementation of just about every keyboard in existence even if those rows and columns aren’t visible from the layout of the keys.

Sure, but if there's a decent abstraction barrier there, it would make things more efficient (function calls needing fewer arguments & less wasted storage when there are different numbers of keys in each "row")…

gedankenexperimenter:
I'm attempting to make the changes necessary to test out the suggested `CREATE_KEYMAP` macro, but I'm stuck at giving it a default value. I've tried dozens of things, and this is making me despise Arduino, C++, and compiled languages in general.

Maybe there's some obscure trick that would work without requiring changes to the sketch file, but if so, it's beyond my capability. I really do think it's more important to fix this bug, and get the other benefits of having access to the number of layers than it is to make older sketches compile successfully without any changes.

obra:
"I really do think it's more important to fix this bug, and get the other
benefits of having access to the number of layers than it is to make older
sketches compile successfully without any changes."

Breaking backward compatibility like that is something that I'm not willing
to do lightly. The specific failure mode here, as I understand it is "user
defines a sketch with layer-change keys that could go off the deep end" and
that the fix of adding a layer count is a guard against the user shooting
themselves in the foot. That's a noble goal and something I'm happy to
have, but not something I'm ok with breaking the docs and every extant
sketch over on its own. If we're doing something _else_ that is going to
force a backwards incompatible change requiring every user to modify their
sketch, that's a point where we can consider lumping in other backwards
incompatible changes.


On Sat, Nov 18, 2017 at 10:13 PM, Michael Richters <notifications@github.com

algernon:
> How would this work? Defining and initializing the variables is easy, but I can't figure out how to get keymap_size initialized with a default value in the case of an old sketch that doesn't use the macro.

By way of using a weak symbol. Something along these lines:

```c++
int8_t keymap_layers __attribute__((weak)) = -1;

#define CREATE_KEYMAP(...) \
  const Key keymaps[][ROWS][COLS] PROGMEM = { __VA_ARGS__ }; \
  int8_t keymap_layers = sizeof(keymaps) / sizeof(keymaps[0])
```

This way if not using the macro, the `keymap_layers` symbol will use the default definition. If using the macro in the sketh, the same symbol will be redefined with a different value. As the default is a weak reference, the new one will shadow it instead of throwing a compilation error.

gedankenexperimenter:
>This way if not using the macro, the keymap_layers symbol will use the default definition. If using the macro in the sketh, the same symbol will be redefined with a different value. As the default is a weak reference, the new one will shadow it instead of throwing a compilation error.

Have you tried that? I still get an error:

```
Building output/Kaleidoscope/Kaleidoscope (0.0.1-gde39-dirty) ...
In file included from /Users/merlin/Documents/Arduino/hardware/keyboardio/avr/libraries/Kaleidoscope/src/Kaleidoscope.h:26:0,
                 from /Users/merlin/Documents/Arduino/hardware/keyboardio/avr/libraries/Kaleidoscope-Macros/src/Kaleidoscope-Macros.h:3,
                 from /Users/merlin/src/keyboardio/Arduino-Boards/libraries/Kaleidoscope/examples/Kaleidoscope/Kaleidoscope.ino:7:
/Users/merlin/Documents/Arduino/hardware/keyboardio/avr/libraries/Kaleidoscope/src/layers.h:9:11: error: redefinition of 'uint8_t LayerCount'
   uint8_t LayerCount = sizeof(keymaps) / sizeof(*keymaps);
           ^
/Users/merlin/src/keyboardio/Arduino-Boards/libraries/Kaleidoscope/examples/Kaleidoscope/Kaleidoscope.ino:60:1: note: in expansion of macro 'CREATE_KEYMAP'
 CREATE_KEYMAP(
 ^
/Users/merlin/Documents/Arduino/hardware/keyboardio/avr/libraries/Kaleidoscope/src/layers.h:12:9: note: 'uint8_t LayerCount' previously defined here
 uint8_t LayerCount __attribute__((weak)) = 0;
         ^
exit status 1
make: *** [build-all] Error 1
```

gedankenexperimenter:
I don't think this is going to work. I'm pretty sure the variable needs to be both `extern` (since it has to be defined in the sketch, not the Kaleidoscope library) and defined with `__attribute__((weak))`, and if it's declared `extern`, it can't be defined. The C++ compiler defeats me again!

gedankenexperimenter:
Aha! I've now got something that compiles, but I don't know if it's correct. If I don't declare the variable `extern`, but just leave it as global, with the `weak` attribute, and put that declaration in `layers.cpp` instead (and the strong declaration in the sketch file, by way of the macro), it compiles. I'm just not sure that the macro will successfully override the weak declaration, since they're in different compilation units. I'm going to try to come up with a way to test that now.

gedankenexperimenter:
Okay, the current version now works with both old and new sketches. I've tested the default firmware both ways. No changes required to the sketch file, but we don't get any benefit unless the sketch is updated.

gedankenexperimenter:
Fixes #243

algernon:
I'm going to give this a try today with my own sketch, and thinking about that led me to a question: what happens when one does *not* have a keymap in PROGMEM, and has it in EEPROM instead? Or uses a combination of both?

We'll need a way to set `LayerCount` in this case too, but that may be `EEPROM-Keymap`s territory.

algernon:
Answering my own question: in that case, one should not be using `CREATE_KEYMAP()`, and then `LayerCount` will default to zero, and things will not work worse than they do now. One can then set `LayerCount` by themselves.

I may teach `EEPROM-Keymap` to reserve space for `LayerCount` layers, unless told otherwise. That is, `EEPROMKeymap.max_layers()` would have an argumentless variant, that'd use `LayerCount`.

gedankenexperimenter:
I admit that I haven't even looked at `EEPROM-Keymap` yet, but I have tested this out on a keyboard, and it worked. I think I'll try deliberately triggering the `Layer.on()` bug to see what happens, and confirm that this change actually fixes it later today.

gedankenexperimenter:
I tested it out on a variation of the default firmware sketch. The only change I made was mapping one of the `fn` keys to `Key_KeymapNext_Momentary`. Then I held the other `fn` key to get to the function layer (`ShiftToLayer(FUNCTION)`), held the remapped `fn` key to trigger reading past the end of the array. I got plenty of random input, including some control characters.

The same sketch defined with `CREATE_KEYMAP()` instead stayed on the function layer, as it should have.

obra:
For naming: maybe KEYMAPS instead of CREATE_KEYMAP? Since it's really declaring or defining multiple keymaps, rather than creating a single one.

Also on naming, I know that LayerCount matches the code around it, but the code doesn't correctly match the current style: https://github.com/keyboardio/Kaleidoscope/blob/master/doc/CODING_STYLE.md#variable-names

gedankenexperimenter:
I completely agree about the names; I was mimicking `LayerState`, but it never looked right to me, anyway. I've changed it to `layer_count`.

Also, `KEYMAPS()` instead of `CREATE_KEYMAP()` makes more sense to me, as well. I'd be inclined to call it `LAYERS()`, myself, but since the array is called `keymaps`, that would not really help.

(There are a whole lot of names that have caused me a bit of confusion, actually, mostly things named "key" which refer to different kinds of things. I'd love to see a major refactoring change at some point, but certainly not without an overall system for it, and a major release.)

obra:
Thanks!

PR #240, opened 2017-11-06
Redesign of the plugin and hooking interface

noseglasses:
# Abstract
The proposed changes remedy some deficiencies of the current implementation
of the Kaleidoscope plugin interface and the hooking system.

# Anamnesis
The plugin interface and the hooking system are closely linked. Their current 
implementation has a number of flaws in terms of performance, maintainability,
usability, flexibility and safety.

1. It hinders further development, e.g. adding additional hooks. (maintainability)
2. It does waste memory for pointer arrays, especially in case that there are only few hooks defined. (performance)
3. It requires the size of arrays that store hook pointers to be manually
    adjusted when more hooks are supposed to be used. If this is 
    not done, out of bounds access is inevitable. (usability)
4. It does not provide the possibility to pass user data to hook methods. (flexibility)
5. It enforces the use of global variables and functions for 
    hooks and data they operate on. (name conflicts, safety)
6. It requires the `loopHook()` function to be called twice with different parameters, 
   which leads to a high number of unnecessary aborted calls 
   (`if(postClear) ...`). (performance)
7. It does not utilize the information available at compile time
    about the number and the types of the plugins that are actually used in a sketch. 
    This prevents a more optimized implementation. (performance)
8. It provides rarely used methods to remove and replace plugins, something
    that could also be handled within the plugins, thereby removing the need for dynamic hook maintainance. Also the dynamic definition of hooks is rarely needed and could also be 
    taken over by the plugins. (performance, usability)
9. It defines the event handler interface of the Kaleidoscope class and all of its
    data inventory as public members. (encapsulation, safety)
11. It is mostly written in a C-style manner that neglects suitable 
    and valuable features of the C++ language, e,g. templates (usability, performance).

# Pros
Most of the listed deficiencies are fixed by the proposed modifications
of the plugin interface and the hooking system.
**Edit 1:** Moreover, the proposed implementation allows hook methods to be inlined. Something that is not possible with the current pointer-array approach. The current use of function pointers prevents inlining.

# Cons
As the proposed redesign turns hook functions into non-virtual methods of the plugin classes that are
possibly (but not necessarily) derived from the `KaleidoscopePlugin`
base class, compiler features such as the use of the keywords `override` and `final`
cannot be used to enforce correct hook method signatures in derived plugin classes. 
**Edit 1:** ~~This is however not a real draw back if compared to the current implementation. 
There global functions are used exclusively as hooks, which also do not support signature checking.~~
    
# Compatibility 
[...]

PR #262, opened 2017-12-05
Change the INJECTED test to a mappedKey test

gedankenexperimenter:
I'm finding that there's a problem with the `INJECTED` key state test in `handleKeyswitchEvent()`. In [Kaleidoscope-Qukeys](https://github.com/gedankenlab/Kaleidoscope-Qukeys), I need to be able to inject events and have other plugins handle them — even plugins that ignore their own injected events in order to prevent infinite loops. To do this, I want to use a static variable to decide which events my plugin will ignore, and call `handleKeyswitchEvent()` with a `mappedKey`, but without the `INJECTED` flag. The trouble is that the `mappedKey` parameter gets clobbered by a lookup unless `INJECTED` is set, but I think the same thing could be accomplished by simply checking to see if it's not the default `Key_NoKey` instead.

This system (plugins using their own state variables to prevent infinite loops, and skip handling events they inject themselves) has other advantages, as well, allowing plugins to work together better in general. However, I don't know all the plugins well enough to be sure that this change won't break other plugins right now. I suspect not, because if an event handler returns `Key_NoKey`, the next one won't be called, and plugins usually just return `mappedKey` to continue.

@algernon: can you think of anything this will break?

The previous test would clobber any `mappedKey` parameter that the caller supplied unless the `keyState` had the `INJECTED` flag. This change uses a different test, and only does the lookup if the `mappedKey` parameter was unset (i.e. `Key_NoKey`).

### DIFF
--- a/src/key_events.cpp
+++ b/src/key_events.cpp
@@ -75,7 +75,7 @@ void handleKeyswitchEvent(Key mappedKey, byte row, byte col, uint8_t keyState) {
       }
     }
 
-    if (mappedKey == Key_NoKey) {
+    if (mappedKey.raw == Key_NoKey.raw) {
       mappedKey = Layer.lookup(row, col);
     }
 
--- a/src/key_events.cpp
+++ b/src/key_events.cpp
@@ -75,7 +75,7 @@ void handleKeyswitchEvent(Key mappedKey, byte row, byte col, uint8_t keyState) {
       }
     }
 
-    if (!(keyState & INJECTED)) {
+    if (mappedKey == Key_NoKey) {
       mappedKey = Layer.lookup(row, col);
     }
 


algernon:
There's something along these lines in #240 that addresses the same issue in a different way. I agree that the issue needs to be dealt with, but I'm not sure this is the right approach. I'll get to the bottom of #240 this week, and hopefully emerge smarter :)

gedankenexperimenter:
Great! I'll check it out, too.

gedankenexperimenter:
It's not obvious to me how #240 addresses this; maybe @noseglasses can explain. I could pass an out-of-bounds `row` and `col` in order to bypass that check, but the plugin that I want to receive that information needs to know the correct row & column, so that won't work.

On a related note, I want plugins to be able to communicate better — in particular, I want to be able to have a plugin call `handleKeyswitchEvent()` with an argument that indicates that the HID report is (tentatively) complete, and when that function returns, a report will be sent. That way, anything a plugin is waiting for the pre-report hook to do can (and should) be done immediately in its event handler hook.

obra:
"On a related note, I want plugins to be able to communicate better — in
particular, I want to be able to have a plugin call handleKeyswitchEvent()
with an argument that indicates that the HID report is (tentatively)
complete, and when that function returns, a report will be sent. That way,
anything a plugin is waiting for the pre-report hook to do can (and should)
be done immediately in its event handler hook"

Is this something that we could enable simply by having more hook points?

gedankenexperimenter:
> Is this something that we could enable simply by having more hook points?

I'm not immediately seeing how that would help. I was thinking of having a boolean somewhere in Kaleidoscope that would indicate if the key report was "complete" (meaning it's not known to be incomplete because not all the keys have been scanned). If a plugin is injecting an event, other plugins could check to see if the report is complete, and use that to make decisions about what to do.

My use case is an unwritten plugin that I'm working on, intended to de-couple shifted symbols from their base counterparts, and it needs to know if the current report is going to be further altered by the scan before it gets sent in order to decide if it should toggle the shift state. If some other plugin is about to send a report as soon as the current `handleKeyswitchEvent()` call returns, I want to know, so I can do what I would normally do during the pre-report `loopHook()` call.

One possible way to accomplish the "I'm about to send a keyboard report, so get ready" is to use one of the five unused bits in the `keyState` parameter to `handleKeyswitchEvent()`, but I think it still would be nice to store a bit that's globally-accessible that says whether the key report is (tentatively) complete — that might also help with the unreliability of `isModifierPressed()`, too.

obra:
"I'm not immediately seeing how that would help. I was thinking of having a boolean somewhere in Kaleidoscope that would indicate if the key report was "complete" (meaning it's not known to be incomplete because not all the keys have been scanned). If a plugin is injecting an event, other plugins could check to see if the report is complete, and use that to make decisions about what to do."

I may be misunderstanding, but that sure sounds like having a hook point after the scan is completed and another hook point 'just before the report is sent' would make this work better. Plugins that want that complete flag you're describing would hook at the later hook point. 

Of course, if there are multiple plugins that want to take ownership of a given key 'just before the report is sent', there's not a whole lot we can do at the platform level.

gedankenexperimenter:
> I may be misunderstanding, but that sure sounds like having a hook point after the scan is completed and another hook point 'just before the report is sent' would make this work better. Plugins that want that complete flag you're describing would hook at the later hook point.

Thanks; now I get what you mean. Thinking about it…

> Of course, if there are multiple plugins that want to take ownership of a given key 'just before the report is sent', there's not a whole lot we can do at the platform level.

True, only one plugin can be last to act before the report is sent. However, this is more about "can I safely do X when I know that there are a bunch of other keys that will be scanned before the report is sent?" If I know this is the last call to `handleKeyswitchEvent()` before a report goes out, I can make much _better_ decisions, even if I can't guarantee that they'll be invalidated by the next plugin in line.

gedankenexperimenter:
One other point I want to clarify: I don't want to "trick" other plugins into thinking they're doing a normal key scan; I want them to know more about what's about to happen, so they can do what's most appropriate for the situation.

I certainly feel that it's better for a plugin to manage its own state regarding whether or not it should process a key event in the handler than to count on all events flagged `INJECTED` and coming from themselves.

More to the point regarding this specific change proposal: I'm asserting that any call to `handleKeyswitchEvent()` that passes a `mappedKey` other than `Key_NoKey` has been explicitly made by a plugin, and is therefore guaranteed to be "injected". Am I correct about that, or have I missed something?  (Not rhetorical; I do think this change is something that needs to be verified as absolutely correct, if it is made at all)

obra:
Once plugin hook points don't add significant overhead for every plugin, regardless of whether they're used, I'd be VERY up for us thinking about new hook points, names, etc.

gedankenexperimenter:
I will definitely spend some time trying out #240 soon. But I may try to write my Unshifter plugin first, now that I think I've figured out how it should work…

noseglasses:
> It's not obvious to me how #240 addresses this; maybe @noseglasses can explain. I could pass an out-of-bounds row and col in order to bypass that check, but the plugin that I want to receive that information needs to know the correct row & column, so that won't work.

https://github.com/keyboardio/Kaleidoscope/pull/240 does not change the INJECTED logic. The only change to `eventHandlerHook` is a modification of the return value. `false` replaces `Key_NoKey`, `true` means that other handlers may deal with `mappedKey`. This logic may be found in many software projects that support event handling. 

I would generally recommend to adopt this logic for all sorts of future hooks (and maybe even the present loopHook variants). If a hook is meant to modify information that it is passed via call arguments, it is, in my opinion, better to pass them by reference. This enables modification of more than one datum. In the case of the original `eventHandlerHook` the two tasks "modify data if necessary" and "allow subsequent handlers" were mixed up.

Apart from that, I totally agree with @obra. More hooks are really desirable and https://github.com/keyboardio/Kaleidoscope/pull/240 makes hooks very cheap, both with respect to run time and memory consumption. Just one function call of an empty function for each hook, if there are no hooks registered of a certain type and some bytes for the empty virtual function and the vtable entry, both in PROGMEM.

In @gedankenexperimenter's case, it sounds to me that a `preDispatchKeyboardReportHook` would make things way more easy. 

Additional hooks as this one still won't solve the general problem that plugins with different intentions might mess with the hid report in incompatible ways. But that problem cannot be solved (through smart implementation), anyway, given the limited resources of the platform. There will always be incompatible plugins and it will always be up to the user to take care that she/he combines only compatible ones. Plugin designers will have to test and guarantee that their plugins cooperate safely with a defined set of other plugins, a python scripted keyboard simulator and a regression testing system are on their way...

Speaking of incompatibilities. It might be worth to inform Kaleidsocope users that different orders of plugin registration may cause, under certain circumstances, different behaviour.

algernon:
> More to the point regarding this specific change proposal: I'm asserting that any call to `handleKeyswitchEvent()` that passes a `mappedKey` other than `Key_NoKey` has been explicitly made by a plugin, and is therefore guaranteed to be "injected". Am I correct about that, or have I missed something?

As far as I see, this is correct. But note that the reverse, passing `Key_NoKey` to `handleKeyswitchEvent()` does not guarantee that the key has been physically pressed. In other words, you can catch plugin-injected keys with 100% certainty this way, but you can't catch physical presses this way.

gedankenexperimenter:
> As far as I see, this is correct. But note that the reverse, passing `Key_NoKey` to `handleKeyswitchEvent()` does not guarantee that the key has been physically pressed. In other words, you can catch plugin-injected keys with 100% certainty this way, but you can't catch physical presses this way.

Good point. I'm more and more convinced that plugins should keep track of state internally when calling `handleKeyswitchEvent()` in order to stop themselves from processing their own injected events, rather than just blindly ignoring anything flagged `INJECTED`.

I still think the proposed change here is an improvement. Like you said, if `mappedKey != Key_NoKey`, the event has been injected, so we shouldn't clobber `mappedKey`. And it _is_ possible for a plugin to call `handleKeyswitchEvent()` _without_ flagging the key state `INJECTED`.

gedankenexperimenter:
Note: It only just occurred to me that this is exactly the same change as #205, which would be better, because @cdisselkoen included some nice explanatory comments, whereas I did not.

gedankenexperimenter:
> In @gedankenexperimenter's case, it sounds to me that a `preDispatchKeyboardReportHook` would make things way more easy.

Yes, I think so. At first I thought, "but that's just the same as the existing (pre-report) `loopHook` — but that's just how I've been thinking of that hook, and others have probably been using it in other ways.

> Speaking of incompatibilities. It might be worth to inform Kaleidsocope users that different orders of plugin registration may cause, under certain circumstances, different behaviour.

Yes, absolutely.

gedankenexperimenter:
I just checked a few plugins yesterday, and found that this is a compatibility issue for Kaleidoscope-Qukeys with at least two other plugins that it should be able to work with: Kaleidoscope-Leader and Kaleidoscope-TapDance. I'm sure there are others, as well.

Note: this is not just a problem for keys that are defined as both a Leader key (for example) and a Qukey; it's a problem that makes Leader keys not function if pressed while a Qukey is being held and hasn't timed out, as well as not functioning at all if defined as a Qukey (which isn't that unreasonable; I can imagine wanting a Leader key if tapped, but a layer shift if held, for instance – something that would be possible if this change (or #205) is implemented).

Note: I could also make it work by updating each of those incompatible plugins, which I think should be partially done anyway (see https://github.com/gedankenlab/Kaleidoscope-Qukeys/issues/18), but only by stopping them all from flagging injected events as `INJECTED`, which I really think makes no sense. There really should be a way to distinguish "my injected events" from "other injected events", and also distinguish injected events from "real" ones.

obra:
I'm closing this because @gedankenexperimenter confirms it's a dupe of #205

PR #281, opened 2018-01-10
Consistent indentation of kaleidoscope-builder & co

gedankenexperimenter:
I hesitate to suggest this, but it's so much easier to work with these scripts if they're indented consistently. All I did was set them up in a proper Emacs mode and run `indent-region` on the whole file.

Only whitespace should have changed at all, except for the one comment at the top of `kaleidoscope-builder.conf` to cause Emacs to open it in `shell-script` mode.

### DIFF
--- a/bin/find-device-port-linux-udev
+++ b/bin/find-device-port-linux-udev
@@ -2,29 +2,30 @@
 
 use warnings;
 use strict;
-my $vid = shift;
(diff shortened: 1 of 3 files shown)


gedankenexperimenter:
I'll run the perl script through `perltidy`, as suggested by @obra.

PR #284, opened 2018-01-23
Removed ChromeOS modifier workaround

### LINKED ISSUE #283: Extra characters input when using modifier flags
An example, using the default Model01-Firmware:

1. Press key `fn`
2. Press key `C`
3. Press and release key `U` (`{` on `FUNCTION` layer)
4. Repeat step 3 several times
5. Release all keys

…the output I see in a terminal on both macOS and Linux is as follows:
```text
ccccccccC{C{C{C{
```

If, instead, I do this:

1. Press key `fn`
2. Press key `C`
3. Press key `shift`
4. Press and release key `O` (`[` on `FUNCTION` layer)
5. Repeat steps 3 & 4 several times
6. Release all keys

…the output is the expected:
```text
cccccccc{{{{{
```

After holding a printable key, when the next key is typed, the previous one is masked at some level by the host, so the `c` stops repeating, but when a key with the `SHIFT_HELD` flag is pressed, an extra modified character for the held key appears in the output.

gedankenexperimenter:
This workaround introduced a bug that caused spurious input to occur when using modifier flags while holding other keys. There is another fix for it in [KeyboardioHID#29](https://github.com/keyboardio/KeyboardioHID/pull/29), which should also fix the ChromeOS problem, but without causing this bug.

Fixes #283

### DIFF
--- a/src/kaleidoscope/hid.cpp
+++ b/src/kaleidoscope/hid.cpp
@@ -17,40 +17,21 @@ void pressRawKey(Key mappedKey) {
 
 }
 
-void _pressModifierKey(Key mappedKey) {
-  pressRawKey(mappedKey);
-
-  /* On at least ChromeOS 51-60, first sending a "press shift" key event in the same report
-   * as the key report for some shifted keys causes the shift event to be missed.
-   *
-   * Specifically, Shift + [ in the same key report do not genereate a {
-   *
-   * This workaround causes an extra key report to be sent to toggle on the modifier key
-   * before sending the key event we're trying to modify as part of a key report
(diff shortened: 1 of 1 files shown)


gedankenexperimenter:
This PR shouldn't be accepted unless https://github.com/keyboardio/KeyboardioHID/pull/29 is merged first.

gedankenexperimenter:
Since the affected file has been moved to its own repository, I'm closing this PR in favor of keyboardio/Kaleidoscope-HIDAdaptor-KeyboardioHID#3.

PR #289, opened 2018-02-15
Treat device ID values as hex

jwakely:
This avoids Perl warnings when trying to do numeric comparisons on
device IDs that don't consist solely of decimal numbers.

Without this patch I see the following:

    Argument "1d50" isn't numeric in numeric eq (==) at /home/jwakely/Arduino/hardware/keyboardio/avr/libraries/Kaleidoscope/bin/find-device-port-linux-udev line 20.


obra:
Thanks! Applied

PR #300, opened 2018-02-22
Improved fallbacks for macOS serial port detection

gedankenexperimenter:
Several people have reported difficulty flashing firmware on macOS High Sierra because the device port filename doesn't match the serial number from system_profiler. In particular, system_profiler would return a string ending in `E` whereas the device filename would have a `1`. This change adds a check for that filename explictily.

I also corrected the location_id fallback (the substring should have been just 3 characters long, not 4), and it works properly on my system if I make the device shortname 7 characters long, and the filename reverts to using the location id instead.

Last, I added one more check, simply listing the filenames, and searching for a match for the string `kbio01`, which should be present (although in one case, it wasn't).

### DIFF
--- a/bin/find-device-port-macos
+++ b/bin/find-device-port-macos
@@ -66,14 +66,33 @@ if ( exists( $device->{'serial_number'} ) ) {
         print $serial_port_name;
         exit 0;
     }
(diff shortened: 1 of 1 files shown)


gedankenexperimenter:
Well, that was fast!

PR #302, opened 2018-02-26
Added constants for german keyboard layout

michabu:
Specifically these are constants that work with a windows computer. For
Mac OS X there will be a different set of constants needed. I'm
currently not sure what would be the best way to actually support the
layouts of different target OSes.

This is the first draft of a definition for the use with german windows computers. As discussed in [Qwertz in general, german in particular](https://community.keyboard.io/t/qwertz-in-general-german-in-particular/454/23)

### DIFF
--- a/src/kaleidoscope/lang/de_qwertz.h
+++ b/src/kaleidoscope/lang/de_qwertz.h
@@ -1,0 +1,281 @@
+#pragma once
+
+// Alt gr
+#define ALTGR RALT
+
+/* German Umlaute */
+#define Key_de_SZ Key_Minus
+#define Key_de_AE Key_Quote
+#define Key_de_UE Key_LeftBracket
+#define Key_de_OE Key_Semicolon
+
+/* These are deviating from the us layout */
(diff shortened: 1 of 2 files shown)


algernon:
Thanks for the draft! We are currently trying to figure out how to best support different languages and layouts, so it may take a while to move forward with this. Once we have a reasonable way, we'll incorporate your work ASAP.

Sohalt:
Any update on this?
Should be easy to incorporate into https://github.com/keyboardio/Kaleidoscope-Languages, shouldn't it?

obra:
I'd love to see a PR against Kaleidoscope-Languages for this :)
ᐧ

Sohalt:
I've done something here https://github.com/keyboardio/Kaleidoscope-Languages/pull/1

algernon:
Seeing as there's a PR against `Kaleidoscope-Languages` now, I think this can be closed now. I do not think language codes belong to the core firmware (there'd be way too many of them in the end), so doing it in a plugin feels like the right thing to do.

Thanks a lot for everyone involved for pushing this forward!

If anyone disagrees, please reopen this ticket.

PR #306, opened 2018-03-10
Update PROG instruction

Gastove:
Words cannot express how excited I am that I'm writing this up on a Model 01, or how cool I think this entire everything is. 

This PR follows up on https://github.com/keyboardio/Model01-Firmware/issues/44

At least on macOS, the PROG key needs to be held until flashing finishes,
otherwise it will loop an unhelpful error message, maybe indefinitely.

### DIFF
--- a/bin/kaleidoscope-builder
+++ b/bin/kaleidoscope-builder
@@ -67,7 +67,7 @@ prepare_to_flash () {
         size
     fi
 
-    echo "Press ENTER when ready..."
+    echo "When ready, hold down PROG. Press ENTER. Keep holding PROG until flash completes..."
     read a
 }
 


gedankenexperimenter:
It's not (generally) necessary to hold down `prog` until the firmware upload is complete. I flash firmware frequently from an iMac, and I always let go once the upload starts and the LED chase gets going. Failures and error messages are very rare.

Gastove:
@gedankenexperimenter interesting. Per the firmware issue link above, I actually see a failure every time if I don't hold `prog` long enough. (I've also never seen the "LED chase" you're talking about.) 

Is there, perhaps, an appropriate point at which the flash code could notify the user that `prog` could be released?

gedankenexperimenter:
The "LED chase" is a sequence of keys lighting up red on the left half of the keyboard, zig-zagging up and down one column at a time. If you're not seeing that when you upload the firmware, there may be something wrong with the bootloader on your keyboard, but that's not my area of expertise.

Gastove:
Ahhh. I do notice the chase now; I only see it when flashing works, and my hand had been obscuring it ;P

PR #311, opened 2018-04-24
Added keyboard report send policy choice

noseglasses:
The original way of sending keyboard reports once per cycle,
proved to be too inflexible for some applications such as event queueing
plugins (e.g. Papageno or Qukeys).

To make development of such plugins possible/easier, the
original keyboard report send policy (collecting several key events during
matrix scan) was supplemented by a new additional send policy that sends
keyboard reports immediately whenever a key is toggled on/off.

The original policy is preserved as default.
The actual policy that is supposed to be used can be
selected by means of access functions setKeyboardReportSendPolicy()
and getKeyboardReportSendPolicy() of class Kaleidoscope_.
Possible policy choices are
kaleidoscope::KeyboardReportSendCollective (default) and
kaleidoscope::KeyboardReportSendOnEvent.

To allow for ommiting the keyreport that was originally send
once per cycle, the Kaleidoscope_ class gained four additional methods

void preClearLoopHooks(void);
void postClearLoopHooks(void);
void processKeyEvents(void);
void processLoopHooks(void);

that allow to call the original loops substeps individually
and thus to generate a custom main loop that ommits parts
of the default loop (e.g. to suppress the afforementioned keyboard report
submission in the middle of every cycle).

 [This forum discussion](https://community.keyboard.io/t/keyboard-report-clear-in-every-cycle-vs-report-on-key-event/1389) is related and somewhat explains the reason for this PR.

### DIFF
--- a/src/Kaleidoscope.cpp
+++ b/src/Kaleidoscope.cpp
@@ -3,6 +3,8 @@
 
 Kaleidoscope_::eventHandlerHook Kaleidoscope_::eventHandlers[HOOK_MAX];
 Kaleidoscope_::loopHook Kaleidoscope_::loopHooks[HOOK_MAX];
+uint8_t Kaleidoscope_::keyboardReportSendPolicy
+  = kaleidoscope::KeyboardReportSendCollective;
 
 Kaleidoscope_::Kaleidoscope_(void) {
 }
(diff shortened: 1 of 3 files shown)


obra:
@noseglasses - I'm closing this old PR for now. qukeys ended up with an alternate implementation to get what they were hoing to do with something like this. Is it still something we need a solution for to enable what you want to do with Papageno? If so, let's talk it through.

PR #317, opened 2018-05-05
A fix for rollover from a mod-flagged key

gedankenexperimenter:
This (with the matching change to Kaleidoscope-HIDAdaptor-KeyboardioHID) makes it so a rollover from a key with a mod flag applied to one without will not result in the flag from the modified key affecting the next keypress.

This is the other half of keyboardio/Kaleidoscope-HIDAdaptor-KeyboardioHID#5. @noseglasses, please try it out and let me know how well this works for you.

(There is still a problem rolling over between two keys with different modifier flags but the same keycode, but that's a different issue, and not as easy to solve.)

### DIFF
--- a/src/kaleidoscope/hid.h
+++ b/src/kaleidoscope/hid.h
@@ -9,6 +9,11 @@ namespace hid {
 
 extern void initializeKeyboard();
 
+extern bool isPureModifier(Key mappedKey);
+extern void addModFlags(byte flags);
+extern void setModFlagsMask(byte flags);
+extern void pressModFlags(byte flags);
+
 extern void pressKey(Key mappedKey);
 extern void releaseKey(Key mappedKey);
 extern void releaseAllKeys();
--- a/src/kaleidoscope/hid.h
+++ b/src/kaleidoscope/hid.h
@@ -9,7 +9,7 @@ namespace hid {
 
 extern void initializeKeyboard();
 
-extern void setLastKeyPress(Key mappedKey);
+extern void pressToggledOnKey(Key mappedKey);
 
 extern void pressKey(Key mappedKey);
 extern void releaseKey(Key mappedKey);
--- a/src/kaleidoscope/hid.h
+++ b/src/kaleidoscope/hid.h
@@ -9,10 +9,7 @@ namespace hid {
 
 extern void initializeKeyboard();
 
-extern bool isPureModifier(Key mappedKey);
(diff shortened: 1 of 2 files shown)


noseglasses:
Thanks a lot for the quick PR! Currently I don't have my M01 with me, so I cannot try your changes before Tuesday but I am really looking forward to it. I will keep you updated.

gedankenexperimenter:
I just had an idea that might solve to problem of rollover from <kbd>{</kbd> to <kbd>[</kbd> (or _vice versa_). We could check, when a key toggles on, if `mappedKey.keyCode` was in the previous report. If so, we could send a report with that keycode released before sending the new report.

gedankenexperimenter:
…the release could be handled entirely in the HIDAdaptor. I'd replace the `setModFlagsMask()` function with one that registers a new `Key`, both setting `mod_flags_mask` and the newly-pressed keycode. Then, when the report is about to be sent, if the new keycode is non-zero, check its bit in the previous key report, and send an extra report if it is.

gedankenexperimenter:
I've also got an idea for masking held keys that contain modifier flags that are not allowed by the latest keypress, but it wouldn't guarantee that the mask would take effect before the new key press is sent.

noseglasses:
@gedankenexperimenter: I tried you changes and they seem to work with stock Kaleidoscope. At least the rollover issue I described [here](https://github.com/keyboardio/Kaleidoscope-OneShot/issues/26#issuecomment-385458744) is gone. This is definitely good news for Kaleidoscope.

Unfortunately, for me there is still a problem to solve. The additional keyboard report that you introduced in Kaleidoscope-HIDAdaptor-KeyboardioHID seems to cause problems with #311, at least the way I use it with Papageno. I switched to a mode where I send keyboard reports after every event. This apparently does not play well with the additional report.

Currently I don not have much time to investigate this further. But hopefully I will find time to dig into it by the end of the week.

gedankenexperimenter:
Sorry; I was not aware of #311 before, and I find the description incomprehensible (and I don't have time to decipher the code), so you're on your own there.

noseglasses:
> so you're on your own there.

Yes, shure. 
Thanks for fixing the rollover issue anyway. I am sure that a great number of users will benefit from it. And I am positive that I can modify my code so it plays along with your modifications.

obra:
Closing this in favor of #347, which is just an update of the API change.

PR #318, opened 2018-05-10
Define all hooks and event handlers at one point

noseglasses:
This commit adds an X Macro _FOR_EACH_EVENT_HANDLER
to define all information
about hook functions and coresponding event handler methods
at a single place.
By this means it is possible to prevent code redundancies
and make it as simple as possible to add new hooks
or to change their signature in future revisions.

_FOR_EACH_EVENT_HANDLER is invoked in various places whenever hooks and
event handlers are defined or instanciated.

A plugin base class EventHandlerBasePlugin is introduced that implements empty
noop methods for all event handlers.

Event handler documentation was moved from class `Plugin` to definition of _FOR_EACH_EVENT_HANDLER
after the event handlers' explicit definitions have been removed.

I am aware of the intention to provide documentation of the hooks in the Plugin class. But leaving the documentation and the description of the event handlers in `kaleidoscope::Plugin`'s definition would 
require to duplicate the signature of the hooks and thus generate a redundancy.

It might be worth considering if the `_FOR_EACH_EVENT_HANDLER` macro (or how ever you intent to call it) deserves its own header, directly in `src`). The user could be pointed to this header for information about the event handlers that are available in class `Plugin`.


noseglasses:
The code provided does probably not build. I found it too time consuming to setup a Arduino-Boards directory that reflects the required commits in all plugins to be compatible with the WIP branch of Kaleidscope. That's why I did not check it it builds.
There might be some small typos or missing header include paths, but I am quite positive that it is easy to fix those.

noseglasses:
I (ab)used travis to make it build without errors. All commits of this PR might best be squashed.

obra:
Hi!

Thanks for the work on this. For the most part, I'm +1. The one thing I
noticed is that you snuck in an architectural change, moving some code from
Plugin to EventHandlerBasePlugin. If that's not fundamentally tied to the
refactoring being functional, I'd prefer to consider that as a separate
proposal.

-j

noseglasses:
> The one thing I noticed is that you snuck in an architectural change, moving some code from Plugin to EventHandlerBasePlugin. If that's not fundamentally tied to the refactoring being functional, I'd prefer to consider that as a separate proposal.

It's tied because this is the only way to use the `_FOR_EACH_EVENT_HANDLER` macro to define `Plugin`'s event handler methods and at the same time allowing that one single method, namely `onSetup` is overridden to support the legacy plugin API. That's only possible by using inheritance.

obra:
Ok. so it should get wrapped in the guards we use to enable the legacy API.

obra:
and possibly named to show that it’s a temporary hack around a compiler limit for the legacy API, so it’s clear it can go away when we kill the old API

noseglasses:
> and possibly named to show that it’s a temporary hack around a compiler limit for the legacy API, so it’s clear it can go away when we kill the old API

Just added another commit that adds a verbose comment about the possible removal of class `EventHandlerBasePlugin` once the V1 plugin API has been removed.

algernon:
This has been manually merged - thank you!

PR #324, opened 2018-05-16
Minor improvements on Key union

MaxG87:
The Key union is slightly improved by

    * marking functions as constexpr and const 
    * taking arguments as const

At the moment, this might be only cosmetics and modernisations. However, especially the constexpr changes will enable to use Keys in compile time evaluaitions, e.g. allowing to check, whether a retrieved key belongs to a list of expected keys.

### DIFF
--- a/src/key_defs.h
+++ b/src/key_defs.h
@@ -26,41 +26,41 @@ union Key {
   };
   uint16_t raw;
 
-  constexpr inline bool operator==(const uint16_t rhs) const {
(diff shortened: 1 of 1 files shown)


MaxG87:
I just noticed that this PR is very close related to #291.  However, it goes some steps further.

noseglasses:
We might also remove the inline keywords and thus improve readability (if defined inside the class definitions, methods are automatically tagged as inline).

MaxG87:
I just removed inline and pushed the changes.

PR #332, opened 2018-06-03
Update build flag to compile using C++14

MaxG87:
Triggerd by a recent bugfix commit I noticed that Kaleidoscope is build using C++11. I briefly looked for a discussion of this but did not find any. Since the fix is very small, I thought I just push my luck and directly propose an update. I appologise if I missed something here.

I am on Debian Stretch, downloaded somewhen the recent version of ArduinoIDE and build command line only. Using '-std=c++17' was not recognised by avr-g++ but '-std=c++14' was. Thus I chose the latter.

The benefit of this is that it allows some neat new syntax and relaxes the constraints on constexpr functions greatly. Especially the latter could turn out to be very beneficial to get both readable and zero footprint code.

### DIFF
--- a/bin/kaleidoscope-builder
+++ b/bin/kaleidoscope-builder
@@ -192,7 +192,7 @@ compile () {
 	      ${EXTRA_BUILDER_ARGS} \
 	      -build-path "${BUILD_PATH}" \
 	      -ide-version "${ARDUINO_IDE_VERSION}" \
-	      -prefs "compiler.cpp.extra_flags=-std=c++11 -Woverloaded-virtual -Wno-unused-parameter -Wno-unused-variable -Wno-ignored-qualifiers ${ARDUINO_CFLAGS} ${LOCAL_CFLAGS}" \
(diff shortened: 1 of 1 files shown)


obra:
Sadly, I don’t believe C++14 support is available in GCC 4.9, which is the version of avr-gcc distributed by Arduino



https://unix.stackexchange.com/questions/265668/g-doesnt-recognize-standard-14-std-c14

MaxG87:
Your right, Arduino ships with GCC 4.9. However, that must not be an show stopper. GCC 4.9 already supports part of C++14 (https://gcc.gnu.org/projects/cxx-status.html#cxx14). Unfortunately, the cool stuff, regarding constexpr functions is not yet supported.

Furthermore, avr-gcc seems to accept the --std=c++14 argument. At least I was able to compile a very simple test programm with it. Using --std=c++14 would allow to use the already available features and start to use new features as soon as the shipped GCC allows to.

obra:
With the compiler not really supporting C++14, I'm not comfortable with us
officially declaring that we're targeting C++14. I worry that it's going to
lead to a lot of user and developer confusion and a bunch of "works on my
system", especially because some linux distributions ship custom builds of
Arduino with different compiler versions built by the distro.

I'd be thrilled to revisit this once Arduino has moved the state of their
compilers forward.

Best,
Jesse
ᐧ

MaxG87:
I understand. Let us hope that Arduino moves on soon.

Sincerely
Max

PR #343, opened 2018-07-17
Add a LIB_PROPERTIES_PATH option to the build system

a-dma:
Add a variable to specify where the "library.properties" file can be
found so that the build can include a user-define version number.


obra:
I'm ok with this change

PR #349, opened 2018-08-11
clarify what happened to flag B00100000

jamadagni:
This is just to clarify to anyone who may be reading the sources as to why B00100000 is missing from the flags defs. Not a big deal but I thought it would be useful…

### DIFF
--- a/src/key_defs.h
+++ b/src/key_defs.h
@@ -79,6 +79,7 @@ typedef kaleidoscope::Key Key_;
 #define RALT_HELD         B00000100
 #define SHIFT_HELD        B00001000
 #define GUI_HELD          B00010000
+// #define IS_MACRO       B00100000  // defined in Kaleidoscope-Macros/src/MacroKeyDefs.h
 #define SYNTHETIC         B01000000
 #define RESERVED          B10000000
 


obra:
Thanks for this. I feel like we -should- be defining it here. @algernon was this just to try to decouple macros from the core?

jamadagni:
What the… Build failure? The report says:
```
/tmp/tmp.pXs3w4ruYy/libraries/Kaleidoscope/Kaleidoscope.a(key_events.cpp.o): In function `handleKeyswitchEventDefault':
/home/travis/build/keyboardio/current-libraries/Kaleidoscope/src/key_events.cpp:39: undefined reference to `kaleidoscope::hid::pressKey(kaleidoscope::Key, bool)'
exit status 1
make: *** [travis-smoke-examples] Error 1
```
A commented line causing a build failure?

jamadagni:
> Yes, the only reason was to decouple Macros from core. It wouldn't hurt much to have it here, but... it wouldn't be great either.
> I wonder how MouseKeys does its thing... probably under SYNTHETIC. Might that be an option for Macros too?

Personally I think the simplest thing to do is to add the comment in this PR.

There are so many other coding issues worthy of your attention rather than trying to remodel the bit flag for macros…

obra:
Indeed, this is better than nothing, though I'd like to figure out something cleaner eventually.

PR #506, opened 2018-12-19
Add flag `immediately_send_key` to disable eating problematic keys on non-windows OSes

ryansch:
Use case:
```
  static kaleidoscope::SpaceCadet::KeyBinding spacecadetmap[] = {
    {Key_LeftShift, Key_LeftParen, 160}
    , {Key_RightShift, Key_RightParen, 160}
    , {Key_LeftGui,Key_LeftCurlyBracket, 160}
    , {Key_LeftAlt,Key_RightCurlyBracket, 160}
    , {Key_LeftControl,Key_LeftBracket, 160}
    , {Key_RightControl,Key_RightBracket, 160}
    , SPACECADET_MAP_END
  };
  SpaceCadet.map = spacecadetmap;
  SpaceCadet.immediately_send_key = true;
```

With that flag enabled, I'm able to press and hold shift and immediately use it with my mouse to select multiple things on screen.  With that flag set to `false`, I have to wait for the timeout to fire before shift is considered held.

I understand that this can cause problems on windows due to the Alt key so I've set the default for `immediately_send_key` to `false`.

### DIFF
--- a/src/kaleidoscope/plugin/SpaceCadet.cpp
+++ b/src/kaleidoscope/plugin/SpaceCadet.cpp
@@ -39,6 +39,7 @@ SpaceCadet::KeyBinding::KeyBinding(Key input_, Key output_, uint16_t timeout_) {
 SpaceCadet::KeyBinding * SpaceCadet::map;
 uint16_t SpaceCadet::time_out = 1000;
 bool SpaceCadet::disabled = false;
+bool SpaceCadet::immediately_send_key = false;
 
 SpaceCadet::SpaceCadet() {
   static SpaceCadet::KeyBinding initialmap[] = {
@@ -146,7 +147,7 @@ EventHandlerResult SpaceCadet::onKeyswitchEvent(Key &mapped_key, KeyAddr key_add
     //of Alt -- sending Alt by itself activates the menubar.  We don't want to send
     //anything until we know that we're either sending the alternate key or we
     //know for sure that we want to send the originally pressed key.
-    if (valid_key) {
+    if (valid_key && !immediately_send_key) {
       return EventHandlerResult::EVENT_CONSUMED;
     }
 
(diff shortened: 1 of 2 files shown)


ryansch:
Honestly I think I started with 250 and worked down to 160 and never tried going lower.  I could see how a lower timeout would basically get the same effect.

I'm using both Qukeys and SpaceCadet for the following use case:

Qukeys is set up with `LT(FUNCTION, Escape)` on the palm keys to allow me to quickly hit escape when I'm using vim.  I use SpaceCadet with the config above to get quick access to symbols while coding.

When I go to play a game I use my "Gaming Mode" plugin to disable SpaceCadet (https://github.com/ryansch/Model01-Firmware/blob/5a5d174fd1a63b3a7f1af2384ad254db32434359/GamingMode.cpp#L28).  I wasn't sure that I could modify Qukeys mapping on the fly and this was the easiest way to get there.

gedankenexperimenter:
I think I included a macro to toggle Qukeys on and off in the example; it can be enabled and disabled basically the same way as SpaceCadet.

ryansch:
@gedankenexperimenter I want the `LT(FUNCTION, Escape)` to stay active even in "Gaming Mode".  I achieve this with leaving Qukeys enabled but disabling SpaceCadet.

ryansch:
Just rebased this on master.

ryansch:
@algernon That’s excellent feedback!  I’ll work on those changes.

gedankenexperimenter:
Now that SpaceCadet has had a significant rewrite, this change is slightly more difficult to implement. I thought at first that it might even warrant an alternate plugin, but I worked out something fairly minimal as an optional SpaceCadet mode. Once I test it out, I'll submit a PR with a different version of this feature.

PR #509, opened 2018-12-20
Fix path to Model01-Firmware directory in README.md

thunderkeys:
I believe the path to the firmware directory in README.md is missing the 'avr/' component, this pull request/diff corrects it.


obra:
Thank you! (For next time, `git commit -s` will "sign off" on the commit, showing that you agree to the Developer Certificate of Origin. But this one's ok, since it's a doc item.

PR #555, opened 2019-01-29
Simplify function naming to remove all_lower option

obra:
The google style allows all_lowercase names for things that are like accessors and mutators, but does not require it. I find that in our codebase, it just makes things look inconsistent


obra:
Gr. (Did this through the webui. A little annoyed that github doesn't let
me do that there)
ᐧ

PR #570, opened 2019-02-13
Enable safer sending of mid-cycle reports

### LINKED ISSUE #568: The mid-cycle report problem
### Summary

It is fairly common for plugins to send keyboard HID reports in the middle of a scan cycle, typically in response to detecting a key press or release event. The problem with this is that simply calling `hid::sendKeyboardReport()` when only part of the keyswitch scan has been processed is that there might be another key that is being held, but which has not yet been processed during the current scan cycle. This means that it's not safe to base these mid-cycle reports on the (incomplete) current keyboard HID report. This is the same reason that, when checking if a modifier key is being held, we check the _previous_ report, not the current one.

If a mid-cycle report is based on the current HID report, it can easily result in sending a report that doesn't have the keycode for a key that is being held, resulting in a key release, then (on the next report) a key press event in the host OS. In the worst case, it can result in a plugin that sends a mid-cycle report in every cycle rapidly causing a keycode to toggle on and off, sending one keypress per cycle to the OS.

### An Example

This example is trivial, but I think it's a clear demonstration of the problem. Suppose a plugin (e.g. Macros) is designed to send one or more keyboard reports when a certain key (<kbd>1</kbd>) is pressed, and that key is the first one scanned. It first sends a press of `Key_A`, then a release of the same key. By itself, this works just fine, but if another key (<kbd>2</kbd>, scanned second) is still being held, perhaps just rollover, and is mapped to `Key_B` (no special plugin action), here's what happens:

#### sequence of events

1. press <kbd>2</kbd>
1. press <kbd>1</kbd>
1. release <kbd>2</kbd>
1. release <kbd>1</kbd>

#### what happens

1. `b` is added to the report. At the end of the scan cycle, that report is sent to the host, then the report is cleared. No problem.
1. `a` is added to the (empty) report (<kbd>2</kbd> hasn't been processed yet), and a mid-cycle report is sent. This report contains `a`, but not `b`, so the OS detects a simultaneous press of `a` and release of `b`. Next, `a` is removed from the report, and another (empty) report is sent. Then `b` is added to the report. At the end of the scan cycle, a report is sent to the host, which perceives a second press of `b`, even though the user didn't release it and press it again.
1. nothing is added to the report, because the plugin responsible for `a` doesn't act on held keys. The OS receives a second release of `b`.
1. both keys are released, so we're done.

The output will be `bab`, but should be just `ba`.

### An Existing Workaround

Qukeys already deals with this problem by directly accessing the `Keyboard.lastKeyReport` and `Keyboard.keyReport` member variables, which were made public for the purpose. It backs up `Keyboard.keyReport`, and replaces it with the contents of `Keyboard.lastKeyReport`. That way, any keys that haven't yet been scanned already have their keycodes added to the report, and it's safe(r) to send a mid-cycle key report. Then it restores `Keyboard.keyReport` from its backup so that the current cycle can continue as normal.

This works, but it's fragile. Qukeys bypasses the HIDAdaptor and reads and writes internal members of a KeyboardioHID class itself. It won't work with any HID module other than KeyboardioHID. And it only works with the NKRO `Keyboard` class, not the `BootKeyboard` class.

### A General Solution

If we added two functions in the HIDAdaptor, to let plugins prepare the HID module for a mid-cycle report and to restore state after, it would be much simpler for plugins to get this right, and those member variables could be kept private. I'm calling the two functions `hid::stashKeyboardReport()` and `hid::restoreKeyboardReport()`, but they would mostly just call similar methods of the HID module's `Keyboard` and `BootKeyboard` classes, as appropriate. Plugins would not need to know what HID module or type of report is being used, because the HIDAdaptor would insulate them from that.

A downside to doing it this way is that the HID `*Keyboard` classes would now need to store three HID report structures. In the case of the NKRO class, that 87 bytes of RAM (29 more than we currently use), and that needs to be allocated permanently, whereas what Qukeys currently does is storing those extra bytes only as long as necessary.

---

### An Alternative

There is, in my opinion, and even better way to solve this problem, but it's a much more radical change. This would be to build the report only when it's time to send it, from a different data structure that stores the state of the keyboard. We (essentially) already have this structure, in `Layer.live_composite_keymap_[]`, though it would need to be changed to clear entries when keys are released. It would then become possible to store only `lastKeyReport` (for comparisons), and it would obviate the need for the stash & restore functions, because the keyboard state wouldn't need to be cleared entirely every cycle. It would also be easier to get up-to-date information about the state of the whole keyboard.  There would be more ways for keys to get "stuck" on, but some plugins would be able to use that as a feature (e.g. OneShot).

I don't expect this alternative solution to be practical in the short term, and I know I've suggested it before, so I won't harp on it. I do have a working version of it in Kaleidoglyph, though, and I found that it vastly simplifies some plugins.

gedankenexperimenter:
When a plugin needs to send a keyboard report in the middle of the scan cycle, that can trigger bugs where a keyswitch processed later gets dropped from the mid-cycle report, then added again in the regular end-of-cycle report. This PR includes the other half of keyboardio/Kaleidoscope-HIDAdaptor-KeyboardioHID#16, and updates Qukeys to use it instead of directly modifying member variables in KeyboardioHID classes.

Also requires keyboardio/KeyboardioHID#49

Fixes #568

### DIFF
--- a/src/kaleidoscope/hid.h
+++ b/src/kaleidoscope/hid.h
@@ -33,6 +33,13 @@ extern void releaseRawKey(Key mappedKey);
 /** Flushes any pending regular key switch events and sends them out */
 extern void sendKeyboardReport();
 
+// Store a copy of the current report so that a mid-cycle report can be sent based on the
+// previous report instead of the incomplete current report
+extern void stashKeyboardReport();
+// Restore the saved copy of the keyboard report so that the current cycle can be resumed
+// properly after sending a mid-cycle report
+extern void restoreKeyboardReport();
+
 extern boolean isModifierKeyActive(Key mappedKey);
 extern boolean wasModifierKeyActive(Key mappedKey);
 
--- a/src/kaleidoscope/plugin/Qukeys.cpp
+++ b/src/kaleidoscope/plugin/Qukeys.cpp
@@ -173,11 +173,7 @@ bool Qukeys::flushKey(bool qukey_state, uint8_t keyswitch_state) {
   // have a full HID report, and we don't want to accidentally turn
   // off keys that the scan hasn't reached yet, so we force the
   // current report to be the same as the previous one, then proceed
-  HID_KeyboardReport_Data_t hid_report;
-  // First, save the current report
-  memcpy(hid_report.allkeys, Keyboard.keyReport.allkeys, sizeof(hid_report));
-  // Next, copy the old report
-  memcpy(Keyboard.keyReport.allkeys, Keyboard.lastKeyReport.allkeys, sizeof(Keyboard.keyReport));
+  hid::stashKeyboardReport();
   // Instead of just calling pressKey here, we start processing the
   // key again, as if it was just pressed, and mark it as injected, so
   // we can ignore it and don't start an infinite loop. It would be
@@ -188,7 +184,7 @@ bool Qukeys::flushKey(bool qukey_state, uint8_t keyswitch_state) {
   hid::sendKeyboardReport();
 
   // Next, we restore the current state of the report
-  memcpy(Keyboard.keyReport.allkeys, hid_report.allkeys, sizeof(hid_report));
+  hid::restoreKeyboardReport();
 
   // Last, if the key is still down, add its code back in
   if (keyswitch_state & IS_PRESSED)


obra:
@gedankenexperimenter my understanding is that the Qukeys rewrite obsoleted this PR. If that's not the case,  please let us know and we can talk it through.

gedankenexperimenter:
I wouldn't quite agree with that assessment. It is true that the Qukeys rewrite doesn't use this PR, and revealed further problems with mid-cycle reports (or, depending on your point of view, the general design where keyboard state is stored in a USB report) that are not addressed by this PR.

The idea here was to provide a simple system for plugins that do send mid-cycle reports to do so with fewer errors. I'm fine with closing it, as it doesn't address the fundamental design issue that's the source of these problems very well, but the intention was not merely to make Qukeys less hardware-dependent.

obra:
Thanks. I appreciate the clarification.
ᐧ

PR #574, opened 2019-02-16
Numlock state restoration

numist:
Detect current numlock state before entering numpad mode and restore it when exiting.

This fixes an issue where the numlock state was always disabled when exiting the numpad (as tested on my local Ubuntu system) instead of being restored to its previous state.


numist:
This doesn't fix the problem where `clear` is sent on macOS when enabling the numpad. Given that it's not sent when disabling the numpad I suspect that macOS just always reports the numlock LED as being off, which is the opposite of useful. I have an idea for fixing this (detecting when changing the numlock state results in no LED state change and disabling numlock state changes thereafter), but it can come as a separate PR.

numist:
Closing this; after chatting with Jesse it's probably better to avoid numlock state management entirely.

PR #582, opened 2019-02-20
Switch from compiling with C++11 to C++14

gedankenexperimenter:
I changed the compiler flags in kaleidoscope-builder to use `-std=c++-14`. This appears to work on my iMac, but I haven't tested it anywhere else yet.

I'm re-submitting this as a new PR because I want TravisCI to try building it, and when I submitted it as a draft, that wasn't done.

### DIFF
--- a/bin/kaleidoscope-builder
+++ b/bin/kaleidoscope-builder
@@ -235,7 +235,7 @@ compile () {
 	      ${EXTRA_BUILDER_ARGS} \
 	      -build-path "${BUILD_PATH}" \
 	      -ide-version "${ARDUINO_IDE_VERSION}" \
-	      -prefs "compiler.cpp.extra_flags=-std=c++14 -Woverloaded-virtual -Wno-unused-parameter -Wno-unused-variable -Wno-ignored-qualifiers ${ARDUINO_CFLAGS} ${LOCAL_CFLAGS}" \
+	      -prefs "compiler.cpp.extra_flags=-std=c++14 -Wno-sized-deallocation -Woverloaded-virtual -Wno-unused-parameter -Wno-unused-variable -Wno-ignored-qualifiers ${ARDUINO_CFLAGS} ${LOCAL_CFLAGS}" \
 	      -warnings all \
         ${ARDUINO_VERBOSE} \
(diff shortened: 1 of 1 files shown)


gedankenexperimenter:
With this change, I do get a couple of warnings from the compiler that didn't show up before:
```
/Applications/Arduino.app/Contents/Java/hardware/arduino/avr/cores/arduino/new.cpp:29:6: warning: the program should also define 'void operator delete(void*, unsigned int)' [-Wsized-deallocation]
 void operator delete(void * ptr) {
      ^
/Applications/Arduino.app/Contents/Java/hardware/arduino/avr/cores/arduino/new.cpp:33:6: warning: the program should also define 'void operator delete [](void*, unsigned int)' [-Wsized-deallocation]
 void operator delete[](void * ptr) {
```
These can be suppressed by also adding the flag `-Wno-sized-deallocation`, which is probably fine, but I'm uncertain about whether it's a good idea to add that, as well. The warnings pertain to an Arduino file that should not be used in Kaleidoscope at all, since we don't use dynamic memory allocation.

gedankenexperimenter:
I added a commit to suppress those compiler warnings, since it's easy enough to take the first one without the second.

obra:
Hm. I'm somewhat leery of switching to a C++ rev that's not "supported" by
our Arduino core.

This isn't something I absolutely won't do, but it's something I'd like us
to be very careful and deliberate about, to avoid boxing ourselves into a
corner or shooting ourselves in the foot.

ᐧ

gedankenexperimenter:
So far, the only real advantage I can name for my use of C++14 over C++11 is the relaxation of `constexpr` function requirements, so it's not a big deal. I do think it's worth keeping an eye on, though.

obra:
It is 100% worth keeping an eye on. And I appreciate the experimentation
and work.
ᐧ

algernon:
This is something we should do, once Arduino updates their toolchain. I'm putting a wontfix label on it 'till then, but keep the PR open so we won't forget.

obra:
Given that this is already conflicting and was just open as a "reminder", I'm closing it for now. I'm fine with a ticket being open, though.

PR #583, opened 2019-02-21
Suppressed a compiler warning and added warning suppression macros

noseglasses:
This commit silences a warning due to an undefined variable.

A first attemt to locally supress the warning using compiler pragmas was
unsuccessfull due to a compiler bug in all gcc versions. This bug
in all gcc versions below 6.1 makes local diagnostics suppression useless.

As future versions of Arduino will ship with later versions of gcc
from a certain point on, the supplied warning suppression macros will become useful
to avoid the necessity to set global compiler flags like -Wno... and to
enable more precise warning suppression.

The new header file comes with an example that explains how to use
the suppression mechanism.

### DIFF
--- a/src/kaleidoscope_internal/compiler_warnings.h
+++ b/src/kaleidoscope_internal/compiler_warnings.h
@@ -35,17 +35,17 @@
 // This enables suppress only those compiler warnings that are safe to
 // silence without using global flags as -Wno...
 //
-// Use e.g. as 
+// Use e.g. as
 //
 #if 0 // Not part of the example
 
 // Suppress Warnings for unused variables
-COMPILER_DIAG_OFF(unused-variable)
+COMPILER_DIAG_OFF(unused - variable)
 
 // Place code here that generates unavoidable unused variables
 
-COMPILER_DIAG_ON(unused-variable)
-// From this point onward, the compiler continues to warn for 
+COMPILER_DIAG_ON(unused - variable)
+// From this point onward, the compiler continues to warn for
 // unused variables.
 
 #endif // Not part of the example
--- a/src/kaleidoscope_internal/compiler_warnings.h
+++ b/src/kaleidoscope_internal/compiler_warnings.h
@@ -1,0 +1,56 @@
+/* Kaleidoscope - Firmware for computer input devices
+ * Copyright (C) 2013-2018  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+// Please note that due to a bug in older gcc versions the following
+// warning suppressions mechanism does not work for gcc versions below 6.1.
+
+//******************************************************************************
+// Start for internal use only
+//******************************************************************************
+#define _GCC_DIAG_STR(s) #s
+#define _GCC_DIAG_JOINSTR(x,y) _GCC_DIAG_STR(x ## y)
+#define _GCC_DIAG_DO_PRAGMA(x) _Pragma (#x)
+#define _GCC_DIAG_PRAGMA(x) _GCC_DIAG_DO_PRAGMA(GCC diagnostic x)
+//******************************************************************************
+// End for internal use only
+//******************************************************************************
+
+//******************************************************************************
+// Use the following macros to temporarily silence compiler diagnostics (warnings).
+// This enables suppress only those compiler warnings that are safe to
+// silence without using global flags as -Wno...
+//
+// Use e.g. as 
+//
+#if 0 // Not part of the example
+
+// Suppress Warnings for unused variables
+COMPILER_DIAG_OFF(unused-variable)
+
+// Place code here that generates unavoidable unused variables
+
+COMPILER_DIAG_ON(unused-variable)
+// From this point onward, the compiler continues to warn for 
+// unused variables.
+
+#endif // Not part of the example
+
+//******************************************************************************
+#define COMPILER_DIAG_OFF(x) _GCC_DIAG_PRAGMA(push)      __NL__ \
+            _GCC_DIAG_PRAGMA(ignored _GCC_DIAG_JOINSTR(-W,x))
+#define COMPILER_DIAG_ON(x) _GCC_DIAG_PRAGMA(pop)
--- a/src/kaleidoscope_internal/eventhandler_signature_check.h
+++ b/src/kaleidoscope_internal/eventhandler_signature_check.h
@@ -19,7 +19,6 @@
 #include "kaleidoscope/macro_helpers.h"
 #include "kaleidoscope/plugin.h"
 
-
 // *************************************************************************
 // *************************************************************************
 // NOTHING IN THIS HEADER WILL RESULT IN ANY CODE COMPILED INTO KALEIDOSCOPE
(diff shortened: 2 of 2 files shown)


noseglasses:
Talking about warnings... there are a lot of them coming from Kaleidoscope code. It is very hard to find the actual compiler errors in between the actual warnings. 

I did not check which of them appear when compiling with Arduino as I still use Leidokos-CMake with ninja to enable lightning-fast rebuilds. The Arduino-based build system apparently suppresses some warning by using -Wno... compiler arguments. It would be really preferable to cure the cause instead of the symptom. BTW, these compiler warnings are actually valuable. Suppressing them removes means that might help to find errors.

Most warnings come e.g. from unused parameters. These could easily be silenced by writing
```
void unMaskKey(byte /*row*/, byte /*col*/) {}
```

noseglasses:
Just noticed that again I forgot to run make astyle. Thereby I noticed that astyle converts a line like 
```
COMPILER_DIAG_ON(unused-variable)
```
to 
```
COMPILER_DIAG_ON(unused - variable)
```
which is an error in this case.

noseglasses:
What I forgot to mention. The nice thing about the macros `COMPILER_DIAG_OFF` and `COMPILER_DIAG_ON` is that they even work within the definition of other macros, e.g.
```
#define FANCY_MACRO(S) \
   COMPILER_DIAG_OFF(unused-variable) \
   { int a = 1; } \
   COMPILER_DIAG_ON(unused-variable)
```

obra:
Thanks for this.

You mention that you're seeing a lot of compiler warnings. With the default build system, I...don't. I'm concerned that warnings are being swallowed somewhere in the chain. Is it just that you're building with different compiler configs? Or are you only seeing the warnings on newer avr-gcc versions?

noseglasses:
> You mention that you're seeing a lot of compiler warnings. With the default build system, I...don't. I'm concerned that warnings are being swallowed somewhere in the chain. Is it just that you're building with different compiler configs? Or are you only seeing the warnings on newer avr-gcc versions?

The answer to your question was well hidden in one of my previous posts.

> I did not check which of them appear when compiling with Arduino as I still use Leidokos-CMake with ninja to enable lightning-fast rebuilds.

I almost never build using kaleidoscope-builder. I am just too impatient to rebuild the overall firmware every time. Using my own build system and ninja on my 12 core machine, a parallel build takes 8 s and a partial rebuild rarely more than 1 s. 

You are right. I am using different compiler flags but only when it comes to warnings. I had no `-Wno-unused-parameters` and `-Wno-unused-parameters` set. There's absolutely no point in suppression those warnings when developing. However, I understand that you want to spare the user the flood of warnings that I see. It would really be great if the cause of these warnings could be fixed.

Unused parameters are only warned for in implementations that actually don't use the parameters. There are several solutions for this, 
* to ommit the parameter name (bad as not very descriptive)
* using compiler attributes to flag the parameter as unused (hard to read)
* commenting the name of the parameter

Commenting the name is in my opinion the best solution as it solves the problem, remains readable and descriptive.
```cpp
inline int aFunction(int /*a*/, int b) {
   return b;
}
```

Unused parameters are less frequent and could easily be silenced with the same technique that this PR uses.

I am volunteering to fix all unused entity warnings in the firmware. After that you could remove the `-Wno-unused-parameters` and `-Wno-unused-parameters` flags from the build system.

Or is it that there are any third party libraries that cause the compiler to emit such warnings? If so, things are less simple as this case would need individual compiler flags for different libraries. Is that supported by Arduino?

obra:
I see what's going on. kaleidoscope-builder adds " -Woverloaded-virtual
-Wno-unused-parameter -Wno-unused-variable -Wno-ignored-qualifiers" to our
C++ compilations.

@algernon - Do you recall why you're suppressing those? Do you have
objections to @noseglasses' proposal to stop swallowing those warnings? Do
you think that commenting the unused parameter names feels like the most
reasonable solution?

algernon:
I'd keep `-Wno-unused-parameters`, because it's in our case, unused params are fairly common (when we don't use some of the hook params, for example), commenting them out in every case looks incredibly ugly,  I'd rather use a `#define UNUSED __attribute__((unused))` macro instead. We can also do something like QT, and have a `#define UNUSED(x) (void)(x)` macro, and call it in the body to silence the warning. The latter makes the signature the same between `.h` and `.cpp`, and easy to copy & paste.

My very, very strong belief is that commenting variable names out just to silence the warnings is wrong, especially in code we want to keep easily hackable. There's absolutely nothing wrong with unused parameters in the vast majority of cases, so the silencing is nothing else but a workaround for the compiler being too pedantic. I'd rather tell the compiler to shut up about it, and keep `-Wno-unused-parameter`.

I'm ok with removing the others and fixing the code to silence _those_, but I'd very much prefer to not have to dance around to please the compiler for no good reason, if I can also tell it to be more sensible.

Another reason for keeping the flag is that unused params can - and will - appear in existing user code: the `Leader` plugin for example uses callbacks with a single param, but many - most? - callbacks do not use it. `MagicCombo` is very similar too. If we require _end-users_ to work around the compiler too, we failed.

obra:
That's a reasonable explanation to me.  I might call "UNUSED"
"UNUSED_PARAM" just to make it a little clearer.

@algernon - Would it make sense to lift the -Wno-unused-parameter to the
platform.txt's compiler settings and comment it there?

noseglasses:
@algernon, no offence, but you are either too young or you must have forgotten how it was to program in the age of bad compilers that gave you no or just poor hints about the sources of errors.

I still can remember that I once spend a whole weekend searching for an error in a C code that was just due to a missing return statement. Yes, back then, the compilers (especially the C compilers did not warn for these things). Today in the age of llvm-based tools seem ridiculous.

Because of many such experiences over the years I am 100% sure that having verbose compiler warnings is a blessing. IMHO compilers cannot be too pedantic about such things.
Name any of gcc's warnings and I can tell you a nice little story about how much live time that very warning could have saved me if it would have been available. In the end I probably could retire a whole year early if I had saved all that time :-(

Just as an example where `-Wunused-parameter` can help a lot
```cpp
void f(int a_paramter) { // a_paramter was added long after the function was write
   //... longer section of code
   int a_parameter = 666; // oops, someone forgot to remove the original version
   // doing things with a_parameter instead of a_paramter
}
```
I've seen that before and - even worse - I've done it _many_ times. With `-Wno-unused-parameters` such types of errors can take a surprising amount of time to be found in an environment where you cannot easily work with a debugger.

Talking about ugliness of commented parameters, yes they are ugly but only in implementation files changes are required. That's where typical users shouldn't have to look into. But I agree, using `unused(...)` phrases is more readable. For C++11 I would recommend the following:

```cpp
template<class... T> void unused(T&&...){}

void f(int a, int b) {
   unused(a, b);
   //...
}
```




> Another reason for keeping the flag is that unused params can - and will - appear in existing user code: the Leader plugin for example uses callbacks with a single param, but many - most? - callbacks do not use it. MagicCombo is very similar too. If we require end-users to work around the compiler too, we failed.

If users have this unused parameter in their sketch, the compiler warning will direct them exactly where the unused parameter is. An there is nothing enigmatic about the warning message in that case.

algernon:
>  Would it make sense to lift the -Wno-unused-parameter to the platform.txt's compiler settings and comment it there?

I believe so, yes.

> no offence, but you are either too young or you must have forgotten how it was to program in the age of bad compilers that gave you no or just poor hints about the sources of errors.

I very much remember. We're not living in that age now thankfully. I love that we have good and verbose error and warning messages now. However, I don't see how unused params are in any way a problem that we should be fixing.

> Just as an example where -Wunused-parameter can help a lot

That's a valid problem, but it's rare enough that I do not think it is worth the effort of making all the current offenders ugly. Over the history of Kaleidoscope, we did not ever make this mistake. We rarely add new parameters anyway, and in the rare case we do, we can pay a bit more attention.

> Talking about ugliness of commented parameters, yes they are ugly but only in implementation files changes are required.

...and plenty of user-sketches (all of them that are forked from the factory firmware, unless they removed the MagicCombo callback), which would suddenly start emitting warnings. For end-users, warnings are scary (see the threads on the forum where people asked about some of the warnings we had, thinking it was an error). They'll have no clue about what to do with them. The warning message is clear as daylight to you and me, but for someone who's happy things just compile, they're a mystery. They won't know how to fix their sketches, because the warning doesn't tell them _that_.

PR #589, opened 2019-02-27
Matrix address classes and interfaces

noseglasses:
I thought quite a while about whether it makes sense to divide this PR up into several smaller PR but finally decided against as in my eyes everything pretty much belongs together.

# Introduction

This PR introduces matrix address classes and intends to deprecate direct row/column indexing throughout the firmware core and in all core plugins.

In the following the motivation for and the advantages of a modified design are provided. Then the committed changes are explained and the services that are provided by the newly introduced classes are described. After that a description follows of how the transition to the new design has taken place. Finally this text explains where changes have been applied, what has been tested so far and the expected impact on the firmware behavior.

# Motivation

## Structured and uniform design

We have structures/classes for keys, colors and many other small pieces of 
information that belong together. To allow for a uniform object-oriented design it is beneficial
to also have have matrix address classes instead of using raw C-style row/col parameters.

The newly introduced classes provide an easy to use interface for key indexing, validity checks and matrix index calculations.

## Key addresses != LED addresses

Some keyboards have no LEDs, others have some LEDs and some keyboards even have LEDs on all keys. That's why row/col potentially mean a different thing for keys and LEDs. 

Although the newly introduced matrix address classes currently allow free conversion between key and LED addresses, they make it easily possible to prevent this possible conversion if necessary in the future. If no conversion between key and LED indices is desired in the future (currently only used in some LED effect plugins), by replacing the templated copy constructors and assignment operators with non-templated version it is possible to add protection against undesired conversions.

## Debugging

Debugging statements (assertions) could be added to the matrix address classes' accessor methods that test if indices are within the allowed range. This will be especially beneficial once there is a x86 based test bench.

## Reduced RAM footprint

One of the classes (`MatrixAddrCompressed`) introduced by this PR stores row/col information in a single byte. Conversion between row/col and raw offset information is
performed in an efficient manner by avoiding any integer divisions through column padding.

## Flexibility in design 

This PR introduces two matrix address classes both of which are interchangeable and interoperable. One is targeted towards processing speed, the other towards saving storage. The latter codes row/col information in one byte. Every hardware class defines the type of address classes that are optimal for its needs.

As both, `MatrixAddr` and `MatrixAddrCompressed` have compatible interfaces, it is easy to replace one with the other by changing one line in the hardware class. Thus, implementors of new hardware classes or features can test the impact of the different matrix address class designs on their particular version of the firmware.

## Remove/wrap boiler plate code

Some boiler plate code such as 
```cpp
row*COLS + col

row = offset/COLS

col = offset%COLS

if(row >= ROWS || col >= COLS)
```
was moved to the matrix addressing class. This makes many regions of code of the core firmware more readable.

## Key/LED address constants

The new approach allows for the definition of `constexpr` compile time constants for key and LED addresses. This was not possible before and required akward and unsafe preprocessor macros or individual constant definitions for row and col values. The following definition of a Qukey in a firmware sketch now becomes more descriptive.
```cpp
static constexpr KeyAddr LeftPalmKey(3, 6);
...
QUKEYS(
  kaleidoscope::plugin::Qukey(0, LeftPalmKey, Key_LeftControl)
  ...
)
```

# Changes

This PR comes with quite a bunch of changes in many files. Fortunately, most of these changes are of little complexity as they only affect function/method interfaces. Appart from its occurence in function signatures, the use of row/col indexing was replaced by matrix address based access wherever beneficial, thereby using the newly introduced methods `MatrixAddr::isValid()` and `MatrixAddr::offset()` to replace boiler plate code.

## Function- and method-wrappers

Although functions and method signatures have been changed, the firmware is still 100% backward compatible. This is accomplished by introducing wrapper functions/methods that replace the `byte row, byte col` part of a function signature with `KeyAddr keyAddr`. All such wrappers are inline functions and methods. Thus, no runtime penalty is to be expected. 

## Posibility of deprecation of old-style functions and methods

The newly introduced interface functions and methods are intented to be the future defaults. To simplify deprecation of their old-style row/col siblings, a macro `KS_ROW_COL_FUNC` was introduced that may in the future serve as a modifier to supply deprecation attributes (see comments at the end of `MatrixAddr.h`).

A typical wrapper method definition looks as follows
```
// The new interface method
void setCrgbAt(LEDAddr ledAddr, cRGB color);

// The old interface method (the wrapper)
KS_ROW_COL_FUNC void setCrgbAt(byte row, byte col, cRGB color) {
   setCrgbAt(LEDAddr(row, col), color);
}
```
`KS_ROW_COL_FUNC` currently is an emptily defined macro. By defining the following in the future (see `MatrixAddr.h`)
```
#define _DEPRECATED_MESSAGE_ROW_COL_FUNC \
   "Row/col based access functions have been deprecated. Please use " \
   "the KeyAttr/LEDAttr based versions instead."
#define KS_ROW_COL_FUNC DEPRECATED(ROW_COL_FUNC)
```
all calls to old-style interface methods will trigger a deprecation warning. 

Once a future deprecation periode expired, the tag macros `KS_ROW_COL_FUNC` will greatly help identifying 
all deprecated functions and macros that are supposed to be removed.

## Hook `onKeyswitchEvent(...)`

One of the plugin hook methods, `onKeyswitchEvent`, currently also features col/row parameters. As since the v2 plugin API hook methods are non virtual, there is no straight forward way to introduce an equally named wrapper-hook with a different signature (hook overloading is not supported). If a plugin defines a second method that is named as one of the hooks, system would regard this as an error and a `static_assert` would fire. Remember, that's intentional behavior as we want to protect the user from messing with the signature in derived plugins.

Fortunately, by design, the hook infrastrukture can deal especially well with unimplemented or empty inlined hook methods. Because of this, versioning for hook methods becomes possible. This PR introduces a new hook method `onKeyswitchEvent2`, its signature is identical to the one of the original `onKeyswitchEvent` except for replacing the parameters `byte row, byte col` with `KeyAddr keyAddr`. 

By calling both versions of the hook sequencially in `handleKeyswitchEvent(...)` (`key_events.cpp`), the firmware remains backward compatible without adding significant runtime overhead. This is due to the compiler's ability to optimize out emptily defined hook methods.
If a plugin only defines the old style hook, the call to the new-style hook will be optimized out in the same way as it happens with any other hooks that a plugin does not implement. The only restriction that comes with this solution is that a plugin must not implement both the new- and the old-style version of the same hook.

With a technique similar to the one that the firmware core applies to check if a hook implemented as a plugin method has the correct signature, we could possibly implement a future mechanism that allows to deprecate hook methods. Thus, a future deprecation of the original `onKeyswitchEvent` will be possible.

The current implementation, however, supports both version of the hook. And as there is no significant penalty in terms of PROGMEM from supporting two types of hook signatures, a deprecation is of lesser importance (see information about PROGMEM usage later on in this document).

# Services provided by the matrix address classes

## Row/col <-> offset conversion

In many places of the current firmware version we can find the construct `row*COLS + col` that calculates a contiguous zero-based integer offset from a row/col pair. The inverse operations `row = offset/COLS` and `col = offset%COLS` can as well be found in many places. Both types of conversions are now taken over by the matrix address classes. The former by the `offset()` method, the latter by the accessors `row()` and `col()`. This removes a lot of boiler plate code in many places of the firmware.

Note: The header `Kaleidoscope/src/kaleidoscope/addr.h` contains functions whose features are now taken over by the matrix address classes. Thus, `addr.h` could be deprecated.

## Validity check

Instances of the new address classes are, when default initialized, intentionally in invalid state. The validity of the coordinates ot a matrix address may be queried through its `isValid()` method. A valid state is 
characterized as row and col being within the allowed range. This wraps and simplifies the following piece of ubiquitous boiler plate code `if(row >= ROWS || col >= COLS)` which becomes the more descriptive `if(!keyAddr.isValid())`.

## Unknown keyswitch location as a flag

The old style functions used the preprocessor macro `UNKNOWN_KEYSWITCH_LOCATION` as a flag value. This has been superseeded by a

```cpp
static constexpr KeyAddr UnknownKeyswitchLocation;
``` 
that is defined in header `key_events.h` where also `UNKNOWN_KEYSWITCH_LOCATION` is defined.

## Module test for matrix address classes

This PR comes with a module test program for header `MatrixAddr.h`. The test resides in `test/MatrixAddr` in the Kaleidoscope git repository. To run the test on a unixoid platform that supports gcc (host version) and make, run the following
```
cd Kaleidoscope/test/MatrixAddr
make
./test
```

# Where have changes been applied

Great care has been taken not to miss any occurences of functions and macros whose signature comprises row/col. The following regex were used to find such occurrences.
```
\w+\s+row\s*,\s*\w+\s+col
row\s*,\s*col
\brow\b
\bcol\b
```
A regex search was ran recursively on all files `*.h,*.cpp,*.ino,*.md` within the Kaleidoscope-Bundle-Keyboardio repo. This affected not only the Kaleidoscope but also the Kaleidoscop-Hardware-Virtual submodule.

It is important to mention that if any occurences might have been missed, the firmware can still be expected to work correctly. No functions and methods have been removed from any interfaces. Only additional wrappers have been added and the implementations have been adapted in a few places.

There is a small chance that in some corners of the firmware core the old row/col values are still used (e.g. if row/col were named foo/bar). This would mean that some of the newly introduced wrappers would be called. But as all of them were inlined there results zero runtime penalty.

# Expected impact on PROGMEM and processing speed

The submitted PR uses `MatrixAddr` throughout (`MatrixAddrCompressed` is currently unused).

A testfirmare has been build with the current master bundle and with this PR included. Both versions have been compared in terms of binary size, symbol count and symbol size (function binary code).
With this PR comes a slight increase of firmware size of ~0.5% (84 bytes) for a medium sized firmware of ~16000 bytes. This can be explained by the newly introduced additional hook function. The firmware can be expected to shrink again back to the original size once the old-style access functions and the superfluous hook have been removed.

Disassemblies have been checked at random. Some functions/methods grow, some shrink due to the submitted changes. In average no significant change in size of symbols can be observed. The total amount of symbols remains almost exactly the same. Only one additional function was found using avr-objdump which is a Qukeys wrapper constructor that the compiler for strange reasons didn't inline.

As algorithms remain untouched and the average binary size of most functions did not change, no change in runtime is to be expected after merging the submitted changes.

An alternative version that uses `MatrixAddrCompressed` was also tested. For the above mentioned test firmware the increase of firmware binary size is ~3% (458 bytes). But same as for the submitted version, some increase in PROGMEM consumption (~84 bytes) results from additional wrappers and hook function.

To use such an an alternative version only makes sense when a plugin is used that needs to store a greater number of key addresses in memory. With `MatrixAddrCompressed` this would half the RAM consumption for matrix address storage. In such a use case a slight increase in binary size would probably be acceptable.

# Tests

As this is a quite intrusive set of changes, special care must be taken not to break any existing code. The following types of tests have been run so far:

* compile and link tests with the Model01 firmware and all available plugins

The author of this PR could only run tests with the Keyboardio Model01. Tests with other hardware would be useful before merging this PR.

In general, before this PR is going to be merged, it is important that as many eyes have checked it as possible!

# Aknowledgements

The design of the MatrixAddr class was in parts influenced by gedankenexperimenters [KeyAddr](https://github.com/gedankenlab/Kaleidoglyph-Hardware-Model01/blob/2f771f63184f6cefee83c20df4bc210fd808b85f/src/model01/KeyAddr.h) class.

### DIFF
--- a/src/kaleidoscope/Hardware.h
+++ b/src/kaleidoscope/Hardware.h
@@ -62,7 +62,7 @@ class Hardware {
   // for matrix addressing we define default key and led address classes.
   // Those typedefs are supposed to overridden by derived hardware classes.
   typedef MatrixAddr<0, 0> KeyAddr;
-  typedef MatrixAddr<0, 0> LEDAddr;
+  typedef MatrixAddr<0, 0> KeyLEDAddr;
   /**
    * @defgroup kaleidoscope_hardware_leds Kaleidoscope::Hardware/LEDs
    * @{
@@ -81,7 +81,7 @@ class Hardware {
    * @param led_addr is the matrix address of the LED.
    * @param color is the color to set the LED to.
    */
-  void setCrgbAt(LEDAddr led_addr, cRGB color) {}
+  void setCrgbAt(KeyLEDAddr led_addr, cRGB color) {}
   /**
    * Set the color of a per-key LED at a given row and column.
    *
@@ -124,7 +124,7 @@ class Hardware {
   * @returns The index of the LED at the given position, or -1 if there are no
   * LEDs there.
   */
-  int8_t getLedIndex(LEDAddr led_addr) {
+  int8_t getLedIndex(KeyLEDAddr led_addr) {
     return -1;
   }
   /**
--- a/src/kaleidoscope/Hardware.h
+++ b/src/kaleidoscope/Hardware.h
@@ -21,6 +21,8 @@
 
 #pragma once
 
+#include "kaleidoscope/MatrixAddr.h"
+
 #ifndef CRGB
 #error cRGB and CRGB *must* be defined before including this header!
 #endif
@@ -55,6 +57,12 @@ namespace kaleidoscope {
  */
 class Hardware {
  public:
+
+  // To satisfy the interface of those methods that allow
+  // for matrix addressing we define default key and led address classes.
+  // Those typedefs are supposed to overridden by derived hardware classes.
+  typedef MatrixAddr<0, 0> KeyAddr;
+  typedef MatrixAddr<0, 0> LEDAddr;
   /**
    * @defgroup kaleidoscope_hardware_leds Kaleidoscope::Hardware/LEDs
    * @{
@@ -70,11 +78,21 @@ class Hardware {
    * Setting the color does not need to take effect immediately, it can be
    * delayed until @ref syncLeds is called.
    *
+   * @param led_addr is the matrix address of the LED.
+   * @param color is the color to set the LED to.
+   */
+  void setCrgbAt(LEDAddr led_addr, cRGB color) {}
+  /**
+   * Set the color of a per-key LED at a given row and column.
+   *
+   * Setting the color does not need to take effect immediately, it can be
+   * delayed until @ref syncLeds is called.
+   *
    * @param row is the logical row position of the key.
    * @param col is the logical column position of the key.
    * @param color is the color to set the LED to.
    */
-  void setCrgbAt(byte row, byte col, cRGB color) {}
+  KS_ROW_COL_FUNC void setCrgbAt(byte row, byte col, cRGB color) {}
   /**
    * Set the color of a per-key LED at a given LED index.
    *
@@ -99,6 +117,17 @@ class Hardware {
     return c;
   }
   /**
+  * Returns the index of the LED at a given row & column.
+  *
+  * @param led_addr is the matrix address of the LED.
+  *
+  * @returns The index of the LED at the given position, or -1 if there are no
+  * LEDs there.
+  */
+  int8_t getLedIndex(LEDAddr led_addr) {
+    return -1;
+  }
+  /**
    * Returns the index of the LED at a given row & column.
    *
    * @param row is the logical row position of the key.
@@ -107,7 +136,7 @@ class Hardware {
    * @returns The index of the LED at the given position, or -1 if there are no
    * LEDs there.
    */
-  int8_t getLedIndex(uint8_t row, byte col) {
+  KS_ROW_COL_FUNC int8_t getLedIndex(uint8_t row, byte col) {
     return -1;
   }
   /** @} */
@@ -153,10 +182,28 @@ class Hardware {
    * Masking a key out means that any other event than a release will be
    * ignored until said release.
    *
+   * @param key_addr is the matrix address of the key.
+   */
+  void maskKey(KeyAddr key_addr) {}
+  /**
+   * Mask out a key.
+   *
+   * Masking a key out means that any other event than a release will be
+   * ignored until said release.
+   *
    * @param row is the row the key is located at in the matrix.
    * @param col is the column the key is located at in the matrix.
    */
-  void maskKey(byte row, byte col) {}
+  KS_ROW_COL_FUNC void maskKey(byte row, byte col) {}
+  /**
+   * Unmask a key.
+   *
+   * Remove the mask - if any - for a given key. To be used when the mask
+   * needs to be removed without the key being released.
+   *
+   * @param key_addr is the matrix address of the key.
+   */
+  void unMaskKey(KeyAddr key_addr) {}
   /**
    * Unmask a key.
    *
@@ -170,12 +217,22 @@ class Hardware {
   /**
    * Check whether a key is masked or not.
    *
+   * @param key_addr is the matrix address of the key.
+   *
+   * @returns true if the key is masked, false otherwise.
+   */
+  bool isKeyMasked(KeyAddr key_addr) {
+    return false;
+  }
+  /**
+   * Check whether a key is masked or not.
+   *
    * @param row is the row the key is located at in the matrix.
    * @param col is the column the key is located at in the matrix.
    *
    * @returns true if the key is masked, false otherwise.
    */
-  bool isKeyMasked(byte row, byte col) {
+  KS_ROW_COL_FUNC bool isKeyMasked(byte row, byte col) {
     return false;
   }
   /** @} */
@@ -224,12 +281,22 @@ class Hardware {
   /**
    * Check if a key is pressed at a given position.
    *
+   * @param key_addr is the matrix address of the key.
+   *
+   * @returns true if the key is pressed, false otherwise.
+   */
+  bool isKeyswitchPressed(KeyAddr key_addr) {
+    return false;
+  }
+  /**
+   * Check if a key is pressed at a given position.
+   *
    * @param row is the row the key is located at in the matrix.
    * @param col is the column the key is located at in the matrix.
    *
    * @returns true if the key is pressed, false otherwise.
    */
-  bool isKeyswitchPressed(byte row, byte col) {
+  KS_ROW_COL_FUNC bool isKeyswitchPressed(byte row, byte col) {
     return false;
   }
   /**
--- a/src/kaleidoscope/MatrixAddr.h
+++ b/src/kaleidoscope/MatrixAddr.h
@@ -20,6 +20,32 @@
 #include <stdint.h>
 
 namespace kaleidoscope {
+   
+namespace internal {
+   
+template<typename T1, typename T2>
+struct IsSame {
+   static constexpr bool value = false;
+};
+
+template<typename T>
+struct IsSame<T, T> {
+   static constexpr bool value = true;
+};
+
+template<typename T>
+struct RemoveReference {
+   typedef T type;
+};
+
+template<typename T>
+struct RemoveReference<T&> {
+   typedef T type;
+};
+
+template<bool B> struct Bool2Type {};
+
+} // end namespace internal
 
 template<uint8_t rows__, uint8_t cols__>
 class MatrixAddr {
@@ -51,6 +77,7 @@ class MatrixAddr {
     : offset_(offset) {}
 
   constexpr MatrixAddr(const ThisType &other) : offset_(other.offset_) {}
+  constexpr MatrixAddr(ThisType &&other) : offset_(other.offset_) {}
 
   template<typename MatrixAddr__>
   constexpr MatrixAddr(const MatrixAddr__ &other)
@@ -81,7 +108,7 @@ class MatrixAddr {
 //   }
 
   constexpr bool isValid() const {
-    return (row() < rows__) && (col() < cols__);
+    return offset_ < upper_limit;
   }
 
   ThisType &operator=(const ThisType &other) {
@@ -110,10 +137,63 @@ class MatrixAddr {
     --*this;         // call the prefix increment
     return copy;
   }
-
+  
+  // The following templates are necessary as we cannot have both,
+  // a templated and a non templated version of operator=.
+  // Because of this, we have to take a detour over auxiliary
+  // assign methods, one that enables to assign the offset_ member directly,
+  // in case a compatible MatrixAddr class is used and 
+  // one that does a row/col conversion otherwise.
+  
+  template<uint8_t xrows__, uint8_t xcols__>
+  static void assign(internal::Bool2Type<true>, ThisType &self, const MatrixAddr<xrows__, xcols__> &other) {
+    self.offset_ = other.offset_;
+  }
+      
+  template<uint8_t xrows__, uint8_t xcols__>
+  static void assign(internal::Bool2Type<false>, ThisType &self, const MatrixAddr<xrows__, xcols__> &other) {
+    self.offset_ = other.row() * cols + other.col();
+  }
+  
+  // Normal assignment
+  //
   template<typename MatrixAddr__>
   ThisType& operator=(const MatrixAddr__ & other) {
-    offset_ = other.row() * cols + other.col();
+     assign(
+        // A dummy value that is only passed to make the compiler select
+        // the right overload of assign(...). It will be optimized away
+        // by the compiler, i.e. not be put on the stack when
+        // calling the function as the class IsSame is empty.
+        //
+        internal::Bool2Type<
+           internal::IsSame<
+              ThisType, 
+              typename internal::RemoveReference<MatrixAddr__>::type
+           >::value
+        >(), 
+        *this, other
+     );
+    return *this;
+  }  
+  
+  // Move assignment
+  //
+  template<typename MatrixAddr__>
+  ThisType& operator=(MatrixAddr__ && other) {
+     assign(
+        // A dummy value that is only passed to make the compiler select
+        // the right overload of assign(...). It will be optimized away
+        // by the compiler, i.e. not be put on the stack when
+        // calling the function as the class IsSame is empty.
+        //
+        internal::Bool2Type<
+           internal::IsSame<
+              ThisType, 
+              typename internal::RemoveReference<MatrixAddr__>::type
+           >::value
+        >(), 
+        *this, other
+     );
     return *this;
   }
 
@@ -145,6 +225,12 @@ class MatrixAddr {
 
   bool operator==(const ThisType &other) const {
     return offset_ == other.offset_;
+  }  
+  
+  template<typename MatrixAddr__>
+  bool operator==(const MatrixAddr__ & other) {
+     return   (this->row() == other.row())
+           && (this->col() == other.col());
   }
 
   class forward_iterator {
--- a/src/kaleidoscope/MatrixAddr.h
+++ b/src/kaleidoscope/MatrixAddr.h
@@ -1,0 +1,385 @@
+/* Kaleidoscope - Firmware for computer input devices
+ * Copyright (C) 2013-2019  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// This header contains two classes that aid
+// indexed access to key and LED matrix entries.
+// There are two different implementations that feature the same
+// interface and are therefore interchangeable and also cooperate, one is
+// more memory efficient, the other prefers speed when it comes to extracting
+// row and column indices.
+//
+// You should always compare the performance of both on your hardware
+// and then select the one that is better suited for your needs.
+//
+// As both implementations are compatible, it is possible to use one type of
+// matrix address class for key addressing and another for LED addressing.
+//
+// To define the types of matrix address storage class that is used
+// throughout the firmware for key and LED accessing, provide the following
+// two typedefs in your keyboard hardware class.
+//
+// typedef MatrixAddr<4, 16> KeyAddr;
+// typedef MatrixAddrCompressed<4, 16> LEDAddr;
+//
+// Those two types are then automatically used throughout the whole
+// firmware for any type of addressing.
+
+#pragma once
+
+#include <limits.h>
+#include <stdint.h>
+
+namespace kaleidoscope {
+namespace internal {
+namespace matrix_addr {
+
+constexpr unsigned log2(unsigned n, unsigned p = 0) {
+  return (n <= 1) ? p : log2(n / 2, p + 1);
+}
+
+template<typename Int1__, typename Int2___>
+constexpr Int1__ shiftNOr(Int1__ v, Int2___ s) {
+  return v | (v >> s);
+}
+
+// The following functions are used to determine the optimal
+// storage row size for compressed storage. They compute the next greater
+// power of two for an unsigned integer value of a given size.
+
+constexpr
+uint32_t upperPowerOfTwo(uint32_t v) {
+  // From https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+
+  // The following is not possible in a c++11 constexpr function
+//     v--;
+//     v |= v >> 1;
+//     v |= v >> 2;
+//     v |= v >> 4;
+//     v |= v >> 8;
+//     v |= v >> 16;
+//     v++;
+//     return v;
+
+  // Thus, we have to reformulate it a little...
+  return shiftNOr(shiftNOr(shiftNOr(shiftNOr(shiftNOr(v - 1, 1), 2), 4), 8), 16) + 1;
+}
+
+constexpr
+uint16_t upperPowerOfTwo(uint16_t v) {
+  // From https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+
+  // The following is not possible in a c++11 constexpr function
+//     v--;
+//     v |= v >> 1;
+//     v |= v >> 2;
+//     v |= v >> 4;
+//     v |= v >> 8;
+//     v++;
+//     return v;
+
+  // Thus, we have to reformulate it a little...
+  return shiftNOr(shiftNOr(shiftNOr(shiftNOr(v - 1, 1), 2), 4), 8) + 1;
+}
+
+constexpr
+uint8_t upperPowerOfTwo(uint8_t v) {
+  // From https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+
+  // The following is not possible in a c++11 constexpr function
+//     v--;
+//     v |= v >> 1;
+//     v |= v >> 2;
+//     v |= v >> 4;
+//     v++;
+//     return v;
+
+  // Thus, we have to reformulate it a little...
+  return shiftNOr(shiftNOr(shiftNOr(v - 1, 1), 2), 4) + 1;
+}
+
+template<typename Int__> struct IntLimitsTrait {};
+template<> struct IntLimitsTrait<uint8_t> {
+  static constexpr uint8_t max = 0xFF;
+};
+template<> struct IntLimitsTrait<uint16_t> {
+  static constexpr uint16_t max = 0xFFFF;
+};
+template<> struct IntLimitsTrait<uint32_t> {
+  static constexpr uint32_t max = 0xFFFFFFFF;
+};
+
+} // end namespace internal
+} // end namespace matrix_addr
+
+// The MatrixAddr template class is meant to be used when speed is
+// more important than saving memory. It uses two bytes to store row and column
+// indices. For most keyboardsw that feature less than 256 indices keys/LEDs/...
+// the row/col information can be stored in a more memory efficient fashion.
+// In that case if saving memory is paramount, use
+// prefer template class MatrixAddrCompressed that is defined later on
+// in this file.
+//
+template<uint8_t rows__, uint8_t cols__>
+class MatrixAddr {
+ private:
+  uint8_t row_;
+  uint8_t col_;
+
+ public:
+
+  static constexpr uint8_t rows = rows__;
+  static constexpr uint8_t cols = cols__;
+
+  static constexpr uint8_t row_size = cols__;
+
+  static constexpr uint8_t matrix_size = rows__ * cols__;
+
+  static constexpr uint8_t invalid_index = 255;
+
+  constexpr MatrixAddr() : row_(invalid_index), col_(invalid_index) {}
+
+  explicit constexpr MatrixAddr(uint8_t row, uint8_t col) : row_(row), col_(col) {}
+
+  explicit constexpr MatrixAddr(uint8_t offset)
+    : MatrixAddr(offset / cols, offset % cols) {}
+
+  template<typename MatrixAddr__>
+  constexpr MatrixAddr(const MatrixAddr__ &other) : MatrixAddr(other.row(), other.col()) {}
+
+  constexpr uint8_t row() const {
+    return row_;
+  }
+  constexpr uint8_t col() const {
+    return col_;
+  }
+
+  void setRow(uint8_t r) {
+    row_ = r;
+  }
+  void setCol(uint8_t c) {
+    col_ = c;
+  }
+
+  constexpr uint8_t offset() {
+    return row_ * cols + col_;
+  }
+
+  constexpr bool isValid() const {
+    return (row_ < rows__) && (col_ < cols__);
+  }
+
+  template<typename MatrixAddr__>
+  MatrixAddr& operator=(const MatrixAddr__ &other) {
+    this->row_ = other.row();
+    this->col_ = other.col();
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  MatrixAddr &operator+(const MatrixAddr__ &other) {
+    this->row_ += other.row();
+    this->col_ += other.col();
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  MatrixAddr &operator-(const MatrixAddr__ &other) {
+    this->row_ -= other.row();
+    this->col_ -= other.col();
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  MatrixAddr &operator+=(const MatrixAddr__ &other) {
+    *this = *this + other;
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  MatrixAddr &operator-=(const MatrixAddr__ &other) {
+    *this = *this - other;
+    return *this;
+  }
+};
+
+// This is a copressed storage version of MatrixAddr.
+// It uses a common integer to store row and column indices.
+// To enable fast row/col access, bitshifts are used instead of
+// integer divisions to compute row indices. This is only possible
+// by padding rows with unused entries to gain a row size that
+// is a power of two.
+//
+template<uint8_t rows__, uint8_t cols__, typename StorageType__ = uint8_t>
+class MatrixAddrCompressed {
+ private:
+
+  StorageType__ raw_;
+
+ public:
+
+  typedef MatrixAddrCompressed<rows__, cols__, StorageType__> ThisType;
+
+  // Row size must be a power of two to enable fast index extraction.
+  static constexpr uint8_t row_size
+    = internal::matrix_addr::upperPowerOfTwo(cols__);
+
+  static constexpr StorageType__ max_raw_value
+    = internal::matrix_addr::IntLimitsTrait<StorageType__>::max;
+  static constexpr StorageType__ invalid_state = max_raw_value;
+  static constexpr uint8_t shift = internal::matrix_addr::log2(row_size);
+
+  static constexpr uint8_t rows = rows__;
+  static constexpr uint8_t cols = cols__;
+
+  static constexpr uint8_t matrix_size = rows__ * cols__;
+
+  static constexpr StorageType__ rowMask = StorageType__(max_raw_value << shift);
+  static constexpr StorageType__ colMask = StorageType__(~rowMask);
+
+  static_assert(rows * row_size < max_raw_value, "Please use a StorageType__ with a higher capacity");
+
+  constexpr MatrixAddrCompressed() : raw_(invalid_state) {}
+
+  explicit constexpr MatrixAddrCompressed(uint8_t row, uint8_t col) : raw_(row * row_size + col) {}
+
+  explicit constexpr MatrixAddrCompressed(uint8_t offset)
+    : MatrixAddrCompressed(offset / cols, offset % cols) {}
+
+  constexpr MatrixAddrCompressed(const ThisType &other) : raw_(other.raw_) {}
+
+  template<typename MatrixAddr__>
+  constexpr MatrixAddrCompressed(const MatrixAddr__ &other)
+    : MatrixAddrCompressed(other.row(), other.col()) {}
+
+  constexpr uint8_t row() const {
+    return raw_ >> shift;
+  }
+  constexpr uint8_t col() const {
+    return raw_ & colMask;
+  }
+
+  void setRow(uint8_t r) {
+    raw_ = (r << shift) | (raw_ & colMask);
+  }
+  void setCol(uint8_t c) {
+    raw_ = (raw_ & rowMask) | c;
+  }
+
+  constexpr uint8_t offset() {
+    return row() * cols + col();
+  }
+
+  constexpr bool isValid() const {
+    return (row() < rows__) && (col() < cols__);
+  }
+
+  ThisType &operator=(const ThisType &other) {
+    raw_ = other.raw_;
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  ThisType& operator=(const MatrixAddr__ &other) {
+    raw_ = other.row() * row_size + other.col();
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  ThisType &operator+(const MatrixAddr__ &other) {
+    *this = ThisType(this->row_ + other.row(),
+                     this->col_ + other.col());
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  ThisType &operator-(const MatrixAddr__ &other) {
+    *this = ThisType(this->row_ - other.row(),
+                     this->col_ - other.col());
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  ThisType &operator+=(const MatrixAddr__ &other) {
+    *this = *this + other;
+    return *this;
+  }
+
+  template<typename MatrixAddr__>
+  ThisType &operator-=(const MatrixAddr__ &other) {
+    *this = *this - other;
+    return *this;
+  }
+};
+
+// Comparison operators only used for module testing. This is because they
+// cause ambiguous symbol lookup when used in the regular firmware.
+//
+// To use them also for the regular firmware they would need to be
+// disambiguated by moving them to the Kaleidoscope.h header and replacing
+// them with non-template versions that operate on the actual typedefed
+// KeyAddr and LEDAddr.
+
+#ifdef MATRIX_ADDR_TESTING
+
+template<typename MatrixAddr1__, typename MatrixAddr2__>
+bool operator==(const MatrixAddr1__&a1, const MatrixAddr2__ &a2) {
+  return (a1.row() == a2.row()) && (a1.col() == a2.col());
+}
+template<typename MatrixAddr1__, typename MatrixAddr2__>
+bool operator!=(const MatrixAddr1__&a1, const MatrixAddr2__ &a2) {
+  return !operator==(a1, a2);
+}
+
+template<typename MatrixAddr1__, typename MatrixAddr2__>
+bool operator>(const MatrixAddr1__&a1, const MatrixAddr2__ &a2) {
+  return (a1.row() > a2.row())
+         || ((a1.row() == a2.row()) && (a1.col() > a2.col()));
+}
+
+template<typename MatrixAddr1__, typename MatrixAddr2__>
+bool operator<(const MatrixAddr1__&a1, const MatrixAddr2__ &a2) {
+  // This could be optimized if necessary
+  return !operator>(a1, a2) && !operator==(a1, a2);
+}
+
+template<typename MatrixAddr1__, typename MatrixAddr2__>
+bool operator>=(const MatrixAddr1__&a1, const MatrixAddr2__ &a2) {
+  // This could be optimized if necessary
+  return operator>(a1, a2) || operator==(a1, a2);
+}
+
+template<typename MatrixAddr1__, typename MatrixAddr2__>
+bool operator<=(const MatrixAddr1__&a1, const MatrixAddr2__ &a2) {
+  return !operator>(a1, a2);
+}
+
+#endif
+
+} // end namespace kaleidoscope
+
+// Row/col based access functions have been superseded by matrix address
+// base access. To deprecate any row/col based access functions, toggle
+// the definitions of macro KS_ROW_COL_FUNC below
+// and uncomment _DEPRECATED_MESSAGE_ROW_COL_FUNC.
+//
+#define KS_ROW_COL_FUNC
+
+#if 0
+#define _DEPRECATED_MESSAGE_ROW_COL_FUNC \
+   "Row/col based access functions have been deprecated. Please use " \
+   "the KeyAddr/LEDAddr based versions instead."
+#define KS_ROW_COL_FUNC DEPRECATED(ROW_COL_FUNC)
+#endif
--- a/src/kaleidoscope/MatrixAddr.h
+++ b/src/kaleidoscope/MatrixAddr.h
@@ -20,32 +20,6 @@
 #include <stdint.h>
 
 namespace kaleidoscope {
-   
-namespace internal {
-   
-template<typename T1, typename T2>
-struct IsSame {
-   static constexpr bool value = false;
-};
-
-template<typename T>
-struct IsSame<T, T> {
-   static constexpr bool value = true;
-};
-
-template<typename T>
-struct RemoveReference {
-   typedef T type;
-};
-
-template<typename T>
-struct RemoveReference<T&> {
-   typedef T type;
-};
-
-template<bool B> struct Bool2Type {};
-
-} // end namespace internal
 
 template<uint8_t rows__, uint8_t cols__>
 class MatrixAddr {
@@ -76,8 +50,20 @@ class MatrixAddr {
   constexpr MatrixAddr(uint8_t offset)
     : offset_(offset) {}
 
-  constexpr MatrixAddr(const ThisType &other) : offset_(other.offset_) {}
-  constexpr MatrixAddr(ThisType &&other) : offset_(other.offset_) {}
+  // Rely on the default copy and move constructor.
+  //
+  // Note: If these were implemented naively as the commented versions below,
+  //       some versions of avr-gcc (e.g. 4.9.2 or 5.4 would generate
+  //       ridiculously bad assembler code for each copy construction,
+  //       that would bloat the default firmware by 1K of PROGMEM!
+  //
+  constexpr MatrixAddr(const ThisType &other) = default;
+  constexpr MatrixAddr(ThisType &&other) = default;
+  //constexpr MatrixAddr(const ThisType &other) : offset_(other.offset_) {}
+  //constexpr MatrixAddr(ThisType &&other) : offset_(other.offset_) {}
+
+  ThisType &operator=(const ThisType &) = default;
+  ThisType &operator=(ThisType &&) = default;
 
   template<typename MatrixAddr__>
   constexpr MatrixAddr(const MatrixAddr__ &other)
@@ -99,21 +85,36 @@ class MatrixAddr {
     offset_ = this->row() * cols + c;
   }
 
+  void shift(int8_t rows, int8_t cols) {
+    offset_ += rows * cols + cols;
+  }
+
+  void rowShift(int8_t rows) {
+    offset_ += rows * cols;
+  }
+
+  void colShift(int8_t cols) {
+    offset_ += cols;
+  }
+
+  constexpr ThisType shifted(int8_t rows, int8_t cols) const {
+    return ThisType(uint8_t(offset_ + rows * cols + cols));
+  }
+
+  constexpr ThisType rowShifted(int8_t rows) const {
+    return ThisType(uint8_t(offset_ + rows * cols));
+  }
+
+  constexpr ThisType colShifted(int8_t cols) const {
+    return ThisType(uint8_t(offset_ + cols));
+  }
+
   constexpr uint8_t toInt() const {
     return offset_;
   }
 
-//   constexpr operator uint8_t() {
-//     return offset_;
-//   }
-
   constexpr bool isValid() const {
     return offset_ < upper_limit;
-  }
-
-  ThisType &operator=(const ThisType &other) {
-    offset_ = other.offset_;
-    return *this;
   }
 
   ThisType operator++() {
@@ -136,65 +137,6 @@ class MatrixAddr {
     ThisType copy(*this);
     --*this;         // call the prefix increment
     return copy;
-  }
-  
-  // The following templates are necessary as we cannot have both,
-  // a templated and a non templated version of operator=.
-  // Because of this, we have to take a detour over auxiliary
-  // assign methods, one that enables to assign the offset_ member directly,
-  // in case a compatible MatrixAddr class is used and 
-  // one that does a row/col conversion otherwise.
-  
-  template<uint8_t xrows__, uint8_t xcols__>
-  static void assign(internal::Bool2Type<true>, ThisType &self, const MatrixAddr<xrows__, xcols__> &other) {
-    self.offset_ = other.offset_;
-  }
-      
-  template<uint8_t xrows__, uint8_t xcols__>
-  static void assign(internal::Bool2Type<false>, ThisType &self, const MatrixAddr<xrows__, xcols__> &other) {
-    self.offset_ = other.row() * cols + other.col();
-  }
-  
-  // Normal assignment
-  //
-  template<typename MatrixAddr__>
-  ThisType& operator=(const MatrixAddr__ & other) {
-     assign(
-        // A dummy value that is only passed to make the compiler select
-        // the right overload of assign(...). It will be optimized away
-        // by the compiler, i.e. not be put on the stack when
-        // calling the function as the class IsSame is empty.
-        //
-        internal::Bool2Type<
-           internal::IsSame<
-              ThisType, 
-              typename internal::RemoveReference<MatrixAddr__>::type
-           >::value
-        >(), 
-        *this, other
-     );
-    return *this;
-  }  
-  
-  // Move assignment
-  //
-  template<typename MatrixAddr__>
-  ThisType& operator=(MatrixAddr__ && other) {
-     assign(
-        // A dummy value that is only passed to make the compiler select
-        // the right overload of assign(...). It will be optimized away
-        // by the compiler, i.e. not be put on the stack when
-        // calling the function as the class IsSame is empty.
-        //
-        internal::Bool2Type<
-           internal::IsSame<
-              ThisType, 
-              typename internal::RemoveReference<MatrixAddr__>::type
-           >::value
-        >(), 
-        *this, other
-     );
-    return *this;
   }
 
   template<typename MatrixAddr__>
@@ -225,52 +167,23 @@ class MatrixAddr {
 
   bool operator==(const ThisType &other) const {
     return offset_ == other.offset_;
-  }  
-  
-  template<typename MatrixAddr__>
-  bool operator==(const MatrixAddr__ & other) {
-     return   (this->row() == other.row())
-           && (this->col() == other.col());
   }
 
-  class forward_iterator {
-
-    ThisType addr_;
-
-   public:
-
-    constexpr explicit forward_iterator(ThisType addr) : addr_(addr) {}
-
-    forward_iterator operator++() {
-      ++addr_;
-      return *this;
-    }
-
-    forward_iterator operator++(int) { // postfix ++
-      forward_iterator copy(*this);
-      ++*this;         // call the prefix increment
-      return copy;
-    }
-
-    ThisType operator*() {
-      return addr_;
-    }
-
-    bool operator==(const forward_iterator &other) {
-      return addr_ == other.addr_;
-    }
-    bool operator!=(const forward_iterator &other) {
-      return !this->operator==(other);
-    }
-  };
-
-  typedef forward_iterator iterator;
-
-  forward_iterator begin() const {
-    return forward_iterator(ThisType(uint8_t(0)));
+  bool operator!=(const ThisType &other) const {
+    return offset_ != other.offset_;
   }
-  forward_iterator end() const {
-    return forward_iterator(ThisType(upper_limit));
+
+  const ThisType &operator*() const {
+    return *this;
+  }
+
+  typedef ThisType iterator;
+
+  ThisType begin() const {
+    return ThisType(uint8_t(0));
+  }
+  ThisType end() const {
+    return ThisType(upper_limit);
   }
 };
 
--- a/src/kaleidoscope/MatrixAddr.h
+++ b/src/kaleidoscope/MatrixAddr.h
@@ -280,7 +280,7 @@ class MatrixAddr {
 // To use them also for the regular firmware they would need to be
 // disambiguated by moving them to the Kaleidoscope.h header and replacing
 // them with non-template versions that operate on the actual typedefed
-// KeyAddr and LEDAddr.
+// KeyAddr and KeyLEDAddr.
 
 #ifdef MATRIX_ADDR_TESTING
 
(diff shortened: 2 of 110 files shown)


noseglasses:
Travis failde with the Atreus build. I will try to fix this...

noseglasses:
`make smoke-examples` is currently running... wow that takes pretty long. How cool would it be for a future CLI build system to support parallel builds.

src/kaleidoscope/Hardware.h
```diff
@@ -55,6 +57,12 @@ namespace kaleidoscope {
  */
 class Hardware {
  public:
+
+  // To satisfy the interface of those methods that allow
+  // for matrix addressing we define default key and led address classes.
+  // Those typedefs are supposed to overridden by derived hardware classes.
+  typedef MatrixAddr<0, 0> KeyAddr;
+  typedef MatrixAddr<0, 0> LEDAddr;
```
gedankenexperimenter: I believe the style guide prefers `LedAddr` to `LEDAddr`.

src/kaleidoscope/Hardware.h
```diff
@@ -64,6 +72,16 @@ class Hardware {
...
+  /**
+   * Set the color of a per-key LED at a given row and column.
+   *
+   * Setting the color does not need to take effect immediately, it can be
+   * delayed until @ref syncLeds is called.
+   *
+   * @param ledAddr is the matrix address of the LED.
+   * @param color is the color to set the LED to.
+   */
+  void setCrgbAt(LEDAddr ledAddr, cRGB color) {}
```
gedankenexperimenter: Style guide again: `led_addr`, not `ledAddr` for variables.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -0,0 +1,385 @@
...
+// base access. To deprecate any row/col based access functions, toggle
+// the definitions of macro KS_ROW_COL_FUNC below
+// and uncomment _DEPRECATED_MESSAGE_ROW_COL_FUNC.
+//
+#define KS_ROW_COL_FUNC
+
+#if 0
+#define _DEPRECATED_MESSAGE_ROW_COL_FUNC \
+   "Row/col based access functions have been deprecated. Please use " \
+   "the KeyAttr/LEDAttr based versions instead."
```
gedankenexperimenter: I believe you meant "KeyAddr/LEDAddr" (or "LedAddr") instead of "*Attr".

src/kaleidoscope/addr.h
```diff
@@ -25,19 +25,19 @@
 // as the keyboard has fewer than 256 keys.
```
gedankenexperimenter: I would expect that it would be possible to simply remove this header file entirely. It came from the Qukeys plugin, and I don't know of anything else that uses it – almost certainly there are no third-party plugins that do so, as it's only been accessible here for a few months.

src/kaleidoscope/MatrixAddr.h
noseglasses: Exactly.

src/kaleidoscope/addr.h
noseglasses: I thought so, too but as there is a small chance that someone might rely on it, I didn't dare to remove it.

noseglasses:
The class templates could be turned into ordinary classes without any impact on the remaining changes (only the hardware headers would require changes). I templated `KeyAddrCompressed` because I was not completely sure if there wasn't anything that has matrix character but more than 255 entries. Mind you that both address classes could also be used locally for other purposes.

src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -61,7 +60,9 @@ namespace plugin {
 struct Qukey {
  public:
   Qukey(void) {}
-  Qukey(int8_t layer, byte row, byte col, Key alt_keycode);
+  Qukey(int8_t layer, KeyAddr keyAddr, Key alt_keycode);
+  KS_ROW_COL_FUNC Qukey(int8_t layer, byte row, byte col, Key alt_keycode)
+    : Qukey(layer, KeyAddr(row, col), alt_keycode) {}
 
   int8_t layer;
   uint8_t addr;
```
gedankenexperimenter: This `addr` member should change from `uint8_t` to `KeyAddr` if this PR is adopted. The `addr` is meant to be a single-value identifier for a particular key, after all, and was only used because there was no pre-existing version of that when Qukeys was first written.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -0,0 +1,385 @@
...
+//
+template<uint8_t rows__, uint8_t cols__, typename StorageType__ = uint8_t>
+class MatrixAddrCompressed {
+ private:
+
+  StorageType__ raw_;
+
+ public:
+
+  typedef MatrixAddrCompressed<rows__, cols__, StorageType__> ThisType;
```
gedankenexperimenter: The code here is all very inscrutable. Some comments with examples (`StorageType_` is very unlikely to ever be anything other than `uint8_t`) would really help anyone who's trying to understand what you're doing here.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -0,0 +1,385 @@
...
+
+  // Row size must be a power of two to enable fast index extraction.
+  static constexpr uint8_t rowSize
```
gedankenexperimenter: Again, the [coding style guide specifies lower case for variable names](https://github.com/keyboardio/Kaleidoscope/blob/master/doc/CODING_STYLE.md#variable-names). I'm finding it much more difficult than necessary reading this code where variables are using the same style as function names.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  void setRow(uint8_t r) {
+    raw_ = (r << shift) | (raw_ & colMask);
+  }
+  void setCol(uint8_t c) {
+    raw_ = (raw_ & rowMask) | c;
+  }
+
+  constexpr uint8_t offset() {
```
gedankenexperimenter: I find `offset()` to be unintuitive, and this is the function I would need when using a bitfield with one bit per key on the keyboard. I think `index()` would be a much better name.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -0,0 +1,385 @@
...
+    return row() * cols + col();
+  }
+
+  constexpr bool isValid() const {
+    return (row() < rows__) && (col() < cols__);
```
gedankenexperimenter: Why continue basing everything on row and column here? Why not just use a single number for the first invalid key index? Also, what if a keyboard with holes in the matrix has some reason to guarantee that `isValid()` will only return true for matrix addresses that correspond to actual keyswitches?

src/kaleidoscope/addr.h
gedankenexperimenter: I would encourage @algernon and @obra to get rid of it right away if they adopt this PR, since it is attempting to serve the same purpose as the thing that should be used instead.

gedankenexperimenter:
@noseglasses – apologies for the random mix of detail comments; I haven't got the time to study and write up anything systematically.

gedankenexperimenter:
One other comment – I just went and checked to see which operators I had defined for `KeyAddr` in Kaleidoglyph, and found that I use `operator==` quite a bit, and also increment and decrement operators. Those were approximately essential for making simple for loops.

src/kaleidoscope/MatrixAddr.h
noseglasses: As I can't look into the future, I can't foresee if there might be applications with more rows or columns than 255.

src/kaleidoscope/MatrixAddr.h
noseglasses: I both agree and disagree. Index is not unique as the term is also used as row- or column-index.
In the context of the firmware the term seems to have a special meaning. There are quite a lot of functions and methods that accept a one-based parameter called index. There the value zero has a special meaning.

obra:
I would not reasonably expect us to support more than 255 keys on an AVR. I
haven't had cycles to dig into the implementation as yet (and you've likely
handled this), but I could see situations where we have wildly varying
numbers of rows and columns between different hardware. I wouldn't want to
hardcode how many bits get used for rows and how many bits get used for
columns.


ᐧ

obra:
@noseglasses: I'd love to see some size numbers with the backward compatibility removed. I understand that there will be some bloat during a transition, but how will things look once everything's shaken out?

src/kaleidoscope/MatrixAddr.h
noseglasses: I have no intention to move away from row/column indexing with this PR. Let's keep things separate.

`isValid` could be easily be renamed `isInRange`. In no placer does it replace anything related to holes.

src/kaleidoscope/plugin/Qukeys.h
noseglasses: No, this PR does not intent to mess with any algorithms. Such plugin interna could be delt with by separate PRs.

noseglasses:
> but I could see situations where we have wildly varying numbers of rows and columns between different hardware. I wouldn't want to hardcode how many bits get used for rows and how many bits get used for columns. ᐧ

That's the whole point of this (templated) implementation, that every hardware decides by itself about the number of bits. But only when the compressed storage version of the address classes would be used. The default `MatrixAddr` does not fiddle with bits in any sort of way.

noseglasses:
> @noseglasses: I'd love to see some size numbers with the backward compatibility removed. I understand that there will be some bloat during a transition, but how will things look once everything's shaken out?

Good point. I will see what I can do. I will take the default M01 firmware as a reference.

src/kaleidoscope/MatrixAddr.h
gedankenexperimenter: The term `offset` implies to me that it is a constant, not something that would vary from one `KeyAddr` to another.

noseglasses: I am not insisting on the naming `offset`. Is there maybe something else apart from `index` that we might name it?

noseglasses:
And here are the numbers, taking the stock firmware as a reference:

| Build | Program `MatrixAddr` | Data | Program `MatrixAddrCompressed` | Data |
|---|---|---|---|---|
| vanilla Kaleidoscope-Bundle-Keyboardio | **25220** | 1414 |  n.a.  | n.a.  |
| with matrix addressing and wrappers | 25268 | 1432 | 26388 | 1428 |
| only `onKeyswitchEvent` removed (wrappers in place) | 25242 | 1432 | 26346 | 1428 |
| `onKeyswitchEvent` and all wrappers removed | **25196** | 1416 | 26076 | 1410 |

The total difference in terms of PROGMEM is marginal but using the `MatrixAddr` version we will eventually even save some bytes.

noseglasses:
While i did the research on PROGMEM consumption, it turned out that in some places the wrappers are still called. This is because in some places the variable names `r` and `c` are used for rows and columns and in very rare cases even others. I will fix the `r`/`c` cases using the search regex `\br\s*,\s*c\b`. The very few remaining calls to wrappers i will leave as is. They will cause compiler errors once the wrappers have been removed but those are quite easy to fix.

src/kaleidoscope/Hardware.h
noseglasses: I stick with `LEDAddr` analogous to e.g. `LEDControl`. Besides, LED is an acronym, thus IMHO every letter deserves its capital. But that corner case is not defined in the stylegide :wink:.

src/kaleidoscope/Hardware.h
noseglasses: Thanks, I'll fix that.

src/kaleidoscope/MatrixAddr.h
noseglasses: Fixed.

src/kaleidoscope/Hardware.h
noseglasses: Fixed.

src/kaleidoscope/addr.h
noseglasses: I'll remove it as soon as A&O approve it.

src/kaleidoscope/MatrixAddr.h
noseglasses: My experience with many other C++ projects is that a mixture of camel and snake case for symbol names as demanded by the Kaleidoscope style guide is rarely used.

Anyway, i adapted all variable and parameter names to be conforming with the style guide.

noseglasses:
I just pushed some commits that are related to code-comments above.

src/kaleidoscope/MatrixAddr.h
noseglasses: Some thoughts about linear indexing: 

There are many places in the firmware where linear indexing instead of two dimensional indexing would likely be more efficient. Nested for loops could be merged into one loop and index computations for lookup in two dimensional arrays could be saved. I think I have even seen places where a row/col pair is converted to a linear index and then back to row/col before it is actually used. That's probably what you have in mind with your linear design.

But this is a keyboard firmware and AFAIK most keyboards come with some sort of matrix hardware. That's why row/col indices make some sense to me. Yet I am not sure if they need to be exposed to the user as part of interfaces. Dealing with row/col indexing could as well be restricted to the hardware classes and the keyscanner code.

Most keyboards - except for some special designs like the Planck - have a weird indexing scheme anyway that makes it impossible for the user to remember the row/col values of specific keys. On such devices linear indexing would not be worse than row/col indexing because when a specific key s supposed to be addressed, the user must look it up in [some sort of table](https://github.com/keyboardio/Kaleidoscope-Hardware-Model01/blob/f469015346535cb864a340bf8eb317d268943248/src/Kaleidoscope-Hardware-Model01.h#L267-L279) anyway.

Nevertheless, I favor evolution over revolution and that's why I would like to see small changes - and this PR is a first step. Based on its changes it will be easier to switch to a linear indexing if the core designers would agree on that.

src/kaleidoscope/MatrixAddr.h
noseglasses: BTW, I don't like snake case because it breaks the ability to rely on case sensitivity when searching for symbol names. 

For search strings that start with a capital letters, with camel case you can always be sure that you will only get hits at the beginning of a word, e.g. searching for `Bus(` would find `readBus(...)` or similar things, while searching for `bus` in the presence of snake case variables also gives you all hits that end with the syllable bus like `find_succubus(`. That's no problem in an environment where advanced search features like regex are available but it makes a huge difference when browsing code in other places where no such features are available, like e.g. web browsers (github).

algernon:
I've yet to review the PR in full (sorry!), but I like the idea in general. Getting away from rows/cols is a good goal. Looking at the size comparsion, I don't think the compressed variant is useful. Compared to `MatrixAddr`, it saves 6 bytes of RAM on the Model01, but PROGMEM is larger by over 800 bytes. I don't think that's a good trade, and I'd rather "waste" 6 bytes of RAM than almost a kilobyte of PROGMEM.

Regarding support for more than 255 keys, I think we should be fine with being limited to 255. We're not going to support keyboards with more keys anytime soon, so there's no harm in having that limit right now. I'm pretty sure we have plenty of other places throughout the codebase that'd break in a major way when faced with a >255 key board.

obra:
One of the things @algernon and I have been discussing is entirely ditching matrix addressing for linear addressing for the logical locations of both keys and LEDs.  If we accept that we'd max out at 255 of each on an AVR, that means we could "just" use a single int.

We'd still need the compat layer kicking around for a while, but I feel like that might achieve what this PR is doing with less complexity. But maybe I'm just being naive and optimistic :)

noseglasses:
Yeah, linear addressing would definitely be favorable. But I would still go with classes for key and LED addresses. Those would still be useful to wrap the code for those cases where row/col values are to be encoded or decoded. They would still be just one byte but be much more convenient with respect to their handling. And don't forget about the debugging features that come with accessors...at zero runtime costs.

Most of the changes of this PR are just replacing the row/col values in function signatures and calls. The part where this PR would become truly less complex is the MatrixAddr header. There, the second of my two implementations of address classes could be un-templated and would then be exactly what we would need for linear addressing.

Do you want me to adapt/simplify this PR in that direction?

obra:
I -think- I do.

gedankenexperimenter:
I've been using classes and linear addressing for `KeyAddr` & `LedAddr` types (one bytes each) in Kaleidoglyph happily for some time. Heartily recommended.

noseglasses:
I am going to do this stepwise. First I force-pushed an update of the current changes that brings it in sync with the changes on origin/master. Next i will do the transition to global linear addressing. The latter means that in all functions that finally decode the `KeyAddr`s/`LEDAddr`s with `row()` and `col()` changes will be required.

noseglasses:
While working on this PR I was confronted with the LED indices that are necessary to support individual LED wiring. Currently the hardware classes expose this detail via methods like e.g. `void setCrgbAt(int8_t i, cRGB crgb)'. Isn't this something that we better encapsulate? `LEDControl` has such methods.

noseglasses:
With the latest commit, linear addressing is used in most parts of the code (except for the keymap access in the layer class that is fixed by another PR currently in progress).

See the commit description for infos on the changes.

noseglasses:
Sorry, I thought that this is done but when I ran a PROGMEM/RAM report over this I noticed i needs some serious tweaking before it could be merged.

Setting back to WIP.

noseglasses:
> While working on this PR I was confronted with the LED indices that are necessary to support individual LED wiring. Currently the hardware classes expose this detail via methods like e.g. void setCrgbAt(int8_t i, cRGB crgb). Isn't this something that we better encapsulate? LEDControl has such methods.

I think, I can now answer my own question. When iterating over all LEDs, it can save some overhead to work with LED ids instead of LEDAddr to save the mapping from LEDAddr to LED ids.

noseglasses:
... for some keyboards like the Model01 this mapping involves a lookup in a PROGMEM stored array, which makes it even more costly.

gedankenexperimenter:
> I think, I can now answer my own question. When iterating over all LEDs, it can save some overhead to work with LED ids instead of LEDAddr to save the mapping from LEDAddr to LED ids.

On the Model01, there are at least two LED modes that use the LED addresses directly for a different reason. Both the Chase and RainbowWave effects cycle through the LED addresses because that's how they look right.

noseglasses:
Yeah, I learned that when I temporarily screwed up those effects. RainbowWave looks cool, though, if displayed transposed :smirk: 

There are also other places where a loop over all LEDs can be found. E.g. to zero them. Then the mapping from LEDAddr is not useful.

noseglasses:
Alright, it took me quite a while to find out why the firmware program memory consumption went up by __~600 bytes__ instead of going down as one would expect, after the transition from row/col indexing to linear indexing.

The problem seems to be related to a bug in gcc. It manifests at least on the avr versions up to gcc v5.4 and causes the generation of very inefficient assembler code.

Changing one line in MatrixAddr.h from
```cpp
MatrixAddr(const MatrixAddr &other) : offset_(other.offset_) {}
```
to
```cpp
MatrixAddr(const MatrixAddr &other) = default;
```
saves __~1000 bytes__ of PROGMEM. Try it yourself, if you don't believe that. i had to play around randomly with Compiler Explorer for two days until I found the culprit. i added comments to the code that explain the matter.

Now we are at __-460 bytes__ of PROGMEM and - because of some temporarily introduced virtual methods, whose original versions will eventually be removed - we are temporarily up by __18 bytes__ of RAM.

Once the deprecated stuff will be removed, we will save __>478 bytes__ of PROGMEM and __14 bytes__ of RAM. This demonstrates that using linear addressing really pays off.

The newly introduced range based for loops also make parts of the code more readable.

See the memory report: [kaleidoscope_change_report_linear_addressing.pdf](https://github.com/keyboardio/Kaleidoscope/files/3030018/kaleidoscope_change_report_linear_addressing.pdf)

obra:
I've been reading the code for this PR again this afternoon, with an eye toward mergeability.

### In general 

Thanks for working on this! I'm looking forward to getting the featureset merged.


### onKeyswitchEvent2

While I understand that versioned method names like `onKeyswitchEvent2` is a common practice in C++ code, I really, really dislike it. I find that it generally makes the wrong tradeoff between backward compatibility and future cognitive load. I believe @noseglasses' assertion that we can't easily support both signatures with the same hook name, though don't quite understand the comment that we want to prevent users from messing with signatures in derived plugins.

### KS_ROW_COL_FUNC

Stylistically,  I think I prefer DEPRECATED(ROW_COL_FUNC) to KS_ROW_COL_FUNC because it makes it clearer what's going on when reading the code. 

### UNKNOWN_KEYSWITCH_LOCATION

The change from UNKNOWN_KEYSWITCH_LOCATION  to UnknownKeyswitchLocation should be a standalone commit.

### TapDance.md

In TapDance.md, it looks like there's an editing error, where you replaced var names with a type:

 
-### `tapDanceAction(tap_dance_index, row, col, tap_count, tap_dance_action)`
+### `tapDanceAction(tap_dance_index, keyAddr, tap_count, tap_dance_action)`

### KeyAddr{}
In Kaleidoscope.cpp,

+  for (auto key_addr : KeyAddr{}) {
+    Layer.updateLiveCompositeKeymap(key_addr);

This is me being inexperienced, but I don't recognize what KeyAddr{} is supposed to do. Would you mind adding a line of comment?

### num_keys naming

in src/kaleidoscope/Kaleidoscope.h:

+static constexpr uint8_t num_keys =  KeyboardHardware.matrix_rows
+                                     * KeyboardHardware.matrix_columns;

I think we want a more descriptive name here, since there are so many things that we might mean by num_keys across kaleidoscope.

Eventually, I think we are going to want to fully break apart the size of the hardware matrix from the number of logical keys. Especially in the case of a sparse matrix, we may be able to get significant savings from linear addressing by being able to compress the size of the keyboard rather than leaving gaps to match the hardware matrix's gaps.

With that in mind, um. num_logical_keys logical_key_count? num_keyswitches? keyswitch_count? num_switches_on_keyboard? keyboard_keyswitch_count? None of these are something I love, but num_keys just feels too generic. Also, I think we may want to push it down to be a method of the hardware. If it's actually offered up by the hardware class, then KeyboardHardware.num_keys() feels ok.

### num_keys usage

src/kaleidoscope/plugin/Heatmap.cpp

It's still using ROWS * COLS

Doing some greps for ROWS and COLS, it does look like there are other places we're not using your new num_keys.

(Ideally, all the changes to num_keys would be done as one standalone commit.)

### Commented code

There are a couple places where you comment out code rather than deleting it or add new code with commented code inline. Generally, that's to be avoided

### Breaking apart commits

Generally for merge, we like to have each logical change or set of the same change broken into its own commit. That makes it much easier to read the history or figure out which change broke something. While working on a PR, it's ok to push updates, but before merge, we should tease things back apart.

### ActiveModColor

With code like this 
```OneShot.isOneShotKey(k)) {
       if (::OneShot.isSticky(k))
-        ::LEDControl.setCrgbAt(r, c, sticky_color);
+        ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), sticky_color);
       else if (::OneShot.isActive(k))
-        ::LEDControl.setCrgbAt(r, c, highlight_color);
+        ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), highlight_color);
       else
-        ::LEDControl.refreshAt(r, c);
+        ::LEDControl.refreshAt(KeyLEDAddr(key_addr));
     } else if (k.raw >= Key_LeftControl.raw && k.raw <= Key_RightGui.raw) {
       if (hid::isModifierKeyActive(k))
-        ::LEDControl.setCrgbAt(r, c, highlight_color);
+        ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), highlight_color);
       else
-        ::LEDControl.refreshAt(r, c);
+        ::LEDControl.refreshAt(KeyLEDAddr(key_addr));
     } else if (k.flags == (SYNTHETIC
```
Is it more space efficient to instantiate KeyLEDAddr inside all those branches or would it make sense to create it once outside the branches?


### AlphaSquare

```+  }
   static void display(uint16_t symbol) {
-    display(symbol, 0, 2);
+    display(symbol, KeyLEDAddr{0, 2});
   }
   static void display(uint16_t symbol, uint8_t col) {
-    display(symbol, 0, col);
+    display(symbol, KeyLEDAddr(0, col));
   }
 ```

More instances of the KeyLEDAddr{} vs KeyLEDAddr() pattern. Is there a reason we want the {} pattern there?

### src/kaleidoscope/plugin/Model01-TestMode.cpp 

key_num to key_id is something that should be its own standalone commit. It adds a bunch of lines to the diff, but isn't really related to the other code changes.

### kaleidoscope/addr.h

Is it the case that this whole file is slated for removal? 

### Commit grouping and ordering

The really big issue we need to overcome is the naming of the hook replacement.
It should be possible to clean up the commits so that the onKeyswitchEvent2 change is on its own and the last thing in the series, after commits for each feature or logical group of features. 



### Abstraction barriers

One thing that this change doesn't fully address (and does not need to address) is really breaking apart the ordered list of keys on the logical keyboard from the physical matrix on the circuit board. For the Model 01, this doesn't really matter because the matrix is fully populated. But for some other keyboards, there are big gaps in the matrix.

Historically, we haven't really differentiated between the logical key list and the physical matrix. This is both due to our abstractions not being good enough. 

This makes me wonder whether anything inside the hardware/ directory should even know about KeyAddrs. 

**Is KeyAddr the Keyswitch Address in in the hardware key matrix or is it the logical key address in our global internal implementation of the list of keys?**

(Does this question make sense? I sort of worry that this is a distinction that only makes sense to me.)

**Comment by gedankenexperimenter** [comment]
> kaleidoscope/addr.h
> Is it the case that this whole file is slated for removal?
That file came from Qukeys, which uses it to accomplish internal single-byte key addressing. I don't think anything else uses it, and Qukeys won't need it either if this PR is adopted, so I think it should be removed.

**Comment by gedankenexperimenter** [comment]
> Historically, we haven't really differentiated between the logical key list and the physical matrix. This is both due to our abstractions not being good enough.
> 
> This makes me wonder whether anything inside the hardware/ directory should even know about KeyAddrs.
>
> Is KeyAddr the Keyswitch Address in in the hardware key matrix or is it the logical key address in our global internal implementation of the list of keys?
> 
> (Does this question make sense? I sort of worry that this is a distinction that only makes sense to me.)


One of the first things I started playing with when I started experimenting with Kaleidoscope code was abstract key addressing divorced from row and column numbers, and it was one of the starting points of my rewrite. `KeyAddr` was probably the first class I wrote for Kaleidoglyph, so I think I have a modicum of expertise in this area. My design choice there was to have the hardware module be wholly responsible for the definition of the `KeyAddr` and `LedAddr` classes, and their mappings to physical keyswitches and LEDs. This allows the hardware module to implement the mapping in a way that makes the most sense for its particular constraints, perhaps compromising between efficiency of the keyswitch scanning and size of the keymap structure. I saw `KeyAddr` as something that would define that abstraction barrier between the physical and logical layers, or would at least be a thin layer on top of that barrier (depending on the hardware module, perhaps). Either way, its definition would naturally belong in the hardware module, so it makes the most sense to me if the `hardware` directory _knows about_ it, even if `KeyAddr` isn't used to address the physical keyswitches.

I am firmly of the opinion that a `KeyAddr` should be (and is, if I understand this PR correctly) the _logical key address_, used by Kaleidoscope and plugins, not the _physical key address_ of the hardware key matrix. It's not that the hardware modules should be ignorant of `KeyAddr`s – it's that everything else should be agnostic about the physical key and LED addresses.

File: src/kaleidoscope/hardware/keyboardio/Model01.h [hunk 10]
```diff
@@ -43,11 +44,20 @@ class Model01 : public kaleidoscope::Hardware {
   static constexpr byte matrix_columns = 16;
   static constexpr int8_t led_count = 64;
 
+  typedef MatrixAddr<matrix_rows, matrix_columns> KeyAddr;
+  typedef MatrixAddr<matrix_rows, matrix_columns> KeyLEDAddr;
```
gedankenexperimenter: I just realized that there's something here that concerns me: `KeyAddr` and `KeyLEDAddr` for the Model01 are the same type, even though they appear to be different types in use (and, for other boards, they are different types). This means that it's not possible to have two polymorphic functions like:
```
int foo(KeyAddr key_addr);
int foo(KeyLEDAddr led_addr);
```
…and it's possible to accidentally pass one type where the other was intended, and get  unexpected results without a compilation error.

noseglasses: This question is connected to the question raised by @obra above, about what a `KeyAddr` or `KeyLEDAddr` actually represents. For me `KeyAddr`s and `KeyLEDAddr`s are both unique ids that the user can use to reference keys or a key-LEDs. The way it is currently implemented, a `KeyAddr` and a `KeyLEDAddr` have identical values if they both reference the same keys. That's not because the two C++ types are identical but because they both replaced the same row/col-pairs. But this very identity raises the question whether we actually need two sets of addresses at all. 

i had introduced `LEDAddr` because i initially thought there was just one set of LEDs that don't necessarily need to be per-key LEDs. Then, obra reminded me that there will be other sets of LEDs coming, like underglow or status LEDs. Because of that i renamed `LEDAddr` to `KeyLEDAddr`. 

My conclusion is: We should go with only a single `KeyAddr` for both keys and key related LEDs. The wiring details should be entirely handled by the hardware classes. For other sets of LEDs we should have additional address-classes like `UnderglowLEDAddr` or `StatusLEDAddr` that we introduce once we need them.

noseglasses:
@obra, thanks for the detailed review.

> onKeyswitchEvent2
> 
> While I understand that versioned method names like onKeyswitchEvent2 is a common practice in C++ code, I really, really dislike it. I find that it generally makes the wrong tradeoff between backward compatibility and future cognitive load. I believe @noseglasses' assertion that we can't easily support both signatures with the same hook name, though don't quite understand the comment that we want to prevent users from messing with signatures in derived plugins.

Could you please point me to the comment you were referring to?

Anyway, whatever the comment was, it appears to not having been very informative. So let explain the problem in greater detail.

`onKeyswitchEvent` is the only function that we cannot simply overload and deprecate like I did it with all the other functions and method that used row/col-pairs as arguments. The reason for this is that we have a template based mechanism that verifies the signature of derived plugins' hook methods. This check is important as we want to make 100% sure that a user does not accidentally re-implement a hook method with a slightly wrong signature that does not get called. That's necessary as we do not use virtual functions here and, thus, cannot rely on the keyword `override` to signal the compiler do the signature check for us.

This verification strategy relies on the hook methods' names, rather than their signatures. Because of that we cannot have two hook-method overloads.

I perfectly agree with you that the name `onKeyswitchEvent2` does not provide a clue to the user about the reason why there are two similar hook methods.

Unfortuatelly, `onKeyswitchEvent` is a very widely used methods in plugins. That's why we would break a lot of user code if just modified the signature without a deprecation period.


> KS_ROW_COL_FUNC
> 
> Stylistically, I think I prefer DEPRECATED(ROW_COL_FUNC) to KS_ROW_COL_FUNC because it makes it clearer what's going on when reading the code.

Yeah, that is more readable and serves the same purpose of being found by grep.

The reason why I generally tend to use prefixes for macros is their lack of namespaces. In larger software projects, they are a frequent source of trouble due to naming collisions.

> UNKNOWN_KEYSWITCH_LOCATION
> 
> The change from UNKNOWN_KEYSWITCH_LOCATION to UnknownKeyswitchLocation should be a standalone commit.

I didn't really want to change it but during the transition period we need two
individual identifiers that represent two different C++ types. The reason why I chose the non all-caps `UnknownKeyswitchLocation` was
that I wanted it to be a `constexpr`. Unfortunately, the compiler for unknown reasons instantiated `UnknownKeyswitchLocation`, thus wasting two bytes of RAM.

> TapDance.md
> 
> In TapDance.md, it looks like there's an editing error, where you replaced var names with a type:
> 
> -### tapDanceAction(tap_dance_index, row, col, tap_count, tap_dance_action)
> +### tapDanceAction(tap_dance_index, keyAddr, tap_count, tap_dance_action)

Almost. `keyAddr`, starting with a lower case letter was meant to be a variable. It's actually a leftover after a global renaming to provide consistency with the coding style guide.

> KeyAddr{}
> 
> In Kaleidoscope.cpp,
> 
>     for (auto key_addr : KeyAddr{}) {
>     Layer.updateLiveCompositeKeymap(key_addr);
> 
> This is me being inexperienced, but I don't recognize what KeyAddr{} is supposed to do. Would you mind adding a line of comment?

`KeyAddr{}` is identical to `KeyAddr()`, both are calls to the default constructor.
C++ introduced the curly brace syntax to disambiguate construction and function calls, which can help to speed up the compile.

Here, this creates an unnamed temporary that is used as a dummy for the range based for loop to query the iteration boundaries. The temporary is later entirely eliminated by the compiler. 

This is equivalent to
```cpp
for(auto it = KeyAddr{}.begin(); it < KeyAddr{}.end(); ++it) {
   auto key_addr = *it;
   ...
}
```

`for (auto key_addr : KeyAddr{})` is a quite frequent idiom in Kaleidoscope, now. Do you want all occurrences to be commented?

> num_keys naming
> 
> in src/kaleidoscope/Kaleidoscope.h:
> 
> +static constexpr uint8_t num_keys = KeyboardHardware.matrix_rows
> 
>                                     * KeyboardHardware.matrix_columns;
> 
> I think we want a more descriptive name here, since there are so many things that we might mean by num_keys across kaleidoscope.
> 
> Eventually, I think we are going to want to fully break apart the size of the hardware matrix from the number of logical keys. Especially in the case of a sparse matrix, we may be able to get significant savings from linear addressing by being able to compress the size of the keyboard rather than leaving gaps to match the hardware matrix's gaps.
> 
> With that in mind, um. num_logical_keys logical_key_count? num_keyswitches? keyswitch_count? num_switches_on_keyboard? keyboard_keyswitch_count? None of these are something I love, but num_keys just feels too generic. Also, I think we may want to push it down to be a method of the hardware. If it's actually offered up by the hardware class, then KeyboardHardware.num_keys() feels ok.

I would opt for `keyswitch_count` if it was left a global variable. But I am a even greater fan of encapsulation. Adding such information to the hardware class instead of to global scope would make perfect sense. Let's have a `numKeys()` method. 

> num_keys usage
> 
> src/kaleidoscope/plugin/Heatmap.cpp
> 
> It's still using ROWS * COLS
> 
> Doing some greps for ROWS and COLS, it does look like there are other places we're not using your new num_keys.
> 
> (Ideally, all the changes to num_keys would be done as one standalone commit.)

Alright, I will replace all occurrences in a separate commit.

> Commented code
> 
> There are a couple places where you comment out code rather than deleting it or add new code with commented code inline. Generally, that's to be avoided

Ok. 

> Breaking apart commits
> 
> Generally for merge, we like to have each logical change or set of the same change broken into its own commit. That makes it much easier to read the history or figure out which change broke something. While working on a PR, it's ok to push updates, but before merge, we should tease things back apart.

Agreed. Once the submitted changes as a whole are agreed upon, I will separate them into logical changes wherever possible.

> ActiveModColor
> 
> With code like this
> 
>        if (::OneShot.isSticky(k))
> -        ::LEDControl.setCrgbAt(r, c, sticky_color);
> +        ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), sticky_color);
>        else if (::OneShot.isActive(k))
> -        ::LEDControl.setCrgbAt(r, c, highlight_color);
> +        ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), highlight_color);
>        else
> -        ::LEDControl.refreshAt(r, c);
> +        ::LEDControl.refreshAt(KeyLEDAddr(key_addr));
>      } else if (k.raw >= Key_LeftControl.raw && k.raw <= Key_RightGui.raw) {
>        if (hid::isModifierKeyActive(k))
> -        ::LEDControl.setCrgbAt(r, c, highlight_color);
> +        ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), highlight_color);
>        else
> -        ::LEDControl.refreshAt(r, c);
> +        ::LEDControl.refreshAt(KeyLEDAddr(key_addr));
>      } else if (k.flags == (SYNTHETIC
> 
> Is it more space efficient to instantiate KeyLEDAddr inside all those branches or would it make sense to create it once outside the branches?

We probably could move it outside of the conditional clauses without degrading performance.

> AlphaSquare
> 
>    static void display(uint16_t symbol) {
> -    display(symbol, 0, 2);
> +    display(symbol, KeyLEDAddr{0, 2});
>    }
>    static void display(uint16_t symbol, uint8_t col) {
> -    display(symbol, 0, col);
> +    display(symbol, KeyLEDAddr(0, col));
>    }
> 
> More instances of the KeyLEDAddr{} vs KeyLEDAddr() pattern. Is there a reason we want the {} pattern there?

Not necessarily. I use the `{}` syntax in any new code I write as it was introduced for good reasons. But I have not 100% adapted so after 11 p.m. I usually start writing `()`. 
But I agree, that it might lead to confusion when it is mixed the way it is here. Will adapt that.

> src/kaleidoscope/plugin/Model01-TestMode.cpp
> 
> key_num to key_id is something that should be its own standalone commit. It adds a bunch of lines to the diff, but isn't really related to the other code changes.
> kaleidoscope/addr.h

We could as well stick with keynum.
 
> Is it the case that this whole file is slated for removal?

@gedankenexperimenter gave its go to remove it. I will do so.

> Commit grouping and ordering
> 
> The really big issue we need to overcome is the naming of the hook replacement.
> It should be possible to clean up the commits so that the onKeyswitchEvent2 change is on its own and the last thing in the series, after commits for each feature or logical group of features.

I assume that you want every individual commit to preserve the integrity of the overall firmware (buildable and runnable). Thus those independent commits might not be as many as you expect as many changes are pretty interrelated. I will see what I can do.

> Abstraction barriers
> 
> One thing that this change doesn't fully address (and does not need to address) is really breaking apart the ordered list of keys on the logical keyboard from the physical matrix on the circuit board. For the Model 01, this doesn't really matter because the matrix is fully populated. But for some other keyboards, there are big gaps in the matrix.
> 
> Historically, we haven't really differentiated between the logical key list and the physical matrix. This is both due to our abstractions not being good enough.

We already had LED row/col addresses and LED indices. So for LEDs there was a differentiation that leaked from the hardware implementation. After this PR we will not need to differentiate this anymore and the mapping between those two addressing types might become obsolete.

Even thought this topic is strongly related to the transition to linear addressing, I did not dare to remove LED indices right away. This is because of the Chase and RainbowWave LED effects whose implementation is somewhat hardware dependent. Those effects will look differently on different types of keyboards. i was not sure where to go here.

The physical matrix setup of keys is something that is mostly interesting for LED effects. But it also matters for e.g. keymaps. My XKeymaps plugin that I hope to release soon, e.g. supports mirroring of keymaps. While only applicable to symmetric keyboards, this requires information about physical key positions.

> 
> This makes me wonder whether anything inside the hardware/ directory should even know about KeyAddrs.
> 
> Is KeyAddr the Keyswitch Address in in the hardware key matrix or is it the logical key address in our global internal implementation of the list of keys?
> 
> (Does this question make sense? I sort of worry that this is a distinction that only makes sense to me.)

This question makes a lot of sense. From the perspective of the user or plugin implementor the `KeyAddr` has at least two independent properties.

1. it uniquely references a key.
2. it provides rough information about the relative position of keys, although it does not grant it.

In both cases the user/plugin implementor does not care for the actual wiring. So the mapping from key position coordinates to electronic circuit coordinates would preferably be abstracted away by the hardware class.

That's why I agree with @gedankenexperimenter that the hardware must know about the `KeyAddr`, to be able to do the mapping. The user, on the other hand, shouldn't need to bother with the internals of the keyboard.

The `KEYMAP` macro is one of the implementation details, where I am not completely decided how I would like to see things handled. We currently use row/col information to reference keys there, but we could as well use the integer value that is stored within the `KeyAddr` class. For some keyboards like the Planck, row/cols make perfect sense. But for most other modern keyboards, including the Model01, the assignment of row/col values to thumb and palm keys reveals internal details that are of no use to the user.

Another problem are the LED effects in general. I assume that most of them have been designed with the Model01 in mind with respect to key-LED positions. That's why some of them might be less portable than others.

For LED effects it would be beneficial to have a x-y LED coordinate system instead of matrix positions. The hardware class could offer a mapping from `KeyAddr` to x-y coordinates to allow for portable LED-effects. The Wavepool effect already implements such a mapping. The related methods of the hardware class would be removed by the linker whenever no plugin uses them.

As a consequence the question arises whether the `KeyAddr` type should expose `row()`/`col()` values at all. Currently we need such methods for the compatibility layer introduced by this PR. But once the deprecated stuff is removed, row/col are only used internally by the hardware class.

One way we could handle a transition would be to declare the leftover compatibility functions/methods/classes that require direct access `row()`/`col()` to be friends of the `MatrixAddr` class. Thus we would prevent any new code from accessing row/col information directly. Then, after the deprecation periode elapsed, we would remove these friend declarations together with anything that reminds of row/col in `MatrixAddr`. The encoding/decoding of `KeyAddr` would then entirely be left to the hardware classes.

I am well aware that this means an additional change to this PR. But from my perspective it makes sense to deal with the abstraction thing right now as we are at it, instead of fixing it later.

What do you think?

src/kaleidoscope/hardware/keyboardio/Model01.h
gedankenexperimenter: Ah, I see. I fell into the trap of assuming that your `KeyLEDAddr` was equivalent to my `LedAddr`, but it's not. I agree with your conclusion — if the same value refers to the same physical position on the keyboard in both cases, it would be better to use just the one name.

gedankenexperimenter:
> `for (auto key_addr : KeyAddr{})` is a quite frequent idiom in Kaleidoscope, now. Do you want all occurrences to be commented?

I think the range-based for loops would be clearer if they were written like this:
```c++
for (KeyAddr k : KeyAddr::Iterator{}) {
  ...
}
```
The addition of the `Iterator` inner type would help guide the reader to its definition in the MatrixAddr.h header, which should include a paragraph comment describing how the iterator works. Yes, it can be written shorter, but then it's less discoverable, and without corresponding duplicated comments everywhere it appears, it will be confusing to those less familiar with modern C++.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -0,0 +1,248 @@
...
+
+  bool operator!=(const ThisType &other) const {
+    return offset_ != other.offset_;
+  }
+
+  const ThisType &operator*() const {
+    return *this;
+  }
+
+  typedef ThisType iterator;
```
gedankenexperimenter: This is a type name, so it should be `Iterator`, with a leading uppercase character. I think there should also be a paragraph comment here describing its purpose, and how the resulting range-based for loops work.

gedankenexperimenter:
> The `KEYMAP` macro is one of the implementation details, where I am not completely decided how I would like to see things handled. We currently use row/col information to reference keys there, but we could as well use the integer value that is stored within the `KeyAddr` class. For some keyboards like the Planck, row/cols make perfect sense. But for most other modern keyboards, including the Model01, the assignment of row/col values to thumb and palm keys reveals internal details that are of no use to the user.

The two innermost columns and the thumb and palm keys have always caused me trouble on the Model01 because of the lack of abstraction. When I look at it, I see a keyboard with 5 rows & two keys that aren't part of a row, but one of my rows needs to be referred to as a column (I can never remember which end of the row that I see is the "top" of the column), and one of the other columns has a key in the wrong row (visually, the <kbd>tab</kbd> & <kbd>enter</kbd> keys are clearly in the same row as <kbd>G</kbd> & <kbd>H</kbd>, but the wiring puts them one row up from there). I would love to see the user-visible keymap addresses change to better reflect spatial positions and abstract away the wiring underneath.

gedankenexperimenter:
> For LED effects it would be beneficial to have a x-y LED coordinate system instead of matrix positions. The hardware class could offer a mapping from `KeyAddr` to x-y coordinates to allow for portable LED-effects. The Wavepool effect already implements such a mapping. The related methods of the hardware class would be removed by the linker whenever no plugin uses them.

Hear, hear! I started working on such a system for Kaleidoglyph about a month ago when I started in on LED modes. I think this is a very good idea.

src/kaleidoscope/MatrixAddr.h
noseglasses: I followed the convention of the STL, but you are right in Kaleidoscope it would be uppercase.

noseglasses:
> for (KeyAddr k : KeyAddr::Iterator{}) {
>  ...
> }

That would make the compiler call the `.begin()` and `.end()` methods of `KeyAddr::Iterator` which do not exist. in general a range based for loop is meant to iterate over a container or array rather than an iterator.

Would you find something like the following more readable?

```cpp
template<typename Addr__>
struct Range
{
   static Addr__ begin() { return Addr__{uint8_t(0)}; }
   static Addr__ end() { return Addr__{Addr__::upper_limit}; }
};
...
for(auto key_addr: Range<KeyAddr>{}) { ... }
```

noseglasses:
@obra, PR #618 removes the restriction of being able to support only one overload per hook/handler. Now cryptical versioned handler names like `onKeyswitchEvent2` are no more required.

gedankenexperimenter:
Re: `KeyAddr::Iterator{}` being used in the range-based for loop – you defined it (albeit all lowercase) as a typedef of `KeyAddr`, so it should be the same as `KeyAddr{}`. I did check to confirm that it works before suggesting it.

noseglasses:
> Re: KeyAddr::Iterator{} being used in the range-based for loop – you defined it (albeit all lowercase) as a typedef of KeyAddr, so it should be the same as KeyAddr{}. I did check to confirm that it works before suggesting it.

Ah, you are right. Due to my typedef it it's supposed to work but it's still highly confusing for anybody like me who is used to pass containers to range based for loops. Passing iterators is more than unconventional. And it only works due to the fact that my iterator class is identical to the container class which was already unorthodox. That's why I suggested the `Range<>` pseudo-container.

gedankenexperimenter:
`KeyAddr::Iterator{}` would give me a better idea of where to look to find the relevant code than `Range<>`, and it provides a much bigger hint about what's going on than `KeyAddr{}`, which certainly doesn't look like a container you can iterate through.

If you don't like using the term "iterator" there, that makes little difference to my argument — the reason I think it would be a good idea is that it tells me to go look for the definition of `KeyAddr`, and find the inner type `Iterator` (or whatever you want to call it), where there would be a comment explaining how it works and how to use it. It still wouldn't be as easy as I'd like to find it, because `KeyAddr` is a typedef of `MatrixAddr<>`, but that's just one more step. The hard part is still finding the definition of `KeyAddr`, which might not be in an obvious place.

gedankenexperimenter:
I'd also prefer not using `auto` as the type, even though it works just as well, because it's less clear to the unfamiliar reader. Likewise, I'm not a fan of the `Range<>` template, because the reader needs to translate through yet another template in order to get to understanding.

obra:
auto is a pretty reasonable modern C++ feature.

gedankenexperimenter:
> auto is a pretty reasonable modern C++ feature.

Certainly. But it does disguise the type of the variable, requiring more of the programmer. In this specific case, I think some people would find it a stumbling block that the type is not explicitly named, and I don't see it providing any advantage other than fewer characters to type.

obra:
> @obra, PR #618 removes the restriction of being able to support only one overload per hook/handler. Now cryptical versioned handler names like `onKeyswitchEvent2` are no more required.

Excellent. Thanks :) That's now merged.

obra:
@noseglasses 
> > KS_ROW_COL_FUNC
> > Stylistically, I think I prefer DEPRECATED(ROW_COL_FUNC) to KS_ROW_COL_FUNC because it makes it clearer what's going on when reading the code.
> 
> Yeah, that is more readable and serves the same purpose of being found by grep.
> 
Ok :)

> The reason why I generally tend to use prefixes for macros is their lack of namespaces. In larger software projects, they are a frequent source of trouble due to naming collisions.

This makes sense.  There's a middle-ground somewhere :)
> 
> > UNKNOWN_KEYSWITCH_LOCATION
> > The change from UNKNOWN_KEYSWITCH_LOCATION to UnknownKeyswitchLocation should be a standalone commit.
> 
> I didn't really want to change it but during the transition period we need two
> individual identifiers that represent two different C++ types. The reason why I chose the non all-caps `UnknownKeyswitchLocation` was
> that I wanted it to be a `constexpr`. Unfortunately, the compiler for unknown reasons instantiated `UnknownKeyswitchLocation`, thus wasting two bytes of RAM.
> 

I don't mind the change. I just want it encapsulated in one commit when we eventually merge

> > TapDance.md
> > In TapDance.md, it looks like there's an editing error, where you replaced var names with a type:
> > -### tapDanceAction(tap_dance_index, row, col, tap_count, tap_dance_action)
> > +### tapDanceAction(tap_dance_index, keyAddr, tap_count, tap_dance_action)
> 
> Almost. `keyAddr`, starting with a lower case letter was meant to be a variable. It's actually a leftover after a global renaming to provide consistency with the coding style guide.

No worries. In general, code should first match the code around it. That takes precedence over the style guide when they conflict.

> 
> > KeyAddr{}
> > In Kaleidoscope.cpp,
> > ```
> > for (auto key_addr : KeyAddr{}) {
> > Layer.updateLiveCompositeKeymap(key_addr);
> > ```
> > This is me being inexperienced, but I don't recognize what KeyAddr{} is supposed to do. Would you mind adding a line of comment?
> 
> `KeyAddr{}` is identical to `KeyAddr()`, both are calls to the default constructor.
> C++ introduced the curly brace syntax to disambiguate construction and function calls, which can help to speed up the compile.

If we're going to use {} for constructors, I'd rather we do it consistently throughout the codebase, rather than piecemeal. (And then enforce it with a c++ style checker.)


> Here, this creates an unnamed temporary that is used as a dummy for the range based for loop to query the iteration boundaries. The temporary is later entirely eliminated by the compiler.


> `for (auto key_addr : KeyAddr{})` is a quite frequent idiom in Kaleidoscope, now. Do you want all occurrences to be commented?

No, but I wonder if we could make it more readable by adding an allKeyAddresses somewhere that returns KeyAddr{} 

In a properly dynamic language, I'd just create a standard method that takes a callback or a closure and iterates against the keys -for- that closure.
> 
> > num_keys naming
> But I am a even greater fan of encapsulation. Adding such information to the hardware class instead of to global scope would make perfect sense. Let's have a `numKeys()` method.
> 
Ok

> > num_keys usage
> > src/kaleidoscope/plugin/Heatmap.cpp
> > It's still using ROWS * COLS
> > Doing some greps for ROWS and COLS, it does look like there are other places we're not using your new num_keys.
> > (Ideally, all the changes to num_keys would be done as one standalone commit.)
> 
> Alright, I will replace all occurrences in a separate commit.

Thanks

> 
> > Commented code
> > There are a couple places where you comment out code rather than deleting it or add new code with commented code inline. Generally, that's to be avoided
> 
> Ok.

Thanks
> 
> > Breaking apart commits
> > Generally for merge, we like to have each logical change or set of the same change broken into its own commit. That makes it much easier to read the history or figure out which change broke something. While working on a PR, it's ok to push updates, but before merge, we should tease things back apart.
> 
> Agreed. Once the submitted changes as a whole are agreed upon, I will separate them into logical changes wherever possible.
> 
Perfect

> > ActiveModColor
> > With code like this
> > ```
> >    if (::OneShot.isSticky(k))
> > ```
> > 
> > * ```
> >      ::LEDControl.setCrgbAt(r, c, sticky_color);
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), sticky_color);
> >    else if (::OneShot.isActive(k))
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.setCrgbAt(r, c, highlight_color);
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), highlight_color);
> >    else
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.refreshAt(r, c);
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.refreshAt(KeyLEDAddr(key_addr));
> >   ```
> >   } else if (k.raw >= Key_LeftControl.raw && k.raw <= Key_RightGui.raw) {
> >   if (hid::isModifierKeyActive(k))
> > 
> > 
> > * ```
> >      ::LEDControl.setCrgbAt(r, c, highlight_color);
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.setCrgbAt(KeyLEDAddr(key_addr), highlight_color);
> >    else
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.refreshAt(r, c);
> >   ```
> > 
> > 
> > * ```
> >      ::LEDControl.refreshAt(KeyLEDAddr(key_addr));
> >   ```
> >   } else if (k.flags == (SYNTHETIC
> > 
> > Is it more space efficient to instantiate KeyLEDAddr inside all those branches or would it make sense to create it once outside the branches?
> 
> We probably could move it outside of the conditional clauses without degrading performance.
> 
Ok

> > AlphaSquare
> > static void display(uint16_t symbol) {
> > 
> > * display(symbol, 0, 2);
> > 
> > 
> > * display(symbol, KeyLEDAddr{0, 2});
> >   }
> >   static void display(uint16_t symbol, uint8_t col) {
> > 
> > 
> > * display(symbol, 0, col);
> > 
> > 
> > * display(symbol, KeyLEDAddr(0, col));
> >   }
> > 
> > More instances of the KeyLEDAddr{} vs KeyLEDAddr() pattern. Is there a reason we want the {} pattern there?
> 
> Not necessarily. I use the `{}` syntax in any new code I write as it was introduced for good reasons. But I have not 100% adapted so after 11 p.m. I usually start writing `()`.
> But I agree, that it might lead to confusion when it is mixed the way it is here. Will adapt that.
> 
As above, let's not use it until we use it everywhere.

> > kaleidoscope/addr.h
> 
> We could as well stick with keynum.

I'm fine with the better name, as long as it's its own commit.

> 
> > Commit grouping and ordering
> > The really big issue we need to overcome is the naming of the hook replacement.
> > It should be possible to clean up the commits so that the onKeyswitchEvent2 change is on its own and the last thing in the series, after commits for each feature or logical group of features.
> 
> I assume that you want every individual commit to preserve the integrity of the overall firmware (buildable and runnable). Thus those independent commits might not be as many as you expect as many changes are pretty interrelated. I will see what I can do.
> 

I'll see what you can do :) 

> > Abstraction barriers
> > One thing that this change doesn't fully address (and does not need to address) is really breaking apart the ordered list of keys on the logical keyboard from the physical matrix on the circuit board. For the Model 01, this doesn't really matter because the matrix is fully populated. But for some other keyboards, there are big gaps in the matrix.
> > Historically, we haven't really differentiated between the logical key list and the physical matrix. This is both due to our abstractions not being good enough.
> 
> We already had LED row/col addresses and LED indices. So for LEDs there was a differentiation that leaked from the hardware implementation. After this PR we will not need to differentiate this anymore and the mapping between those two addressing types might become obsolete.

It occurs to me that LEDKeyAddr may not be necessary, Since a LEDKeyAddr points to the LED for a given Key, we ought to just be able to use the KeyAddr for that key. The mapping is internal to the Hardware class.


> Even thought this topic is strongly related to the transition to linear addressing, I did not dare to remove LED indices right away. This is because of the Chase and RainbowWave LED effects whose implementation is somewhat hardware dependent. Those effects will look differently on different types of keyboards. i was not sure where to go here.

Yeah.

> 
> The physical matrix setup of keys is something that is mostly interesting for LED effects. But it also matters for e.g. keymaps. My XKeymaps plugin that I hope to release soon, e.g. supports mirroring of keymaps. While only applicable to symmetric keyboards, this requires information about physical key positions.
> 
Understood

> > This makes me wonder whether anything inside the hardware/ directory should even know about KeyAddrs.
> > Is KeyAddr the Keyswitch Address in in the hardware key matrix or is it the logical key address in our global internal implementation of the list of keys?
> > (Does this question make sense? I sort of worry that this is a distinction that only makes sense to me.)
> 
> This question makes a lot of sense. From the perspective of the user or plugin implementor the `KeyAddr` has at least two independent properties.
> 
> 1. it uniquely references a key.
I agree with this usage.

> 2. it provides rough information about the relative position of keys, although it does not grant it.

I don't know if this is how we want to guarantee this. At most, it provides rough information about the -logical- layout of the keys, not the physical layout. 

> In both cases the user/plugin implementor does not care for the actual wiring. So the mapping from key position coordinates to electronic circuit coordinates would preferably be abstracted away by the hardware class.
> 
> That's why I agree with @gedankenexperimenter that the hardware must know about the `KeyAddr`, to be able to do the mapping. The user, on the other hand, shouldn't need to bother with the internals of the keyboard.
> 
> The `KEYMAP` macro is one of the implementation details, where I am not completely decided how I would like to see things handled. We currently use row/col information to reference keys there, but we could as well use the integer value that is stored within the `KeyAddr` class. For some keyboards like the Planck, row/cols make perfect sense. But for most other modern keyboards, including the Model01, the assignment of row/col values to thumb and palm keys reveals internal details that are of no use to the user.
> 
> Another problem are the LED effects in general. I assume that most of them have been designed with the Model01 in mind with respect to key-LED positions. That's why some of them might be less portable than others.
> 
> For LED effects it would be beneficial to have a x-y LED coordinate system instead of matrix positions. The hardware class could offer a mapping from `KeyAddr` to x-y coordinates to allow for portable LED-effects. The Wavepool effect already implements such a mapping. The related methods of the hardware class would be removed by the linker whenever no plugin uses them.
> 
> As a consequence the question arises whether the `KeyAddr` type should expose `row()`/`col()` values at all. Currently we need such methods for the compatibility layer introduced by this PR. But once the deprecated stuff is removed, row/col are only used internally by the hardware class.

Yeah, I think that KeyAddr is not the ultimate right place for relative key information. We -should- eventually have a good way to get that, but it's a complicated thing to get right and I'd rather not conflate this and that..  (And storing X,Y,Z positions along with rotations can get....to be a bit verbose for the limited amount of space in a 32u4)

> One way we could handle a transition would be to declare the leftover compatibility functions/methods/classes that require direct access `row()`/`col()` to be friends of the `MatrixAddr` class. Thus we would prevent any new code from accessing row/col information directly. Then, after the deprecation periode elapsed, we would remove these friend declarations together with anything that reminds of row/col in `MatrixAddr`. The encoding/decoding of `KeyAddr` would then entirely be left to the hardware classes.

Up to now, we've depended on the deprecation messages to be enough for that. I think that should be ok here, too. I don't -think- that it's worth the overhead of the friend class dance.

> 
> I am well aware that this means an additional change to this PR. But from my perspective it makes sense to deal with the abstraction thing right now as we are at it, instead of fixing it later.
> 
> What do you think?

I agree that we can make things better now, but better is better than perfect. Let's get the next "right" step cleaned up and merged, then we can iterate from there.

noseglasses:
>> KeyAddr{} is identical to KeyAddr(), both are calls to the default constructor.
>> C++ introduced the curly brace syntax to disambiguate construction and function calls, which can >> help to speed up the compile.

>If we're going to use {} for constructors, I'd rather we do it consistently throughout the codebase, >rather than piecemeal. (And then enforce it with a c++ style checker.)

Replacing the constructor syntax ((...) -> {...}) is something that could only be done by a sophisticated tool as it requires syntax parsing to distinguish constructor calls form ordinary function calls or just expressions. I am currently not aware of any open source program/editor that can do that out of the box.

noseglasses:
> In a properly dynamic language, 

Though i love C++ for its mindblowing complexity, I would be happier to develop this firmware in Python :-)

> I'd just create a standard method that takes a callback or a closure > and iterates against the keys -for- that closure.

You don't need a dynamic language for that, checkout [std::for_each](https://en.cppreference.com/w/cpp/algorithm/for_each). Unfortunately, for the time being we have to live without the standard library on Arduino.

noseglasses:
> And storing X,Y,Z positions along with rotations can get....to be a bit verbose for the limited amount of space in a 32u4

Wow, yeah, we definitely need rotations. As soon as we support keyboards with highly curved surfaces, LED rotations might be inevitable. Are we going for quaternions or rotation matrices to do the job ;-)

Here's an example for [LEDs on curved surfaces](https://www.youtube.com/watch?v=qFlijYZxgkg) - they are even rotating (a project i was involved with some years ago).

noseglasses:
Closing this in favor of a cleaned up version: #621

PR #593, opened 2019-03-02
Led control efficiency changes

gedankenexperimenter:
I'm proposing two very small changes to how LEDControl operates:

First, changing the frequency at which updates are pushed to the hardware from ~60Hz to ~30Hz. 60Hz is overkill for this application; with such a grainy "display", I would be quite surprised if it was possible for humans to detect a difference. However, I have only one Model01, so I don't have the ability to do a side-by-side blind comparison to verify that this is the case.

Second, I'm proposing only calling the `update()` method of the active LED mode once per LED update cycle, rather than once per scan cycle. It doesn't make sense to me to be doing (depending on the LED mode) all that computation when it doesn't show up in the output.

I've tried both of these things out, and have not been able to detect a visual difference with any of the standard sketch's LED modes. For some modes, however, there is a marked decrease in the average scan cycle time.

### DIFF
--- a/src/kaleidoscope/plugin/LEDControl.cpp
+++ b/src/kaleidoscope/plugin/LEDControl.cpp
@@ -22,7 +22,7 @@ namespace plugin {
 
 LEDMode *LEDControl::modes[LED_MAX_MODES];
 uint8_t LEDControl::mode;
-uint16_t LEDControl::syncDelay = 16;
+uint16_t LEDControl::syncDelay = 32; // 32ms interval => 30Hz refresh rate
 uint16_t LEDControl::syncTimer;
 bool LEDControl::paused = false;
 
--- a/src/kaleidoscope/plugin/LEDControl.cpp
+++ b/src/kaleidoscope/plugin/LEDControl.cpp
@@ -176,8 +176,8 @@ kaleidoscope::EventHandlerResult LEDControl::beforeReportingState(void) {
   if (elapsed > syncDelay) {
     syncLeds();
     syncTimer += syncDelay;
+    update();
   }
-  update();
 
   return kaleidoscope::EventHandlerResult::OK;
 }


gedankenexperimenter:
I considered marking this as a draft PR because it begs for testing that I do not have the hardware for, and because each change could be accepted independently. Those changes are complete, however, and I'm not concerned about it being merged without sufficient testing.

obra:
I think I'd prefer that we expose the FPS control to the user so that
it's possible for them to make the choice about the perf tradeoff.
Talking to other keyboard implementers, they're finding value in
pushing framerates up to 100FPS or higher.

gedankenexperimenter:
Making it user-configurable sounds great to me.

I'm _very_ skeptical of a claim that high frame rates like 100fps are perceptibly different from 30fps on keyboard LEDs, unless they're cycling on and off that fast, which is not the case with any Kaleidoscope plugin as far as I know, and doesn't seem like it would serve a useful purpose. Have you heard what value those other implementors have found in high frame rates?

On the Model01, 100fps is getting close to the theoretical limit; it would be spending the vast majority of its time pushing LED updates.

gedankenexperimenter:
Perhaps it would be best to have a reasonable default set in the hardware module, with a way for the user to override it in a sketch.

algernon:
>  think I'd prefer that we expose the FPS control to the user so that it's possible for them to make the choice about the perf tradeoff.

The rate is already user-controlled, because `syncDelay` is public, users are free to change the value. We *could* make it easier by providing some constexprs like `constexpr uint16_t LEDControl::FPS_30 = 32` and `constexpr uint16_t LEDControl::FPS_60 = 16`.

Mind you, 33 is closer to 30FPS, 32 is a bit over 31FPS. We used 16 for 60FPS because 17 would be 58, 16 is 62, and better higher than lower. 33 would be 30.3 FPS.

The constants are fine in a separate PR I think.

> Have you heard what value those other implementors have found in high frame rates?

IIRC InputClub was able to do some very, very smooth animations, but they're using ARM MCUs, and the animation in question was on a single-piece keyboard, so no I2C or similar required either, making LED updates considerably faster. 100FPS is certainly not a viable option for AVR-powered boards, those simply lack the power to push out updates fast enough.

See [this video](https://www.youtube.com/watch?v=UCVXgRHDaqY) for an example.

Since we *are* going to support ARM boards soon, being able to push way higher FPS is a good thing. Thankfully, with `syncDelay` being public, we can do that with reasonable ease.

> Perhaps it would be best to have a reasonable default set in the hardware module, with a way for the user to override it in a sketch.

This is a good idea. We're going to rework `LEDControl` a bit anyway for Underglow support, this is something we can do around the same time.

gedankenexperimenter:
Only slightly related, but at some point, presumably, you'll want to make LEDControl comply with the coding standards doc, whereupon `syncDelay` should become private (and become `sync_delay_`), and a setter will need to be added. When that happens, I would like to suggest also replacing the `delay` in its name with `interval`. That much better describes what it does.

gedankenexperimenter:
I see the use case for high frame rates on LED updates for now — very fast changes that are meant to look smooth. But none of the Model01's default modes do that, and they all look just as smooth to me at 30fps as they do at 60.

@algernon, as you plan your changes to LEDControl, I think it would be beneficial to let each LED mode request the update interval it wants, and allow the board hardware or the user sketch set a minimum (i.e. max fps). Maybe a maximum, too. Then, when switching to a mode that benefits from higher frame rates, the interval can be changed on the fly, but limited so as not to overwhelm the hardware. And when switching to a mode that doesn't request a particular rate, a sketch-defined (or hardware-defined) default can be used.

I think it would be nice for users to be able to switch on the fly between typing-performant and LED-performant modes, and it makes sense to tie that to the LED mode that has been chosen.

noseglasses:
I'm following this discussion for a while, please allow me to chime in.

> I think it would be nice for users to be able to switch on the fly between typing-performant and LED-
performant modes, and it makes sense to tie that to the LED mode that has been chosen.

I agree, let the `LEDMode` class decide on the synchronization interval. Two ways to enable this come to my mind, one with and one without breaking changes.

We could add a virtual `getDesiredUpdateInterval()` method to `LEDMode` whose default implementation would return 32 (non-breaking change but one more function in the virtual function table which is expensive as only few LEDModes would probably want to change the default) or modify the `update()` callback's signature to `virtual void update(uint16_t &sync_interval)`, which would be a breaking change but might cause less groth in terms of PROGMEM. Only those derived LEDModes that require it would let their `update(...)` modify the `sync_interval` parameter.
 
The latter could be called by `LEDContol::update()` as
```cpp
sync_interval_ = default_sync_interval_;
modes[mode]->update(sync_interval_);
//... restrict sync_interval_ ...
```

I'm not 100 % sure which version would increase PROGMEM more. The first one adds an additional virtual function pointer and some code for possible `getDesiredUpdateInterval()` implementations, the other would slightly increase the call overhead on the caller side. Only for those methods that would actually use the `sync_interval` parameter would cause code on the callee's side.

I'd go for the latter and live with the breaking changes but one could also have two versions of `LEDMode::update` for a transition period, deprecate the old one but call both. This would enable a smooth transition without immediate breaking changes.

gedankenexperimenter:
I would just add a function (or two) to LEDControl that LED modes could call from `onActivate()`, rather than `update()`.  Then LEDControl could reset to the default interval before calling a new mode's `onActivate()` method.

gedankenexperimenter:
For something like the RainbowWave effect, the update interval could depend on how fast the changes are going to happen — if the user configures it for a very high speed, it would request a faster update interval than a very slow speed wave.

noseglasses:
> I would just add a function (or two) to LEDControl that LED modes could call from onActivate(), rather than update(). Then LEDControl could reset to the default interval before calling a new mode's onActivate() method.

Nice approach. That would avoid any breaking changes.

gedankenexperimenter:
> Nice approach. That would avoid any breaking changes.

Sometimes, having fewer tools to work with is a good thing. =)

algernon:
Allowing plugins to select their own refresh rate is a good idea. Lets open a new issue about it, so we can discuss the API and how it should behave under various scenarios.

PR #595, opened 2019-03-08
A more widely usable version of the Model01 test mode

obra:
This PR generalizes and improves the Model01 test mode such that it should work for other hardware besides the Model 01.

To do this, we add several new APIs to the low level hardware implementations. We add these APIs to all devices currently in the Kaleidoscope tree.

The new plugin has a different activation mechanism than the old one -- It's not strictly compatible. Sketches will need to be modified to use the new plugin.

A PR for the Model01-Firmware sketch will follow this PR, but should not be applied until this is in tree.

As currently implemented, using this test mode takes the Model01 firmware from:

  - Program:   26076 bytes (90.9% Full)
  - Data:       1437 bytes (56.1% Full)

to

  - Program:   25928 bytes (90.4% Full)
  - Data:       1441 bytes (56.3% Full)

for a savings of 118 bytes of program at a current cost of 4 bytes of data

### DIFF
--- a/src/Kaleidoscope-HardwareTestMode.h
+++ b/src/Kaleidoscope-HardwareTestMode.h
@@ -1,0 +1,19 @@
+/* Kaleidoscope-Hardware-TestMode - A factory test mode for the Model 01.
+ * Copyright (C) 2017-2019  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "kaleidoscope/plugin/HardwareTestMode.h"
--- a/src/kaleidoscope/hardware/ATMegaKeyboard.cpp
+++ b/src/kaleidoscope/hardware/ATMegaKeyboard.cpp
@@ -85,6 +85,20 @@ bool ATMegaKeyboard::isKeyswitchPressed(uint8_t keyIndex) {
                             keyIndex % KeyboardHardware.matrix_columns);
 }
 
+
+
+bool ATMegaKeyboard::wasKeyswitchPressed(uint8_t row, byte col) {
+  return (bitRead(KeyboardHardware.previousKeyState_[row], col) != 0);
+
+}
+
+bool ATMegaKeyboard::wasKeyswitchPressed(uint8_t keyIndex) {
+  keyIndex--;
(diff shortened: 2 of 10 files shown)


src/Kaleidoscope-HardwareTestMode.h
```diff
@@ -0,0 +1,19 @@
+/* Kaleidoscope-Model01-TestMode - A factory test mode for the Model 01.
+ * Copyright (C) 2017-2018  Keyboard.io, Inc.
```
algernon: I guess this part of the header could use an update? A few other files would need the name & copyright dates updated too.

src/kaleidoscope/hardware/ATMegaKeyboard.cpp
```diff
@@ -85,6 +85,27 @@ bool ATMegaKeyboard::isKeyswitchPressed(uint8_t keyIndex) {
                             keyIndex % KeyboardHardware.matrix_columns);
 }
 
+
+uint8_t ATMegaKeyboard::previousPressedKeyswitchCount() {
+  uint8_t count = 0;
+
+  for (int8_t r = 0; r < KeyboardHardware.matrix_rows; r++) {
+    count += __builtin_popcount(KeyboardHardware.previouskeyState_[r]);
```
algernon:
```suggestion
    count += __builtin_popcount(KeyboardHardware.previousKeyState_[r]);
```

algernon:
One more thing I was thinking about: would it be possible to make a compat plugin, so that any existing user sketches that use the old Model01 hardware test plugin, would still work?

Also, this should be added to NEWS and UPGRADING (I can do that part).

src/kaleidoscope/hardware/ATMegaKeyboard.cpp
obra: force-pushed a fix

src/Kaleidoscope-HardwareTestMode.h
obra: force-pushed a cleanup of this.

obra:
I'd be happy to have you do upgrading and news.

The old plugin with the old name continues to work with old sketches. I don't think a compat shim makes much more sense than just keeping the old one around for a while.

algernon:
Oh, right, this doesn't remove the old one. Awesome. Perhaps we could add a deprecation notice to it then?

I'll let Travis pass, write the NEWS and UPGRADE parts and merge it, we can deprecate the old plugin later.

Thanks!

src/kaleidoscope/hardware/ATMegaKeyboard.cpp
```diff
@@ -85,6 +85,27 @@ bool ATMegaKeyboard::isKeyswitchPressed(uint8_t keyIndex) {
...
 
+
+uint8_t ATMegaKeyboard::previousPressedKeyswitchCount() {
+  uint8_t count = 0;
+
+  for (int8_t r = 0; r < KeyboardHardware.matrix_rows; r++) {
+    count += __builtin_popcount(KeyboardHardware.previousKeyState_[r]);
+  }
+  return count;
+
```
algernon:
```suggestion
}

```

obra: pushed a fix

obra:
@algernon@keyboard.io <algernon@keyboard.io>  I think this is ready for you
now
ᐧ

PR #600, opened 2019-03-12
KEYMAP_*+_GENERIC macros and keymasks

noseglasses:
I am working on a plugin that provides extended keymap features:
https://github.com/CapeLeidokos/Kaleidoscope-XKeymaps

It's almost finished but to make it work, I need all keyboard hardware headers to define a macro that works similar to the existing `KEYMAP_STACKED` macro. The only difference is that the new macro generates a linear list of entities instead of a braced list. The new macro has an additional first macro parameter `dflt` (=default, name shortened to match row/col macro arguments as r0c0) that is used where the `KEYMAP_STACKED` macro would specify `XXX` for unused key matrix positions.

I rely on these macros with `Kaleidoscope-XKeymaps` to conveniently initialize keymap bitfields (keymasks) that are stored in PROGMEM (see the usage of `KEYMASK_STACKED_PROGMEM` in the example provided in the README.md of `Kaleidoscope-XKeymaps`). As a sidenote, the whole bitfield for the Model01 is really stored in 8 PROGMEM bytes and automatically initialized at compile time.

The newly introduced macros will possibly be useful in other places whenever information is supposed to be assigned to keys and this information is meant to be specified in a `KEYMAP_STACKED`-like fashion but passed as a linear list of entities, e.g. as arguments to a function or as an initialization list.

### DIFF
--- a/examples/LEDs/LED-Wavepool/LED-Wavepool.ino
+++ b/examples/LEDs/LED-Wavepool/LED-Wavepool.ino
@@ -20,25 +20,26 @@
 #include <Kaleidoscope-LEDControl.h>
 #include <Kaleidoscope-LED-Wavepool.h>
 
-const Key keymaps[][ROWS][COLS] PROGMEM = {
+KEYMAPS(
   [0] = KEYMAP_STACKED
-  (
-    Key_LEDEffectNext, Key_1, Key_2, Key_3, Key_4, Key_5, Key_LEDEffectNext,
-    Key_Backtick,      Key_Q, Key_W, Key_E, Key_R, Key_T, Key_Tab,
-    Key_PageUp,        Key_A, Key_S, Key_D, Key_F, Key_G,
-    Key_PageDown,      Key_Z, Key_X, Key_C, Key_V, Key_B, Key_Escape,
+        (
+          Key_LEDEffectNext, Key_1, Key_2, Key_3, Key_4, Key_5, Key_LEDEffectNext,
+          Key_Backtick,      Key_Q, Key_W, Key_E, Key_R, Key_T, Key_Tab,
+          Key_PageUp,        Key_A, Key_S, Key_D, Key_F, Key_G,
+          Key_PageDown,      Key_Z, Key_X, Key_C, Key_V, Key_B, Key_Escape,
 
-    Key_LeftControl, Key_Backspace, Key_LeftGui, Key_LeftShift,
-    Key_NoKey,
+          Key_LeftControl, Key_Backspace, Key_LeftGui, Key_LeftShift,
+          Key_NoKey,
 
-    Key_skip,  Key_6, Key_7, Key_8,     Key_9,      Key_0,         Key_skip,
-    Key_Enter, Key_Y, Key_U, Key_I,     Key_O,      Key_P,         Key_Equals,
-    Key_H, Key_J, Key_K,     Key_L,      Key_Semicolon, Key_Quote,
-    Key_skip,  Key_N, Key_M, Key_Comma, Key_Period, Key_Slash,     Key_Minus,
+          Key_skip,  Key_6, Key_7, Key_8,     Key_9,      Key_0,         Key_skip,
+          Key_Enter, Key_Y, Key_U, Key_I,     Key_O,      Key_P,         Key_Equals,
+          Key_H, Key_J, Key_K,     Key_L,      Key_Semicolon, Key_Quote,
+          Key_skip,  Key_N, Key_M, Key_Comma, Key_Period, Key_Slash,     Key_Minus,
 
-    Key_RightShift, Key_RightAlt, Key_Spacebar, Key_RightControl,
-    Key_NoKey),
-};
+          Key_RightShift, Key_RightAlt, Key_Spacebar, Key_RightControl,
+          Key_NoKey
+        )
+) // KEYMAPS(
 
 KALEIDOSCOPE_INIT_PLUGINS(
   LEDControl,
--- a/src/kaleidoscope/Kaleidoscope.h
+++ b/src/kaleidoscope/Kaleidoscope.h
@@ -40,6 +40,14 @@ void setup();
 
 extern HARDWARE_IMPLEMENTATION KeyboardHardware;
 
+#ifdef KEYMAP_STACKED_GENERIC
+#define KEYMAP_STACKED(...) { KEYMAP_STACKED_GENERIC(XXX, __VA_ARGS__) }
+#endif
+
+#ifdef KEYMAP_GENERIC
+#define KEYMAP(...) { KEYMAP_GENERIC(XXX, __VA_ARGS__) }
+#endif
+
 #define ROWS (KeyboardHardware.matrix_rows)
 #define COLS (KeyboardHardware.matrix_columns)
 #define LED_COUNT (KeyboardHardware.led_count)
--- a/src/kaleidoscope/bitfields.cpp
+++ b/src/kaleidoscope/bitfields.cpp
@@ -1,0 +1,53 @@
+/* Kaleidoscope - Firmware for computer input devices
+ * Copyright (C) 2013-2018  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "Kaleidoscope.h"
+
+#include "bitfields.h"
+
+namespace kaleidoscope {
+namespace bitfields {
+namespace internal {
+
+bool _BaseBitfield::isBitSetP(const void *bit_field, uint8_t raw_pos) {
+  uint8_t byte_id = raw_pos >> 3;
+  uint8_t bit_pos = raw_pos & 0x7;
+  const uint8_t *bytes = reinterpret_cast<const uint8_t *>(bit_field);
+  return bytes[byte_id] & (0x1 << bit_pos);
+}
+
+void _BaseBitfield::setBitP(void *bit_field, uint8_t raw_pos, uint8_t val) {
+  uint8_t byte_id = raw_pos >> 3;
+  uint8_t bit_pos = raw_pos & 0x7;
+  uint8_t *bytes = reinterpret_cast<uint8_t *>(bit_field);
+  if (val) {
+    bytes[byte_id] |= (0x1 << bit_pos);
+  } else {
+    bytes[byte_id] &= ~(0x1 << bit_pos);
+  }
+}
+
+bool _BaseBitfield::isBitSetPROGMEM_P(const void *bit_field, uint8_t raw_pos) {
+  uint8_t byte_id = raw_pos >> 3;
+  uint8_t bit_pos = raw_pos & 0x7;
+  const uint8_t *bytes = reinterpret_cast<const uint8_t *>(bit_field);
+  uint8_t byte = pgm_read_byte(&(bytes[byte_id]));
+  return byte & (0x1 << bit_pos);
+}
+
+} // end namespace internal
+} // end namespace bitfields
+} // end namespace kaleidoscope
--- a/src/kaleidoscope/bitfields.h
+++ b/src/kaleidoscope/bitfields.h
@@ -1,0 +1,200 @@
+/* Kaleidoscope - Firmware for computer input devices
+ * Copyright (C) 2013-2018  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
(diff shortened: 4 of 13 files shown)


algernon:
This makes sense, though I wonder if it wouldn't be better to replace the current `KEYMAP_STACKED` / `KEYMAP` macros with something similar, instead of adding a new representation?

(Also, can you add a `Signed-Off-By:` line to your commits?)

noseglasses:
The `KEYMAP_STACKED` macros create lists that are brace initialized. They are used to initialize 2D arrays. 

The newly introduced `KEYMAP_STACKED_LIST` macros generate linear lists without braces. Those are e.g. useful to pass information that is defined in the sketch in a stacked-keymap fashion as parameters to functions or other macros. There the braced `KEYMAP_STACKED` cannot be used.

The `KEYMAP_STACKED` macros could be replaced by the newly introduced macros only if the two dimensional keymap array would be replaced by a 1D array. That's something that has been suggested in other places before, e.g. [here](https://github.com/keyboardio/Kaleidoscope/pull/589#pullrequestreview-208709849). I would also favor such a design for multiple reasons, e.g. key address handling.

But as that is a different matter I introduced the `KEYMAP_STACKED_LIST` for the time being. If there would one day be a decision to switch over to 1D arrays, `KEYMAP_STACKED` could become a simple wrapper in `Kaleidoscope.h` that wraps specialized versions of `KEYMAP_STACKED_LIST` that are defined by the hardware headers.

P.S.: I tried the signoff stuff and force pushed the signed version but with no luck (see the listed force push above). I will that again, though.

noseglasses:
Took me a while to find out what was wrong. All commits must of course be signed off. One was, the other wasn't. Github does not show which commit was the culprit. Squashed and force pushed.

obra:
Sorry, my comment from earlier in the week didn't go through.

I really, really don't like requiring device implementors to create additional versions of these macros.  Is there anything you can do to not require every hardware implementation to have duplicated "list" versions of the keymap macros?

noseglasses:
> I really, really don't like requiring device implementors to create additional versions of these macros.

Very much understandable.

I know a lot of macro black magic but magical power is not strong enough to which away those curly braces created by the current `KEYMAP_STACKED` macro. Macro tricks can remove ordinary brackets but curly braces are untouchable.

If it was me to decide, I would throw away the current implementation of `KEYMAP_STACKED` immediately because of its lack of flexibility compared to the proposed linear approach. This would require moving from two dimensional `keymap` arrays to one dimensional ones. This is no big deal as those are only used internally by the `Layer` class and everything is wrapped by macros already, thus the transition would be pretty painless. 

As it cannot be 100% excluded that someone relied on the current 2D implementation in a plugin, i propose a [wrapper approach](https://godbolt.org/z/fcS35m) with deprecation for a transition time. As it turns out, the wrapper access, the direct access to the linear array and the access to 2D arrays requires pretty much the same number of instructions.

noseglasses:
To make it easier to compare the performance of access to a linear array compared to a two-dimensional array, I [slightly modified](https://godbolt.org/z/9KmOIP) the previous example and wrapped the array access with individual functions.

I simply counted the number of instructions of the individual functions in the example (`oldStyleAccess`, `wrapperAccess`, `indexedLinearAccess`, `indexedLinearAccessThroughInlinedAccessFunction`). The number of instructions required for the different modes of access are as follows. 

Old style indexed row/col access: __26__
Wrapper access with same signature (for transition period): __29__
Direct access to linear array: __27__
Access to linear array through inlined convenience function: __27__ (preferred solution)

This means that eventually we will have only one instruction per access more by going from 2D to 1D keymap arrays but gain much more widely usable `KEYMAP...` macros.

FWIW, the access through a convenience function would e.g. read `keyFromPROGMEMKeymap(layer, row, col)`.

I'd be happy to submit a PR for the overall change (2D->1D + modified `KEYMAP_STACKED` and `KEYMAP` macros) that would be entirely backward compatible. This would completely avoid the `KEYMAP_STACKED_LIST` macros proposed by this PR.

noseglasses:
@obra, could you please comment on my ideas about a transition to 1D keymap arrays. The reason why i bother you is that the transition of the `KEYMAP_STACKED` macros is the only missing peace that prevents me right now from publishing my [Kaleidoscope-XKeymaps](https://github.com/CapeLeidokos/Kaleidoscope-XKeymaps) plugin.

Also the keymask feature proposed [here](https://github.com/keyboardio/Kaleidoscope/issues/604) would rely on 1D `KEYMAP...` macros.

I would be happy to submit a PR that demonstrates the necessary changes in the Kaleidoscope core.

obra:
@algernon and I have discussed it and agree that moving to a 1D array of keys is probably the right thing for kaleidoscope globally going forward.

noseglasses:
Alright, thanks, I will submit a PR.

noseglasses:
I just force-pushed an update that does away with the additional `LINEAR` versions of keymaps and instead modifies the overall keymap handling to go with linear keymap arrays. As suggested above in this thread, a compatibility wrapper class was introduced that mimics 2D array access.

Thus, this PR is entirely backward compatible.

We even save some bytes of PROGMEM: [kaleidoscope_change_report__linear_keymap_arrays.pdf](https://github.com/keyboardio/Kaleidoscope/files/3008743/kaleidoscope_change_report__linear_keymap_arrays.pdf)

noseglasses:
I just added another commit that adds the keymasks that I had announced here: https://github.com/keyboardio/Kaleidoscope/issues/604

Keymasks will be a key feature of my new xkeymaps plugin but I am pretty sure that they will beneficial in other places of the firmware, too.

Technically the second commit is orthogonal to the first one of this PR. But the second one does depend on the modifications to the `KEYMAP` and `KEYMAP_STACKED` macros that are induced by the first commit. Thus, I find it convenient to have them both together. 

I added a detailed description of the keymask features to the commit message of the second commit.

I changed the title of this PR accordingly.

@algernon and @obra, is that ok for you or do you want me to add a separate PR?

noseglasses:
There's one thing that i forgot to ask.

I added the files `keymask.h` and `keymask.cpp` to `Kaleidoscope/src/kaleidoscope`. As I know little about Arduinos internal policy about what to compile and what not, i wonder if `keymask.cpp` will always be compiled? On the one hand, that's good, because there are no linker errors of missing symbols to be expected and the linker does eliminate unused symbols anyway. On the other hand we should restrict the amount of compiled code to only such parts that are actually going to be linked into the firmware. Firmware compiles should be as fast as possible to make users and developers (smoke-examples) happy.

i am a bit lost here.

gedankenexperimenter:
From experience, I can tell you that arduino-builder compiles all `*.cpp` files it finds in the source tree of any libraries that get included via the header files referred to in the sketch. Just try putting an unrelated `.cpp` file with a syntax error somewhere, and you'll see.

noseglasses:
> From experience, I can tell you that arduino-builder compiles all *.cpp files it finds in the source tree of any libraries that get included via the header files referred to in the sketch. Just try putting an unrelated .cpp file with a syntax error somewhere, and you'll see.

Ok, thanks for the clarification. That's what i suspected. In the case of the `Bitfield` stuff it's fortunately just four tiny functions.

src/kaleidoscope/keymask.cpp
```diff
@@ -0,0 +1,53 @@
+/* Kaleidoscope-XKeymaps
```
obra: Looks like this was extracted from elsewhere..

src/kaleidoscope/keymask.h
```diff
@@ -0,0 +1,202 @@
+/* Kaleidoscope-XKeymaps
```
obra: extracted from elsewhere?

obra:
I don't -think- I like the term 'keymask'  - since it's a keymap bitfield, maybe we should just call it a keymap bitfield?

You mention "As it cannot be 100% excluded that someone relied on the current 2D implementation in a plugin, i propose a wrapper approach with deprecation for a transition time. As it turns out, the wrapper access, the direct access to the linear array and the access to 2D arrays requires pretty much the same number of instructions." 

I'm ok with wrappers and deprecation, but I'm having trouble thinking of where this -could- be used from a Plugin.

noseglasses:
> but I'm having trouble thinking of where this -could- be used from a Plugin.

Any plugin could theoretically contain code that does `pgm_read_word(&keymap[layer][row][col])`. Although I agree that it is not very likely.

noseglasses:
I updated the second commit
* files `keymask.*` are now named `bitfields.*`
* the file headers have been fixed for Kaleidoscope
* the macros `KEYMASK` and `KEYMASK_STACKED` have been renamed `KEYMAP_BITFIELD` and `KEYMAP_BITFIELD_STACKED`

noseglasses:
Just noticed that the ergodox hardware internally uses a keymap bitifield. This would be a perfect use case for the newly introduced `Bitfield` class.

noseglasses:
`ATMegaKeyboard` also uses a self made bitfield that could be wrapped/replaced.

noseglasses:
@algernon, @obra, could you please have a look at this PR. I really would like to release my XKeymaps-plugin which unfortunately depends on this.

obra:
I've asked @algernon to dig into this ASAP
ᐧ

obra:
I'm really sorry we're late on this. Algernon and I have both been slammed

> ᐧ

obra:
I started splitting this up into smaller commits, breaking out the bitfield classes into a commit and the wavepool example cleanup into a commit.

I've pushed both of those to master.

This makes it a little bit easier to deal with the rest of it.

I'm once again trying to figure out if there are any actual usages of the "old-style" keymap lookups outside the core. Ideally, we'd be able to avoid having the "_GENERIC" variants.

noseglasses:
The  `KEYMAP_*_GENERIC` macros are useful far beyond the definition of actual keymaps. They can serve to elegantly define general per-key information. For such purposes, it is useful to be able to assign special values to those keys that are not used (holes). When defining a bitfield, one might want to choose if the holes are assigned a one or a zero value.

I'm not sure about the names `KEYMAP_*_GENERIC`, though. The term `KEYMAP` might be missleading. But to let each hardware define such a macro instead of the original `KEYMAP_*` macros provides much greater flexibility.

For my [Kaleidoscope-LEDEffect-Maelstrom](
https://github.com/CapeLeidokos/Kaleidoscope-LEDEffect-Maelstrom) plugin, I use the `KEYMAP_STACKED_GENERIC` macro to specify per-key cartesian coordinates. The outcome looks as follows:

whirl

See [here](https://github.com/CapeLeidokos/Kaleidoscope/blob/ng_coordinate_infrastructure/src/kaleidoscope/hardware/keyboardio/Model01.h) for definition of the cartesian key-coordinates (at the bottom of the file).

FWIW, there's no floating point math involved at runtime. Everything happens at compiletime and it costs only 128 bytes of PROGMEM to store some integer representations of per-key polar coordinates.

The [Kaleidoscope-XKeymaps](https://github.com/CapeLeidokos/Kaleidoscope-XKeymaps) plugin will also rely on this PR. it uses key-bitfields that are defined via the `*_GENERIC` macros.

obra:
Right. I was trying to figure out if we can just drop the _GENERIC part of the name. There’s gotta be a good descriptive name. FWIW, that’s my only current concern about the PR. Naming is hard.

noseglasses:
Some naming ideas: `KEYWISE_DATA`, `PER_KEY_DATA`, `KEY_PAYLOAD` or, as those macros generate a list, `KEYWISE_DATA_LIST`, etc.

obra:
PER_KEY_DATA is what I was going to propose. Given two independent suggestions of that name, let’s run with it.

obra:
Ok. I've made the naming change and rebased on Master. 

https://github.com/keyboardio/Kaleidoscope/commit/547db75474fd48d42df5ae17ec39f1e942086d68 is the commit containing everything as-yet unmerged.

@algernon - I'd love a thumbs up / down on the current state. @noseglasses: have I butchered your work?

noseglasses:
@obra, changes look good. I definitely like the new macro name.

algernon:
:+1: 

We should add this to `UPGRADING.md` and `NEWS.md`, and decide on a removal date. I think we used about 3 months in the past. How does a removal date of August 13 sound?

obra:
great

obra:
It took a while, but this should now be fully merged, including NEWS and
UPGRADING entries. @noseglasses, thanks so much for your patience. Can you
confirm that everything looks ok?
ᐧ

noseglasses:
@obra, if you are talking about https://github.com/keyboardio/Kaleidoscope/commit/547db75474fd48d42df5ae17ec39f1e942086d68, that looks ok to me. Did you run smoke-examples on that version? 

But i did not quite understand you statement about NEWS and UPGRADING. https://github.com/keyboardio/Kaleidoscope/commit/547db75474fd48d42df5ae17ec39f1e942086d68 doesn't seem to touch these.

obra:
The merged head is
https://github.com/keyboardio/Kaleidoscope/commit/b818be1d8ebb376bae86c49082d6405f943dac56

The smoke test from travis with the examples run is
https://travis-ci.org/keyboardio/Kaleidoscope/builds/532989203
ᐧ

noseglasses:
I see. I got a bit confused due to your use of separate branches. Anyway, thanks a lot for review and merge!

Do you want me to close this PR?

Can the https://github.com/keyboardio/Kaleidoscope/tree/ng_keymap_stacked_list branch be deleted?

noseglasses:
BTW, both NEWS and UPGRADING in head look ok.

obra:
I've closed it :)

noseglasses:
@obra, could you please pull up the bundle repo.

obra:
Done. (Though I -believe- you have permissions to do that, too, now :)
ᐧ

noseglasses:
> Done. (Though I -believe- you have permissions to do that, too, now :)

Permissions yes, but so far no official approval ;-)

Are there any restrictions or is it always ok to pull up any of the bundle repos to the current HEAD?

algernon:
It is usually ok to pull up any of the repos to current HEAD, so go ahead.

obra:
HEAD of the various repositories -should- always be releasable. In general,
if the subrepository travis builds are coming up green, it's always ok to
"make blindly-commit-updates" in Kaleidoscope-Bundle-Keyboardio.

Best,
Jesse
ᐧ

PR #601, opened 2019-03-12
Introduced nextLowerLayer(layer) to support non stacked keymap implem…

noseglasses:
I am working on a plugin that provides extended keymap features:
https://github.com/CapeLeidokos/Kaleidoscope-XKeymaps

XKeymaps allows to replace Kaleidoscope's traditional layer stack with a layer tree. Every node in this tree is represented by an integer id. This means that when traversing the layers, the ids are not contiguously decaying towards zero but they form an arbitrary sequence of unsigned integers whose last member is zero.

The only thing that it requires to couple this approach with Kaleidoscope's existing layer class is to replace `--layer` with a call to a newly introduced function `nextLowerLayer(layer)`. 

The default implementation of `nextLowerLayer` still resembles `--layer`. Thus, this change is fully backward compatible.

### DIFF
--- a/src/kaleidoscope/layers.cpp
+++ b/src/kaleidoscope/layers.cpp
@@ -120,7 +120,7 @@ void Layer_::updateActiveLayers(void) {
             break;
           }
         }
-        layer--;
+        layer = nextLowerLayer(layer);
       }
     }
   }
@@ -202,6 +202,11 @@ void Layer_::activateNext(void) {
 
 void Layer_::deactivateTop(void) {
   deactivate(top_active_layer_);
+}
+
+__attribute__((weak))
+uint8_t nextLowerLayer(uint8_t layer) {
+  return layer - 1;
 }
 
 }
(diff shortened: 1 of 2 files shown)


noseglasses:
To me `previousLayer` sounds something like _the layer that I came from_, which is not what the function is intended to do. Upper and lower to me sounds more reasonable in the context of stacks and trees, although in a tree structure like the one I am going to introduce with my XKeymaps plugin, only the lower direction, i.e. the one towards the tree root, is uniquely defined. But let's see what Jesse says.

noseglasses:
After thinking about this further and experimenting with my implementation, it turns out that it is not possible to implement a layer tree without further changes to the layer class. This is due to the current handling of the top active layer and the layer evaluation algorithm. A layer tree would require a fundamentally different implementation of the layer class itself.

I'm closing this PR for now and will go with the traditional stack based approach until the layer class will hopefully become a plugin in the future.

Issue #613, opened 2019-03-18
Dynamic livetime for LEDModes 


noseglasses:
This is not an issue but rather a proposal.

As LEDModes are used exclusively, why not let them have dynamic lifetime? This would safe a lot of RAM and would not be too complicated given the fact that all information about LEDModes (type, etc.) is available at compile time in the sketch. 

The only major change this change would require is to not have LEDModes being plugins but rather let plugins handle lifetime of led modes through abstract factories.

```cpp
// Use the abstract factory design pattern

class LEDModeFactory_ {
   public:
      virtual ~LEDModeFactory_() {}
      virtual LEDMode *generate(void *raw_buffer) const = 0;
};

template<typename LEDMode__>
class LEDModeFactory : public LEDModeFactory_
{
   public:
      virtual LEDMode *generate(void *raw_buffer) const override { 
         return new (raw_buffer) LEDMode__{}; 
      }
};

// The base class
//
class Plugin {
   ...
   public:

      static constexpr bool hasLEDMode = false; // By default plugins do not have LEDModes
};

// Derived plugin classes
//
class APlugin : public Plugin {
   ...
   public:

      // All the listed class content below could easily be wrapped up in a macro, say
      // SERVICE_LED_MODE(ALEDMode)

      // By providing a boolean flag, we simplify template separation of 
      // LEDMode and non-LEDMode plugins
      //
      static constexpr bool hasLEDMode = true;

      // This constexpr method enables access to pointers to led mode factories
      // of global plugin instances and let them be stored in PROGMEM
      //
      constexpr const LEDModeFactory_ *getLEDModeFactory() { return led_mode_factory_; }

   private: 
      LEDModeFactory<ALEDMode> led_mode_factory_;
};
```

I could think of an interface
```cpp
void activateLEDMode(uint8_t mode_id);
LEDMode *getCurrentLEDMode();
```

A possible implementation could be to determine the maximum storage requirement by checking all available LEDMode types at compile time. Then a common buffer would be statically allocated and placement-new could be used to generate new LEDModes. As those are polymorphic classes they would have a virtual destructor anyway.

Roughly the code would look as follows
```cpp
constexpr int max_led_mode_size = ...; // Some template and macro magic based on
                                       // the list of supplied plugins
uint8_t led_mode_buffer[max_led_mode_size];
LEDMode *cur_led_mode = nullptr;

// A container class for LEDModeFactory_ pointers
[...]

PR #617, opened 2019-03-26
Introduced dynamic LED modes

noseglasses:
This PR extends the capability of https://github.com/keyboardio/Kaleidoscope/pull/612. I opened a separate PR as the code largely differs. Explanations of some of the concepts used in this PR can be found in #612 and https://github.com/keyboardio/Kaleidoscope/issues/613.

# Dynamic LED Modes with Kaleidoscope

## Purpose of this Change

This PR introduces the concept of dynamic LED modes. Those are LED modes whose class instances
have a restricted lifetime that lasts only as long as a LED mode is active. By this means
it is possible to support a greater amount of LED modes - especially RAM-hungry ones - in the same firmware build. The amount of RAM used to store dynamic LED modes is now bounded
by the maximum size (`sizeof(...)`) of the largest dynamic LED mode.

Old-style LED modes are furtheron called _static_ in the terminology of this PR. They are still supported and blend in nicely with the newly introduced dynamic LED modes.

All changes are entirely backward compatible. No user sketches or existing user plugins require changes.

## Benefits

The greatest benefit of this change is that it drastically reduces the consumption of RAM
when multiple complex LED modes are used. Currently the most complex stock LED mode is
the wavepool effect. Its plugin requires around 140 bytes of RAM that are statically allocated and cannot be shared with any other features.
With this change it becomes possible to have a large number of such resource-hungry LED modes in parallel without a significant gain in RAM consumption.

## Resource Consumption

For the stock firmware this change means a small (~30 byte) growth in terms of PROGMEM. On the other hand it reduces the amount of statically consumed RAM by ~90 bytes. As the current atmel architectures come with around ten times as much PROGMEM as RAM, this means a great improvement as RAM is the more critical resource.

If the wavepool effect, a especially RAM-hungry LED mode is added to the stock firmware,
the saving of RAM increases to 160 bytes which is almost 8% of RAM of the Keyboardio Model01.

### `LEDModeInterface`

A new interface class `LEDModeInterface` was introduced that those plugins
that export dynamic LED modes inherit from. To remain backward compatible, the `LEDMode` class that all pre-existing LED mode plugins inherited from is also derived from `LEDModeInterface`.

The new interface class currently lives in header `LEDMode.h` (see information about this new header below). This is because `LEDMode` and `LEDModeInterface` will
always be used together by dynamic LED modes. Thus, an extra header for `LEDModeInterface` would only mean extra include work for users writing plugins.

### Dynamic LED Modes

Those plugins that export dynamic LED modes must furtheron provide a exported type `DynamicLEDMode`.
This can either be done by defining a nested class of that name or by typedef-ing a class that is defined at global scope to `DynamicLEDMode`. See the modified stock LED modes for examples.

### `ACCESS_THIS_LED_MODE`

Some of those plugins that export dynamic led modes require access to their particular
dynamic LED mode. By adding the macro `ACCESS_THIS_LED_MODE` to the plugin class definition,
additional data and methods (an integer `led_mode_id_) are synthesized, that enable the plugin class to gain access to their particular dynamic LED mode instance (as long as it is active).

The synthesized integer member `led_mode_id_` can be used to query if the currently active LED mode is the oned handled by the plugin class instance (note that there might be more than one plugin instance of the same class and thus also several dynamic LED modes, see e.g. the solid color LED mode).

A query in the plugin's event handler e.g. looks as follows.
```cpp
  if (::LEDControl.get_mode_index() != led_mode_id_)
```

### Stock LED Modes

All stock LED modes have been adapted to export dynamic LED modes (if possible).
This does not apply to all of them as for some the transition would have provided no gain.
It would even have meant a deterioration of resource consumption for those few pre-existing stock LED mode plugins that hardly have no (static) data-members at all (like e.g. `LEDOff`).

### `LEDMode`

To reduce the amount of compile unit and header interdependencies, the class `LEDMode` has been moved to a header/implementation file of its own.

### `LEDControl`

The `LEDControl` class now does not have a static array anymore to store LED mode pointers.
Instead, it delegates the core LED mode handling to a newly introduced `LEDModeManager` class
that lives in internal namespace. The `LEDModeManager` class is there to restrict access
to LED modes but also to wrap up core LED mode handling. If this functionality would
have been added to class `LEDControl`, far too much of the internals of LED mode handling would have been exposed to users through header `LEDControl.h`.

### `array_like_storage.h`

The new internal header `array_like_storage.h` contains a template class that is used to generate
array-like storages. Here array-like means that the contained pieces of information
are stored contiguously in memory in the same way as they would be when defining
language intrinsic (C-style) arrays. This type of storage is especially useful to generate array-like data struktures
in PROGMEM at compile time based on a list of global objects or POD data. By casting the array-like storage's address
to the content's pointer type, an array-like indexed access is possible.

In this PR an array-like data structure is used to generate a PROGMEM
array of LED mode factories. Array-like data structures could also become useful in other places and for future applications.

## Implementation

The most complex part of the implementation of the new LED mode handling is wrapped up in
`LEDModeManager.h` and `LEDModeManager.cpp` to hide it from users' site.

There, recursive template classes are used to setup an array-like data structure of `LEDModeFactory` instances in PROGMEM. Each of the stored `LEDModeFactory`s are associated with one LED mode-plugin as specified in the sketch. The template mechanism filters out any other plugins unrelated to LED modes. `LEDModeFactory`s thereby handle both static and dynamic LED modes.

Class `LEDModeManager` provides access to the LED mode factories and LED modes in general. It exports methods to query the number of LED modes and to activate a LED mode by its mode-ID. Most of this is only available to `LEDControl` that represents the actual user interface.

When a dynamic LED mode is activated, a dedicated `LEDModeFactory` generates an instance of the dynamic LED mode class in the
LED mode buffer. This buffer is shared by all dynamic LED modes. Its size has been determined at compile time by examining all exported dynamic LED mode types and determining the maximum necessary amount of RAM to store any of those.

All LED mode handling related data structures are generated at compile time, based on
the list of plugins that are passed to `KALEIDOSCOPE_INIT_PLUGINS(...)`. This function macro invokes a new function macro `_INIT_LED_MODE_MANAGER` from `LEDModeManager.h` that handles the LED mode related stuff.

## Change Report

Consult the following change report for information about how this PR affects RAM and PROGMEM consumption.

[kaleidoscope_change_report_dynamic_LED_modes.pdf](https://github.com/keyboardio/Kaleidoscope/files/3008128/kaleidoscope_change_report_dynamic_LED_modes.pdf)

### DIFF
--- a/src/kaleidoscope/plugin/Heatmap.h
+++ b/src/kaleidoscope/plugin/Heatmap.h
@@ -22,7 +22,9 @@
 
 namespace kaleidoscope {
 namespace plugin {
-class Heatmap : public LEDMode {
+class Heatmap : public Plugin,
+  public LEDModeInterface,
+  public AccessTransientLEDMode {
  public:
   Heatmap(void) {}
 
@@ -34,16 +36,38 @@ class Heatmap : public LEDMode {
   EventHandlerResult onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state);
   EventHandlerResult beforeEachCycle();
 
- protected:
-  void update(void) final;
+  // This class' instance has dynamic lifetime
+  //
+  class TransientLEDMode : public LEDMode {
+   public:
 
- private:
-  static uint16_t heatmap_[ROWS][COLS];
-  static uint16_t highest_;
-  static uint32_t next_heatmap_comp_time_;
+    // Please note that storing the parent ptr is only required
+    // for those LED modes that require access to
+    // members of their parent class. Most LED modes can do without.
+    //
+    TransientLEDMode(const Heatmap *parent);
 
-  static void shiftStats(void);
-  static cRGB computeColor(float v);
+    void resetMap();
+    EventHandlerResult onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state);
+    EventHandlerResult beforeEachCycle();
+
+   protected:
+
+    virtual void update() final;
+
+   private:
+
+    const Heatmap *parent_;
+
+    uint16_t heatmap_[ROWS][COLS];
+    uint16_t highest_;
+    uint32_t next_heatmap_comp_time_;
+
+    void shiftStats(void);
+    cRGB computeColor(float v);
+
+    friend class Heatmap;
+  };
 };
 }
 }
--- a/src/kaleidoscope/plugin/LED-Wavepool.cpp
+++ b/src/kaleidoscope/plugin/LED-Wavepool.cpp
@@ -27,52 +27,64 @@ namespace plugin {
 #define MS_PER_FRAME 40  // 40 = 25 fps
 #define FRAMES_PER_DROP 120  // max time between raindrops during idle animation
 
-int8_t WavepoolEffect::surface[2][WP_WID * WP_HGT];
-uint8_t WavepoolEffect::page = 0;
-uint8_t WavepoolEffect::frames_since_event = 0;
 uint16_t WavepoolEffect::idle_timeout = 5000;  // 5 seconds
 int16_t WavepoolEffect::ripple_hue = WavepoolEffect::rainbow_hue; // automatic hue
 
 // map native keyboard coordinates (16x4) into geometric space (14x5)
-PROGMEM const uint8_t WavepoolEffect::rc2pos[ROWS * COLS] = {
+PROGMEM const uint8_t WavepoolEffect::TransientLEDMode::rc2pos[ROWS * COLS] = {
   0,  1,  2,  3,  4,  5,  6,     59, 66,    7,  8,  9, 10, 11, 12, 13,
   14, 15, 16, 17, 18, 19, 34,    60, 65,   35, 22, 23, 24, 25, 26, 27,
   28, 29, 30, 31, 32, 33, 48,    61, 64,   49, 36, 37, 38, 39, 40, 41,
   42, 43, 44, 45, 46, 47,     58, 62, 63, 67,    50, 51, 52, 53, 54, 55,
 };
 
+WavepoolEffect::TransientLEDMode::TransientLEDMode(const WavepoolEffect *parent)
+  : parent_(parent),
+    frames_since_event_(0),
+    surface_{},
+    page_(0)
+{}
+
 EventHandlerResult WavepoolEffect::onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state) {
   if (row >= ROWS || col >= COLS)
     return EventHandlerResult::OK;
 
+  if (::LEDControl.get_mode_index() != led_mode_id_)
+    return EventHandlerResult::OK;
+
+  return ::LEDControl.get_mode<TransientLEDMode>()
+         ->onKeyswitchEvent(mapped_key, row, col, key_state);
+}
+
+EventHandlerResult WavepoolEffect::TransientLEDMode::onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state) {
   if (keyIsPressed(key_state)) {
     uint8_t offset = (row * COLS) + col;
-    surface[page][pgm_read_byte(rc2pos + offset)] = 0x7f;
-    frames_since_event = 0;
+    surface_[page_][pgm_read_byte(rc2pos + offset)] = 0x7f;
+    frames_since_event_ = 0;
   }
 
   return EventHandlerResult::OK;
 }
 
-void WavepoolEffect::raindrop(uint8_t x, uint8_t y, int8_t *page) {
+void WavepoolEffect::TransientLEDMode::raindrop(uint8_t x, uint8_t y, int8_t *page_) {
   uint8_t rainspot = (y * WP_WID) + x;
 
-  page[rainspot] = 0x7f;
-  if (y > 0) page[rainspot - WP_WID] = 0x60;
-  if (y < (WP_HGT - 1)) page[rainspot + WP_WID] = 0x60;
-  if (x > 0) page[rainspot - 1] = 0x60;
-  if (x < (WP_WID - 1)) page[rainspot + 1] = 0x60;
+  page_[rainspot] = 0x7f;
+  if (y > 0) page_[rainspot - WP_WID] = 0x60;
+  if (y < (WP_HGT - 1)) page_[rainspot + WP_WID] = 0x60;
+  if (x > 0) page_[rainspot - 1] = 0x60;
+  if (x < (WP_WID - 1)) page_[rainspot + 1] = 0x60;
 }
 
 // this is a lot smaller than the standard library's rand(),
 // and still looks random-ish
-uint8_t WavepoolEffect::wp_rand() {
+uint8_t WavepoolEffect::TransientLEDMode::wp_rand() {
   static uint16_t offset = 0x400;
   offset = ((offset + 1) & 0x4fff) | 0x400;
   return (millis() / MS_PER_FRAME) + pgm_read_byte(offset);
 }
 
-void WavepoolEffect::update(void) {
+void WavepoolEffect::TransientLEDMode::update(void) {
 
   // limit the frame rate; one frame every 64 ms
   static uint8_t prev_time = 0;
@@ -89,11 +101,11 @@ void WavepoolEffect::update(void) {
   static uint8_t current_hue = 0;
   current_hue ++;
 
-  frames_since_event ++;
+  frames_since_event_ ++;
 
   // needs two pages of height map to do the calculations
-  int8_t *newpg = &surface[page ^ 1][0];
-  int8_t *oldpg = &surface[page][0];
+  int8_t *newpg = &surface_[page_ ^ 1][0];
+  int8_t *oldpg = &surface_[page_][0];
 
   // rain a bit while idle
   static uint8_t frames_till_next_drop = 0;
@@ -112,11 +124,11 @@ void WavepoolEffect::update(void) {
       raindrop(prev_x, prev_y, oldpg);
       prev_x = prev_y = -1;
     }
-    if (frames_since_event
+    if (frames_since_event_
         >= (frames_till_next_drop
             + (idle_timeout / MS_PER_FRAME))) {
       frames_till_next_drop = 4 + (wp_rand() % FRAMES_PER_DROP);
-      frames_since_event = idle_timeout / MS_PER_FRAME;
+      frames_since_event_ = idle_timeout / MS_PER_FRAME;
 
       uint8_t x = wp_rand() % WP_WID;
       uint8_t y = wp_rand() % WP_HGT;
@@ -195,7 +207,7 @@ void WavepoolEffect::update(void) {
       uint8_t value = (intensity >= 128) ? 255 : intensity << 1;
       int16_t hue = ripple_hue;
 
-      if (ripple_hue == rainbow_hue) {
+      if (ripple_hue == WavepoolEffect::rainbow_hue) {
         // color starts white but gets dimmer and more saturated as it fades,
         // with hue wobbling according to height map
         hue = (current_hue + height + (height >> 1)) & 0xff;
@@ -209,10 +221,10 @@ void WavepoolEffect::update(void) {
 
 #ifdef INTERPOLATE
   // swap pages every other frame
-  if (!(now & 1)) page ^= 1;
+  if (!(now & 1)) page_ ^= 1;
 #else
   // swap pages every frame
-  page ^= 1;
+  page_ ^= 1;
 #endif
 
 }
--- a/src/kaleidoscope/plugin/LEDControl.h
+++ b/src/kaleidoscope/plugin/LEDControl.h
@@ -18,88 +18,20 @@
 
 #include <Kaleidoscope.h>
 
-#define LED_MAX_MODES 24
-
 #define LED_TOGGLE   B00000001  // Synthetic, internal
 
 #define Key_LEDEffectNext Key(0, KEY_FLAGS | SYNTHETIC | IS_INTERNAL | LED_TOGGLE)
 #define Key_LEDEffectPrevious Key(1, KEY_FLAGS | SYNTHETIC | IS_INTERNAL | LED_TOGGLE)
 
+#define _DEPRECATED_MESSAGE_LED_CONTROL_MODE_ADD                               \
+  "LEDControl::mode_add(LEDMode *mode) is deprecated. LEDModes are now \n"     \
+  "automatically registered. You can safely remove any calls to \n"            \
+  "LEDControl::mode_add from your code."
+
 namespace kaleidoscope {
 namespace plugin {
-/** Base class for LED modes.
- *
- * LED modes are a special kind of plugin, they are in charge of updating LED
- * colors, setting a theme. While it is possible to have other plugins
- * override the mode's colors, the LED mode is the baseline.
- *
- * Most of its functionality is called via @ref LEDControl, with only a few
- * public methods.
- *
- * A LED mode **must** implement at least one of @ref onActivate or @ref
- * update, and possibly @ref refreshAt too.
- */
-class LEDMode : public kaleidoscope::Plugin {
-  friend class LEDControl;
- protected:
-  // These methods should only be called by LEDControl.
 
-  /** One-time setup, called at keyboard boot.
-   *
-   * Any hooks that need registering, any one-time setup that needs to be
-   * performed, shall be done here. This is purely for preparation purposes, the
-   * LEDs should not be touched yet at this time.
-   */
-  virtual void setup(void) {}
-
-  /** Function to call whenever the mode is activated.
-   *
-   * Like @ref setup, this method need not touch LEDs, @ref update will be
-   * called right after it. The purpose of this callback is to allow a plugin to
-   * do some preparation whenever it is activated, instead of only on boot, or
-   * always at each cycle.
-   *
-   * However, unlike @ref setup, this method can change LED colors, if so
-   * desired. Either to provide an initial state, or a static color set. In the
-   * latter case, consider implementing @ref refreshAt too, because other
-   * plugins may override some of the colors set at activation time, and @ref
-   * refreshAt can be used to restore them when needed.
-   *
-   * Before the callback runs, LEDs will be blanked.
-   */
-  virtual void onActivate(void) {}
-
-  /** Update the LEDs once per cycle.
-   *
-   * Usually the brains of the plugin, which updates the LEDs each cycle. It is
-   * called after the matrix has been scanned, once per cycle.
-   */
-  virtual void update(void) {}
-
-  /** Refresh the color of a given key.
-   *
-   * If we have another plugin that overrides colors set by the active LED mode
-   * (either at @onActivate time, or via @ref update), if that plugin wants to
-   * restore whatever color the mode would set the key color to, this is the
-   * method it will call.
-   *
-   * @param row is the row coordinate of the key to refresh the color of.
-   * @param col is the column coordinate of the key to refresh the color of.
-   */
-  virtual void refreshAt(byte row, byte col) {}
-
- public:
-  /** Activate the current object as the LED mode.
-   */
-  void activate(void);
-
-  /** Plugin initialization.
-   *
-   * Called via `Kaleidoscope.use()`, registers the LED mode, and does the
-   * necessary initialization steps. Calls @ref setup at the end.
-   */
-  kaleidoscope::EventHandlerResult onSetup();
-};
+class LEDMode;
 
 class LEDControl : public kaleidoscope::Plugin {
  public:
@@ -112,19 +44,26 @@ class LEDControl : public kaleidoscope::Plugin {
     if (!Kaleidoscope.has_leds)
       return;
 
-    if (modes[mode])
-      modes[mode]->update();
+    cur_led_mode_->update();
   }
   static void refreshAt(byte row, byte col) {
     if (!Kaleidoscope.has_leds)
       return;
 
-    if (modes[mode])
-      modes[mode]->refreshAt(row, col);
+    cur_led_mode_->refreshAt(row, col);
   }
-  static void set_mode(uint8_t mode);
-  static uint8_t get_mode_index();
-  static LEDMode *get_mode();
+  static void set_mode(uint8_t mode_id);
+  static uint8_t get_mode_index() {
+    return mode_id;
+  }
+  static LEDMode *get_mode() {
+    return cur_led_mode_;
+  }
+  template<typename LEDMode__>
+  static LEDMode__ *get_mode() {
+    return static_cast<LEDMode__*>(cur_led_mode_);
+  }
+
   static void refreshAll() {
     if (!Kaleidoscope.has_leds)
       return;
@@ -133,11 +72,14 @@ class LEDControl : public kaleidoscope::Plugin {
       return;
 
     set_all_leds_to({0, 0, 0});
-    if (modes[mode])
-      modes[mode]->onActivate();
+
+    cur_led_mode_->onActivate();
   }
 
-  static int8_t mode_add(LEDMode *mode);
+  DEPRECATED(LED_CONTROL_MODE_ADD)
+  static int8_t mode_add(LEDMode *mode) {
+    return 0;
+  }
 
   static void setCrgbAt(int8_t i, cRGB crgb);
   static void setCrgbAt(byte row, byte col, cRGB color);
@@ -147,7 +89,11 @@ class LEDControl : public kaleidoscope::Plugin {
   static void set_all_leds_to(uint8_t r, uint8_t g, uint8_t b);
   static void set_all_leds_to(cRGB color);
 
-  static void activate(LEDMode *mode);
+  // We restict activate to LEDModeInterface to make sure that
+  // a compiler error is thrown when activate() is accidentally
+  // applied to a non-LED mode plugin.
+  //
+  static void activate(LEDModeInterface *plugin);
 
   static uint16_t syncDelay;
   static bool paused;
@@ -158,8 +104,9 @@ class LEDControl : public kaleidoscope::Plugin {
 
  private:
   static uint16_t syncTimer;
-  static LEDMode *modes[LED_MAX_MODES];
-  static uint8_t mode;
+  static uint8_t mode_id;
+  static uint8_t num_led_modes_;
+  static LEDMode *cur_led_mode_;
 };
 
 class FocusLEDCommand : public Plugin {
--- a/src/kaleidoscope/plugin/LEDMode.h
+++ b/src/kaleidoscope/plugin/LEDMode.h
@@ -1,0 +1,144 @@
+/* Kaleidoscope-LEDControl - LED control plugin for Kaleidoscope
+ * Copyright (C) 2017-2018  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "kaleidoscope/plugin.h"
+
+namespace kaleidoscope {
+
+namespace internal {
+// Forward declaration
+class LEDModeManager;
+} // end namespace internal
+
+namespace plugin {
+
+class LEDModeInterface {
+ public:
+
+  void activate();
+
+  // This auxiliary class helps to generate a verbose error message
+  // in case that there is no TransientLEDMode typedef or nested
+  // class present in a derived class of LEDModeInterface.
+  //
+  struct _____LEDModeInterface_derived_class_is_missing_a_TransientLEDMode_typedef_or_nested_class_____ {};
+  typedef _____LEDModeInterface_derived_class_is_missing_a_TransientLEDMode_typedef_or_nested_class_____ NoLEDMode;
+
+  // By redefining TransientLEDMode, derived plugins export
+  // a LED mode that becomes part of the set of LED modes whose
+  // lifetime is handled dynamically.
+  //
+  typedef NoLEDMode DynamicLEDMode;
+};
+
+class AccessTransientLEDMode {
+ public:
+
+  // This method is called when a plugin's LED mode is activated.
+  // Derived plugins may reimplement it to store the id of their
+  // exported LED mode. A plugin can thus check
+  // whether their LED mode is currently active.
+  //
+  void registerLEDModeActivated(uint8_t led_mode_id) {
+    led_mode_id_ = led_mode_id;
+  }
+
+ protected:
+
+  uint8_t led_mode_id_ = 255; /* 255 means uninitialized */
+};
+
+/** Base class for LED modes.
+*
+* LED modes are a special kind of plugin, they are in charge of updating LED
+* colors, setting a theme. While it is possible to have other plugins
+* override the mode's colors, the LED mode is the baseline.
+*
+* Most of its functionality is called via @ref LEDControl, with only a few
+* public methods.
+*
+* A LED mode **must** implement at least one of @ref onActivate or @ref
+* update, and possibly @ref refreshAt too.
+*/
+class LEDMode : public kaleidoscope::Plugin,
+  public LEDModeInterface {
+  friend class LEDControl;
+  friend class kaleidoscope::internal::LEDModeManager;
+ protected:
+  // These methods should only be called by LEDControl.
+
+  /** One-time setup, called at keyboard boot.
+   *
+   * Any hooks that need registering, any one-time setup that needs to be
+   * performed, shall be done here. This is purely for preparation purposes, the
+   * LEDs should not be touched yet at this time.
+   */
+  virtual void setup(void) {}
+
+  /** Function to call whenever the mode is activated.
+   *
+   * Like @ref setup, this method need not touch LEDs, @ref update will be
+   * called right after it. The purpose of this callback is to allow a plugin to
+   * do some preparation whenever it is activated, instead of only on boot, or
+   * always at each cycle.
+   *
+   * However, unlike @ref setup, this method can change LED colors, if so
+   * desired. Either to provide an initial state, or a static color set. In the
+   * latter case, consider implementing @ref refreshAt too, because other
+   * plugins may override some of the colors set at activation time, and @ref
+   * refreshAt can be used to restore them when needed.
+   *
+   * Before the callback runs, LEDs will be blanked.
+   */
+  virtual void onActivate(void) {}
+
+  /** Update the LEDs once per cycle.
+   *
+   * Usually the brains of the plugin, which updates the LEDs each cycle. It is
+   * called after the matrix has been scanned, once per cycle.
+   */
+  virtual void update(void) {}
+
+  /** Refresh the color of a given key.
+   *
+   * If we have another plugin that overrides colors set by the active LED mode
+   * (either at @onActivate time, or via @ref update), if that plugin wants to
+   * restore whatever color the mode would set the key color to, this is the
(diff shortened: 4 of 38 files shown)


src/kaleidoscope/plugin/LEDControl.h
```diff
@@ -158,8 +100,9 @@ class LEDControl : public kaleidoscope::Plugin {
 
  private:
   static uint16_t syncTimer;
-  static LEDMode *modes[LED_MAX_MODES];
-  static uint8_t mode;
+  static uint8_t mode_id;
```
gedankenexperimenter: This should be named `mode_id_`.

src/kaleidoscope/plugin/LED-Wavepool.cpp
```diff
@@ -27,52 +27,64 @@ namespace plugin {
...
+         ->onKeyswitchEvent(mapped_key, row, col, key_state);
+}
+
+EventHandlerResult WavepoolEffect::DynamicLEDMode::onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state) {
   if (keyIsPressed(key_state)) {
     uint8_t offset = (row * COLS) + col;
-    surface[page][pgm_read_byte(rc2pos + offset)] = 0x7f;
-    frames_since_event = 0;
+    surface_[page_][pgm_read_byte(rc2pos + offset)] = 0x7f;
+    frames_since_event_ = 0;
```
gedankenexperimenter: I know I'm usually the one asking for more conformity with the coding style guides, but these name changes are unrelated to the rest of this very large PR, which is difficult enough to decipher as it is.

gedankenexperimenter:
I can't easily tell if this is orthogonal to, or if it obviates the your proposal to store the array of LED mode pointers in PROGMEM. Is that included here, but it's changed enough that it's hard to recognize?

Also, if you were unfettered by backward compatibility constraints, is this the way you would do it, or did you make some compromises for the sake of not requiring changes to user sketches?

noseglasses:
I started with a version that was based on #612, but it turned out that it is much more efficient to address both issues together. That's why this PR does both, it moves `LEDControl`'s `LEDMode*` array to PROGMEM and introduces dynamic LED modes. For both it uses the same optimized datastructures and algorithms.

Backwards compatibility comes at zero cost. This is because for few of the very simple LED modes there is no benefit in making them dynamic. The code for leaving them static for efficiency reasons coincides with the code that would otherwise be required to stay backward compatible.

src/kaleidoscope/plugin/LED-Wavepool.cpp
noseglasses: As (technically) I had to introduce new classes, I implemented them according to the rules of the style guide.

src/kaleidoscope/plugin/LEDControl.h
noseglasses: I was tempted to adapt the `LEDControl` class to be in accordance with the style guide but i resisted. The only symbol name change I made was to rename `mode` to `mode_id` because of a name conflict. The rest remains to be done in a separate PR that will clean up `LEDControl`.

gedankenexperimenter:
> That's why this PR does both, it moves `LEDControl`'s `LEDMode*` array to PROGMEM and introduces dynamic LED modes.

Would you mind pointing out where this array is in the code now? I can't find it.

src/kaleidoscope/plugin/LEDControl.h
gedankenexperimenter: Okay, but since you changed the name anyway, why not make it compliant now? There's another spot i noticed where `mode_id` is used as the name of a function parameter, and this makes it more confusing.

src/kaleidoscope/plugin/LED-Wavepool.cpp
gedankenexperimenter: If this is a new class, why is the diff showing only these variable names as having changed?

src/kaleidoscope_internal/LEDModeManager.h
```diff
@@ -0,0 +1,428 @@
...
+                  _LED_MODE_MANAGEMENT__PLUGIN_CONTROLS_LED_MODE,       __NL__ \
+                  __VA_ARGS__                                           __NL__ \
+               )                                                        __NL__ \
+      > ArrayType;                                                      __NL__ \
+                                                                        __NL__ \
+      /* Generate the actual instance of template class                 __NL__ \
+       * TypedPluginArray and initialize it with a list of              __NL__ \
+       * LEDModeFactories.                                              __NL__ \
+       */                                                               __NL__ \
+      const PROGMEM ArrayType led_mode_factories(                       __NL__ \
```
noseglasses: >> That's why this PR does both, it moves LEDControl's LEDMode* array to PROGMEM and introduces dynamic LED modes.

> Would you mind pointing out where this array is in the code now? I can't find it.

The former `LEDMode` pointers, now `Plugin` pointers are stored in the `led_mode_factories` pseudo-array. For every LED mode related plugin instance the array stores a pointer to the instance and - if a dynamic LED mode - a pointer to a generator function (`generateLEDMode<...>`). This function generates the actual dynamic LED mode in the shared buffer. As it is templated, the factory function knows the exact type of the dynamic LED to create. For static LED modes, no factory function is required as the LED mode managing plugin is by itself a LED mode. See the implementation of `LEDModeManager::getLEDMode(...)` to see where this distinction is considered.

All this is an implementation of the [factory method](https://en.wikipedia.org/wiki/Factory_method_pattern) design pattern.

gedankenexperimenter:
Can you provide step-by-step instructions for generating the post-preprocessor code that gets generated by the macro functions? The actual code that's being generated is too well-hidden for me to figure out what's going on.

noseglasses:
A recipe to step-by-step unroll the pre-processor macro replacements i had provided already here: https://github.com/keyboardio/Kaleidoscope/pull/612#issuecomment-474394413

Is that what you were asking for?

gedankenexperimenter:
> A recipe to step-by-step unroll the pre-processor macro replacements i had provided already here: #612 (comment)

> Is that what you were asking for?

No, that's not a set of step-by-step instructions; it's missing information on where to add the `-E` flag (what file does one put it into, and where, exactly?), exactly what commands to execute, and where to look for the preprocessed code after those commands finish.

noseglasses:
Err... alright.

1. cd to the sketch directory `.../hardware/keyboardio/avr/libraries/Model01-Firmware`
2. Execute `VERBOSE=1 BOARD_HARDWARE_PATH=$PWD/../../../.. make  > build_output.txt 2>&1`
3. Open `build_output.txt` in a text editor
4. Search for a line that reads `Compiling sketch...`
5. The line that follows is the compile command. Copy it to your console.
6. Add `.i -E` to the end of the compiler line (the generated file now is named `..../Model01-Firmware.ino.cpp.o.i`)

Don't forget about the `__NL__` trick.
7. Execute the build command
8. Open the generated file `.../Model01-Firmware.ino.cpp.o.i` in your favorit editor

src/kaleidoscope/plugin/LED-Wavepool.cpp
gedankenexperimenter: I do see your point. However, these name changes make it even more difficult to figure out what the relevant changes are for this PR. It would be much clearer if you at least separated these name changes into a different commit, and even better if they were in a separate PR entirely, since they affect plenty of lines of code that are not related to the dynamic LED mode changes.

gedankenexperimenter:
Thank you. _That_ was an excellent set of step-by-step instructions, and I can now examine the preprocessed code. One note to others trying this: `BOARD_HARDWARE_PATH` shouldn't necessarily be set relative to the sketch directory (and if you've set things up the "normal" way, it's not necessary to set it at all), and the location of the generated `Model01-Firmware.ino.cpp.o.i` file is system-dependent. It's easiest to replace the whole filename with something of your choice, rather than let it be written in whatever temp directory the *.o file would normally get placed in.

src/kaleidoscope_internal/LEDModeManager.h
```diff
@@ -0,0 +1,428 @@
...
+// cost of vtables which safes some program memory.
+// Conceptually, the factory function approach is similar to having
+// a polymorphic class with only one method, that is accessed directly
+// instead of the detour via the vtable.
+//
+typedef kaleidoscope::plugin::LEDMode*
+(*LEDModeFactoryFunc)(void *raw_buffer,
+                      void *parent,
+                      uint8_t mode_id
+                     );
```
gedankenexperimenter: Can you explain why the `typedef` is here?

noseglasses: Because function pointer types are ugly to read, so you only want to write them once and then go with the typedef-ed type.

src/kaleidoscope_internal/LEDModeManager.h
```diff
@@ -0,0 +1,428 @@
...
+template<typename ParentPlugin__>
+static kaleidoscope::plugin::LEDMode *
+generateLEDMode(void *raw_buffer,
+                void *parent_plugin,
+                uint8_t mode_id
+               ) {
+  // We know the type of the parent plugin via the template parameter
+  // ParentPlugin__, thus it is safe to cast to the actual type.
+  //
+  auto parent_plugin_actual = static_cast<ParentPlugin__*>(parent_plugin);
```
gedankenexperimenter: I'm guessing that the reason `parent_plugin` is a generic `void*` pointer is because the caller is reading it from PROGMEM, and doesn't bother to do the conversion before calling `generateLEDMode()`. Is that right?

noseglasses:
> BOARD_HARDWARE_PATH shouldn't necessarily be set relative to the sketch directory (and if you've set things up the "normal" way, it's not necessary to set it at all)

There is no _normal_ way. When working with a higher number of copies of the bundle, that's the most simple way to build without changing symbolic links. Are you aware of any other solution?

src/kaleidoscope_internal/LEDModeManager.h
```diff
@@ -0,0 +1,428 @@
...
+
+  typedef typename ParentPlugin__::DynamicLEDMode DynamicLEDMode;
```
gedankenexperimenter: Is this `typedef` here for some reason other than shortening the line below where it is used? I think this code would actually be clearer if you just used `ParentPlugin__::DynamicLEDMode{parent_plugin_actual}` instead of the generic-looking `DynamicLEDMode{parent_plugin_actual}` there, and dropped this line (unless there's something more going on here that's not apparent to me).

src/kaleidoscope_internal/LEDModeManager.h
noseglasses: No, it's because `LEDModeManager` is designed entirely type agnostic and thus generic with respect to plugins and LED modes. Only the factory functions know the actual type of the plugins they operate on. 

This mechanism would work the same if every plugin that deals with LED modes would have its own static `LEDModeFactoryFunc` implementation. Those pointers could as well be stored in the `LEDModeFactory`s. 

The reason why I went with the template function based approach is that I wanted to spare the plugin developer to implement such functions or add some magic macros that would implement them. Templates are just easier to handle in terms of transparency and compiler errors than magic macros.

src/kaleidoscope_internal/LEDModeManager.h
noseglasses: Yeah, right. That's a reminiscence.

src/kaleidoscope_internal/LEDModeManager.h
gedankenexperimenter: In that case, I guess I have a more fundamental question – what does this line do? It looks like a combination of a definition of a type alias and a function declaration, and doesn't make any sense to me at all.

src/kaleidoscope_internal/LEDModeManager.h
gedankenexperimenter: If the caller already knows what `ParentPlugin__` is (maybe it would be better to call it `ParentPluginType__`), why is `parent_plugin` of type `void*` instead of `ParentPlugin__*`?

gedankenexperimenter:
> There is no _normal_ way. When working with a higher number of copies of the bundle, that's the most simple way to build without changing symbolic links. Are you aware of any other solution?

The "normal way" I alluded to is the setup described in the [README](https://github.com/keyboardio/Kaleidoscope/blob/master/README.md#install-the-kaleidoscope-hardware-definitions), which is to clone the repository, then symlink it from the standard Arduino directory. Some people prefer to do it without creating those symlinks, but I think it's fair to consider that less standard or "normal" than the instructions given on the Kaleidoscope repository's front page.

noseglasses:
The normal way is what suits best your workflow. How I handle this and the way I explained it in the step-by-step instructions is what i consider the normal way for developers that are juggling tens of copies of the firmware bundle. Also I wanted to write the step-by-step instructions in a way that nothing can go wrong. 

The Kaleidoscope frontpage is directed to consumers. Those are unlikely to end up here.

noseglasses:
Anyway, it's ok to have different concepts of _normal_.

gedankenexperimenter:
> The normal way is what suits best your workflow. How I handle this and the way I explained it in the step-by-step instructions is what i consider the normal way for developers that are juggling tens of copies of the firmware bundle. Also I wanted to write the step-by-step instructions in a way that nothing could go wrong.

The "normal" way is certainly not whatever suits an individual. And your instructions would fail for someone who uses a sketch in its own directory, separate from the bundle, which is something that can be done by someone who uses a symlink (or simply checks out the bundle where Arduino expects it).

> The Kaleidoscope frontpage is directed to consumers. Those are unlikely to end up here.

That an exclusionary attitude that I do not share. I would like the Kaleidoscope code to be as accessible (and understandable) as possible to as wide an audience as possible – even "mere consumers".

gedankenexperimenter:
> Anyway, it's ok to have different concepts of _normal_.

Thus my original use of quotes when referring to the "normal" repository setup. Perhaps that was too subtle.

noseglasses:
> And your instructions would fail for someone who uses a sketch in its own directory, separate from the bundle, which is something that can be done by someone who uses a symlink (or simply checks out the bundle where Arduino expects it).

Point 1. of my instructions is based on a path that goes down to the hardware directory. This was meant to indicate that the sketch file rests at its standard position, where the bundle repo places it. I am not sure what is supposed to go wrong here. If someone places the sketch somewhere else it will of course not work. That's why I wrote it in such a detailed fashion.

> Those are unlikely to end up here.

That's sadly a matter of fact, nothing exclusionary. I would be happy if more people would get involved with developing this lovely piece of software.

As you might have noticed, I already spend a considerable amount of time on explaining stuff to you. And I do the same for other people as well. If you are honest, you might agree that you have learned some C++ here, through my explanations and examples. Is that how you define "exclusionary"?

src/kaleidoscope_internal/LEDModeManager.h
noseglasses: That defines a typename alias for a function pointer. The syntax is different from that of ordinary (non-function pointer) typedefs. In general
```cpp
typedef ReturnType (*PtrTypeAlias)(...arguments...);
```
says that `PtrTypeAlias` can furtheron be the type of a pointer that points to a function of given type.

Example
```
typedef int (*f)(int); // function pointer alias
int f1(int a) { return a; } // function of same type

int main() {
    f my_f_1 = f1; // function pointer assignment
    f my_f_2 = &f1; // alternative syntax
    (*my_f_1)(10); // function invocation via function pointer
}
```

src/kaleidoscope_internal/LEDModeManager.h
noseglasses: > If the caller already knows what ParentPlugin__ is (maybe it would be better to call it ParentPluginType__)

I know, that's confusing. The caller does not know the type of `ParentPlugin__`. `ParentPlugin__` is a template parameter. It is only known at compile time and only in the context, where the function pointer is assigned (which is in line 202). The caller, `LEDModeManager` is completely agnostic of the plugin type. Thus, a generic implementation is possible. Whether we call it `ParentPlugin__` or `ParentPluginType__` doesn't matter. As from the template definition it is obvious that it is a type, not a variable, I tend to skip the suffix `Type`.

> why is parent_plugin of type void* instead of ParentPlugin__*?

Same answer as above, because the calling context does not know the actual type. We could change the `void *parent_plugin` to something less frightening than a void ptr, though. I will think about that and push a change with some more comments, later.

noseglasses:
I am currently working on a document that explains the whys and hows of this implementation. I will post it ASAP.

noseglasses:
Here is a document that describes this PR in greater detail: https://gist.github.com/noseglasses/194259425b508c4ffe220596a02822be

obra:
This is so well put together. I haven't digested it all yet, but <3 <3 <3.
Thank you for the hard work to put together the code and document it.
ᐧ

obra:
@algernon is in the process of setting up Kaleidoscope's website and blog.
This writeup is exactly the sort of technical content that I thin would
make a great blog post.

Best,
Jesse
ᐧ

src/kaleidoscope/plugin/Heatmap.h
```diff
@@ -34,16 +35,40 @@ class Heatmap : public LEDMode {
...
+
+    virtual void update() final;
+
+   private:
+
+    const Heatmap *parent_;
+
+    uint16_t heatmap_[ROWS][COLS];
+    uint16_t highest_;
+    uint32_t next_heatmap_comp_time_;
```
gedankenexperimenter: I'm not sure that the Heatmap plugin should operate this way. It builds its heatmap data over time, as the user types, but if it only exists while it is the active LED mode, then a user needs to have the heatmap turned on all the time in order to work. It seems more useful to me if I can have the LEDs off, or switch between modes, and still have the heatmap accumulating data while I type. Then I could switch to the Heatmap mode and see what it looks like every once in a while. @algernon – what do you think?

src/kaleidoscope/plugin/Heatmap.h
```diff
@@ -34,16 +35,40 @@ class Heatmap : public LEDMode {
...
+    // for those LED modes that require access to
+    // members of their parent class. Most LED modes can do without.
+    //
+    DynamicLEDMode(const Heatmap *parent);
 
-  static void shiftStats(void);
-  static cRGB computeColor(float v);
+    void resetMap();
+    EventHandlerResult onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state);
+    EventHandlerResult beforeEachCycle();
```
gedankenexperimenter: I'm curious about these hook functions in a dynamic LED mode. @noseglasses – how do they get called if the object doesn't always exist?

src/kaleidoscope/plugin/Heatmap.h
noseglasses: Right, I didn't think of it that way. We might better stick with the original (static) version of heatmap. Going  with a dynamic LED mode where the dynamic part only stores `next_heatmap_comp_time_` would save three bytes of permanent RAM (four bytes for `next_heatmap_comp_time_` saved minus one byte for `led_mode_id_` to enable the plugin to access its dynamic LED mode).

Let's see what @algernon says.

src/kaleidoscope/plugin/Heatmap.h
noseglasses: By adding the `ACCESS_THIS_LED_MODE` macro to the parent plugin of a dynamic LED mode. The plugin is able to be informed about the mode id of its LED mode. It can then check if its LED mode is the current one. The heatmap plugin checks this in its handlers
```cpp
EventHandlerResult Heatmap::DynamicLEDMode::onKeyswitchEvent(Key &mapped_key, byte row, byte col, uint8_t key_state) {
   ...
   // Is my plugin the current one?
   if (::LEDControl.get_mode_index() != led_mode_id_)
   ...
}
```
This would be more readable if `LEDControl`'s method would be named `get_current_mode_index` instead.

FWIW, every instance of a LED mode plugin (static and dynamic) is assigned an individual mode id (every solid color, e.g. has its own). This id is the corresponding index in the factory array.

noseglasses:
> @algernon is in the process of setting up Kaleidoscope's website and blog. This writeup is exactly the sort of technical content that I thin would make a great blog post. 

i am glad you like it and find it useful. Feel free to use it for the blog.

src/kaleidoscope/plugin/Heatmap.h
noseglasses: > how do they get called if the object doesn't always exist?

I didn't answer that part. The way I implemented it for some of the LED modes is to let the plugins handler check if its dynamic LED mode is current and then delegate the handler to the dynamic LED mode. For some the plugins, like the wavepool effect, this makes the handler code more readable as it is not cluttered with `dynamic_led_mode->...` statements. Those would be needed if the plugins handler would want to access members of its dynamic LED mode.

src/kaleidoscope/plugin/Heatmap.h
gedankenexperimenter: Yeah, it sucks that it works out this way for one of the bigger LED modes, but if the heatmap loses its data when a different mode is activated, it becomes useless for someone who wants to see the heatmap, but finds it distracting to have the LEDs on while typing.

I can think of a compromise. If Heatmap adds a boolean that controls whether the LEDs are turned on or not, it would be possible to toggle them off without turning off the dynamic LED mode. It's not a great compromise, because it wouldn't be possible for a user to turn on (for example) ActiveModColor and still have Heatmap collecting data.

gedankenexperimenter: Another idea is to have two different versions of Heatmap: one dynamic, for people who need the RAM, but don't mind if it loses its data when another mode is active, and one static, for people who can spare the RAM and want it collecting data all the time.

noseglasses: Those are both very nice ideas! For the first one we could add some sort of signal while it is recording, lets say one of the keys blinking red like a camera would do.

The second idea is also feasible. We could implement the static version of the plugin as a simple wrapper for the `LEDMode` derived class that the dynamic version exports. Thus, we could avoid code duplication.

noseglasses: FWIW, I suggested the red blinking light because I find it sometimes difficult to keep track of which of the LED modes is active. That's difficult with those that only light up LEDs as a reaction to typing.

src/kaleidoscope/plugin/Heatmap.h
gedankenexperimenter: This is the downside of using that macro; from the point of view of a plugin developer, `ACCESS_THIS_LED_MODE` is a mysterious black box, without much of a hint about where to find it. To me, it would help a great deal if this code was commented, with a reference to the file and line where the macro is defined, and a brief explanation of what it does.

Reducing the amount of boilerplate code is great, and providing a better guarantee that it's correct is even better, but needing to run a recursive `grep` in order to find all the methods and members belonging to a class whose header file I'm already looking at is a pretty big downside.

noseglasses: `ACCESS_THIS_LED_MODE` is defined in header `LEDMode.h`. it works out of the box as every plugin that supplies LED modes no matter if static or dynamic, must include this header.

About the blackboxness of macros I totally agree with you. But I assume that most people that implement new LED modes will have a look on how the existing ones are done and just copy stuff from there. That's why I am less worried about the magic macro. It would of course be good to document it properly in those places where it is used right now.

obra:
Could we instead break the heatmap into two different plugins? One to
record key frequencies and another that displays the heat map?

Best,
Jesse
ᐧ

noseglasses:
> Could we instead break the heatmap into two different plugins? One to record key frequencies and another that displays the heat map? 

Nice idea. Like that the key recording plugin could become a service that would be available also to other plugins. I still like the idea, though, of having two heatmap plugins, a static one that is backward compatible and carries the original name and a new dynamic one that does almost the same thing but a bit more resource friendly at the price that its information is volatile. I would also consider gedankenexperimenter's idea to switch the LEDs off temporarily while the LED mode stays active.

Both versions would retrieve their key statistics from the newly introduced recording plugin. The static LED mode would simply wrap the dynamic LED mode. That would give us maximum flexibility without any code duplication.

As these ideas go far beyond the original object of this PR, I suggest that I create an issue that reminds us of these refactoring ideas for a follow-up PR. Also I would propose to turn the heatmap back to its original state for this PR, because, as gedankenexperimenter pointed out, I introduced a breaking change with respect to its behavior that I had overlooked.

Does that make sense?

obra:
Since the heatmap plugin is non-default and its data is ephemeral anyway,
I'm not too concerned about the behaviour change to make it tied to the
mode being enabled (temporarily), especially if it eases our maintenance
burden. Creating two versions of the plugin as a temporary fix feels like
it's going to add an even bigger overhead for us long-term.
ᐧ

noseglasses:
> I'm not too concerned about the behaviour change to make it tied to the mode being enabled (temporarily), especially if it eases our maintenance burden.

Ok, so I leave the changes to it in the PR.

As you just mentioned the term "temporarily". I am still not entirely happy with the terms "static" and "dynamic" that I introduced for the LED modes. Those were the first set of terms that came to my mind when I started working on this. But after a while it ocurred to me that those might be confusing to users. I'm afraid that _dynamic_ and _static_ could be misread as applying to the effects of the LED mode, like LED lights staying statically on or changing dynamically. Also static is a C++ language keyword.

Any ideas for possibly better suited and less confusing names?

obra:
Can you point me to a prose description of what the two types of modes are
or how they differ? I'm happy to chatter about naming.
ᐧ

noseglasses:
> Can you point me to a prose description of what the two types of modes are or how they differ?

I explained it briefly at the top of this PR under _Purpose of this Change_. 

The most important difference is that the old-style LED modes are global C++ objects that live forever and the new-style LED modes come with instances that have a restricted lifetime where only one new-style LED mode can be active at a time.

obra:
This reminds me that we have an as-yet-unimplemented upcoming feature to
support things like LED indicators or underglow, which might mean that you
have different LED modes running for different groups of LEDs at the same
time.

In terms of naming, the first thing that comes to mind is "short-lived LED
modes" and "long-lived LED modes"
ᐧ

noseglasses:
> This reminds me that we have an as-yet-unimplemented upcoming feature to support things like LED indicators or underglow, which might mean that you have different LED modes running for different groups of LEDs at the same time.

We could copy the whole thing and reuse great parts of the implementation by templating it. That could be a second step based on this PR's changes. 

LED indicators, underglow and by-key LEDs appear to be distinct enough so that every group needs it's own type of LED modes.

> "short-lived LED modes" and "long-lived LED modes"

That's already more precise than static/dynamic and doesn't conflict with the very different meanings of _static_ in C++.

noseglasses:
Or what about _permanent_ and _transient_ LED modes, to get rid of the hyphens?

noseglasses:
...the longer I think about it, I notice that permanenet might again be misleading :-|

gedankenexperimenter:
How about "persistent" instead of "static" or "permanent"?

gedankenexperimenter:
> ...the longer I think about it, I notice that permanenet might again be misleading :-|

It wasn't clear why you might think that at first, but now I'm guessing that it could indicate to the reader an LED mode that can't be switched off. Assuming my guess is correct, I think my "persistent" suggestion has the same problem, but to a smaller degree. But what if the "dynamic" mode wasn't called "transient", "temporal", or any "short-lived" synonym, but rather "replaceable" or a synonym?

I got the idea when I thought of using the term "deciduous", which would fit perfectly, since it indicates something that exists for a time, then is replaced. I love the idea of using that term for this, but it's obviously terrible for people who aren't into dentistry. =)

gedankenexperimenter:
One more name suggestion: "resident" _vs_ "itinerant". It's probably too obtuse, but the metaphor is very pleasing.

noseglasses:
`Co

src/kaleidoscope/plugin/Heatmap.h
gedankenexperimenter: > The plugin is able to be informed about the mode id of its LED mode. It can then check if its LED mode is the current one. The heatmap plugin checks this in its handlers…

Thinking about this further, I think I can see an alternative, and I'd like to know what you think of the idea. Instead of passing the mode ID to the factory function when creating the dynamic LED mode, the constructor for the dynamic mode could use its "parent" pointer to call an "activate" function that sets a boolean so the parent knows its dynamic mode is active. Then, when it gets swapped out, the virtual destructor function could call "deactive" in the parent mode. The parent mode stores one byte either way (unless it combines the boolean into a bitfield), the factory function uses one fewer argument, and there's a vtable entry for the dynamic mode's destructor. I don't know how that compares, or if there's another reason this idea wouldn't work.

If this would work, and it's not worse in memory consumption, maybe it would be worth considering.

noseglasses: I considered different approaches for the activity notification and went with the the least resource consuming one.

There are two problems with your proposed approach.
* Dynamic LED modes can but don't have to store a parent pointer. Some are completely independent from their parents (those that are not configurable). For those we can save two bytes.
* LED mode currently does not have a virtual destructor. If we would add it, we would add two bytes of PROGMEM for the additional entry in each LEDMode derived class' vtable. Also for those that have nothing particular to clean up.

gedankenexperimenter:
On naming the different modes:
I've been adapting this excellent technique for Kaleidoglyph, in which I am unconstrained by backwards compatibility requirements. Since I was free to just decide that all LED modes would be "dynamic", I thought about renaming everything, and so found myself considering the question of "What _is_ this thing that gets swapped into and out of memory?" The answer I came up with was that it is an LED mode updater. It handles computing the values for the LEDs in each update cycle, so I think I'm going to use the name `LedModeUpdater` for the parent class, and maybe just `SomeMode::Updater` for the inner classes that inherit from the former.

I'm less convinced that this is the best name for Kaleidoscope, where there's more need to distinguish between the two types of LED modes, but even in that case, naming the class by the function it performs might be a better choice than naming it after some property that it has.

noseglasses:
Travis failed as branches diverged. Force pushed all changes after rebasing on current master.

noseglasses:
@obra, could you please comment on the current set of name suggestions. Up to now we have

|static|dynamic|
|---|---|
|short-lived|long-lived|
|permanent|transient|
|persistent|replaceable|
||deciduous|
||conjured|
||buffer sharing|

Any other suggestions?

As soon as we agree on a set of terms, I can update this PR.

gedankenexperimenter:
You've got "short-lived" and "long-lived" swapped.

noseglasses:
> You've got "short-lived" and "long-lived" swapped.

Fixed.

obra:
I think "persistent" and "transient" is my favorite pairing of the options
listed.
ᐧ

noseglasses:
Rebased on top of current master after branches diverged.

src/kaleidoscope/plugin/LEDMode.h
```diff
@@ -0,0 +1,148 @@
...
+  }
+};
+
+} // end namespace plugin
+} // end namespace kaleidoscope
+
+// Invoke this macro in the class definition of a plugin that requires
+// access to its associated transient LED mode.
+//
+#define ACCESS_THIS_LED_MODE                                                   \
```
obra: I think we can give this a better name, but I'm not sure what that name is. 

@algernon?

noseglasses: I really did not like the magic macro `ACCESS_THIS_LED_MODE` either. Not only because of its name but because of its mere existence. I really would have liked to implement that using inheritance of an auxiliary interface class but I could not because Kaleidoscope at the point where I wrote the code for this PR lacked the ability to check if a class implements a method. 

Now, through the recent enhancement of the event handler signature checks, I sneaked in this ability. By means of an additional commit this PR, I refactored a bit and added a general traits class setup macro `DEFINE_HAS_MEMBER_TRAITS` in a header `has_method.h`. This is now used to check if a LED mode implements the method `registerLEDModeActivated`. This method is inherited from an interface class `AccessTransientLEDMode` that replaces magic macro `ACCESS_THIS_LED_MODE`.

noseglasses: BTW, `AccessTransientLEDMode` lives in header `LEDMode.h`.

noseglasses:
I have put this back to [wip] as I would like to refactor the commits in a way that they represent orthogonal changes that represent a consistent firmware state each.

noseglasses:
Phew! I am still struggling with the DCO and astyle. Sorry for the many forced pushes.

After checking all commits I was only two split this PR up into two separate commits. The first introduces the new `has_member.h` header, the rest does the transition to persistent/transient LED modes. Some of the previous commits came from pushing changes that considered reviewers' comments. Those were squashed.

I did a final check with elf_diff which shoes that the overall changes in resource consumption didn't change compared to the initial resource report that I posted at the top of this PR.

Removing [wip].

obra:
Thanks! I'd like @algernon to have a look over the final form before we
merge.
ᐧ

algernon:
Sorry for being slow with reviewing, I have scheduled enough time tomorrow for reviews, including this PR.

src/kaleidoscope/plugin/Heatmap.h
algernon: I don't think having two versions of Heatmap is worth the effort, even if that effort is small. The original idea was to collect data all the time, even when the LED mode is not the active one. However, for the plugin to display useful information, you need to have fairly long typing sessions without rebooting the keyboard. This makes the always-collect goal much less important, because many people move with their laptops, or shut down / hibernate / etc their computers often enough to lose the collected data.

So, in a sense, the Heatmap plugin is most useful *while you're looking at it* anyway, while it is active, because the data it collects otherwise is neither persistent, nor particularly much, most likely. For those who want real heatmap data, host-side collection and analytics are far superior.

For this reason, I think it's ok if the Heatmap plugin becomes transient, and only collects data while it is active.

src/kaleidoscope_internal/LEDModeManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+  //              that is optimized out by the compiler. It is there
+  //              to enable the possible future introduction of a virtual
+  //              destructor for class LEDMode.
+  //
+  //              A virtual destructor would enable LEDModes to do any types of
+  //              cleanup on destruction. But as it's a virtual method,
+  //              introducing such a destructor would create additional
+  //              entries in vtables and mean higher PROGMEM consumption.
+  //              If it will be introduced in future versions of Kaleidoscope
+  //              uncomment the commented lines below.
```
algernon: I might be missing something, but which commented lines does this comment refer to?

src/kaleidoscope/plugin/Heatmap.h
gedankenexperimenter: > For those who want real heatmap data, host-side collection and analytics are far superior.

…unless you want information about keypresses that don't get sent to the host (e.g. layer shift keys). And don't want spurious information about virtual keys (e.g. `LSHIFT(Key_X)`).

I, for one, leave my Model01 powered for many days at a time, but I neither want the LEDs on while I'm typing, nor would I want them to stay on overnight. I forget if HostPowerManagement changes the running LED mode, but if it does, that would destroy collection for someone who wants the _keyswitch_ data, rather than the _keycode_ data.

src/kaleidoscope_internal/LEDModeManager.cpp
gedankenexperimenter: It must be the call to `~LEDMode()` on line 77.

src/kaleidoscope/plugin/Heatmap.h
algernon: > …unless you want information about keypresses that don't get sent to the host (e.g. layer shift keys)

There are still options for host-side collection (KeyLogger + custom script, for example), though not ready-made.

> I forget if HostPowerManagement changes the running LED mode

It doesn't, not by default.

I too, leave the keyboard powered for many days, but my vague feeling is that we're the outliers. Nevertheless, *if* there is a need, we can add a static variant then. Until then, I think it's more practical to have it transient.

src/kaleidoscope_internal/LEDModeManager.cpp
algernon: Oh, I see. I was interpering "uncomment" as removing comments, not as commenting out. Thanks for clearing that up!

obra:
Thank you! Merged :)

src/kaleidoscope_internal/LEDModeManager.cpp
noseglasses: Sorry for the late reply, I was out of town for some weeks. The above comment 
```
  //              If it will be introduced in future versions of Kaleidoscope
  //              uncomment the commented lines below.
```
was a leftover. i had commented the destructor call in earlier versions and the left it uncommented as it is a noop in the absence of a virtual destructor that does cause no overhead and would only become active if a virtual destructor might be added.

PR #618, opened 2019-04-03
Event handler versioning, deprecation and extended signature checks

noseglasses:
Recently a proposed change to the firmware (#589) required a modification of one of
the event handlers'/hooks' signatures.

Up to now, we only supported a single overload/implementation of a
event handler of a specific type. This means that changing handler signatures
was only possible by adding a handler method with a similar name and
a different signature like

EventHandlerResult
  onKeyswitchEvent(Key &mappedKey, byte row, byte col, uint8_t keyState);

and

EventHandlerResult
  onKeyswitchEvent2(Key &mappedKey, KeyAddr key_addr, uint8_t keyState);

As this was regarded as unacceptable confusion of the users of our
programming interface, this commit introduces some changes to the
event handler management.

Important changes:

* handlers can now be overloaded,
* handler signatures can be versioned,
* individual versions can be declared as deprecated,
* we now abort the compile if a handler reimplementation with a bad signature
  is found,
* or if two or more handlers with correct signatures are detected that
  are furtheron considered as ambiguous.

Impact on resources:

The proposed changes only affect the way things are handled at compile time.
No changes in terms of PROGMEM or RAM are to be expected.

### DIFF
--- a/src/kaleidoscope/event_handlers.h
+++ b/src/kaleidoscope/event_handlers.h
@@ -35,15 +35,68 @@
 #define _ABORTABLE true
 #define _NOT_ABORTABLE false
 
+#define _NOT_DEPRECATED
+
+//******************************************************************************
+//******************************************************************************
+// Important
+//******************************************************************************
+//******************************************************************************
+// When adding new handlers or handler versions, don't forget
+// to add them both to the _FOR_EACH_EVENT_HANDLER and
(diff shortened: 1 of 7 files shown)


obra:
Thanks so much for this. I'm going to make one small change to naming (NOT_DEPRECATED to "CURRENT_IMPLEMENTATION) and merge.

obra:
Merged.

PR #620, opened 2019-04-04
Fix Qukeys mod flag rollover

### LINKED ISSUE #619: Qukeys causes incorrect rollover of modifier flags
Here we go again…

There was a [problem reported yesterday on the forum](https://community.keyboard.io/t/unexpected-characters-using-qukeys/2726) that appeared to be Qukeys adding both the primary and secondary keycodes to the report. What is actually going on is that if the `Key` being sent as the primary keycode contains a modifier flag, that modifier flag also gets applied to the key that was rolled over to. This almost certainly will also happen if three keys are involved and the middle one (not the qukey) contains a modifier flag.

Ironically, the work we did to fix the general modifier flag rollover problem actually makes this more difficult to fix. After struggling all day attempting to find a workaround, I finally came up with something, but it's a terrible hack. I can fix it, and it will work, but it just makes the worst bit of Qukeys code even worse. I'll bring up a better solution in #568, but I want a short-term fix to be available right away, so there will be a PR coming shortly.

gedankenexperimenter:
This is a short-term solution for the problem of qukeys rollover from mod-flagged keys. When flushing its queue, Qukeys stashes a copy of the current keyboard HID report, and bases the new (usually mid-cycle) report on the previous one, because it's complete, unlike the partial current report. This prevents bugs where it would cause undesired repeats of keys scanned later in the cycle, but results in a bypassing of the system in the HIDAdaptor that masks modifier flags when rolling over to a key without those flags.

By first storing a copy of the previous key report's modifiers byte and then restoring it after the report for a released qukey is sent, we remove any modifier flags it added to that report, so the next key in the queue (if any) won't base it's report on those modifiers (which would normally be masked by the HIDAdaptor).

This is a temporary fix. The real fix will probably involve changes to both KeyboardioHID and the HIDAdaptor, and will allow Qukeys to stop accessing the HID reports, which really ought to be private member variables of the Keyboard class.

Fixes #619.

### DIFF
--- a/src/kaleidoscope/plugin/Qukeys.cpp
+++ b/src/kaleidoscope/plugin/Qukeys.cpp
@@ -173,11 +173,12 @@ bool Qukeys::flushKey(bool qukey_state, uint8_t keyswitch_state) {
   // have a full HID report, and we don't want to accidentally turn
   // off keys that the scan hasn't reached yet, so we force the
   // current report to be the same as the previous one, then proceed
-  HID_KeyboardReport_Data_t hid_report;
-  // First, save the current report
-  memcpy(hid_report.allkeys, Keyboard.keyReport.allkeys, sizeof(hid_report));
+  HID_KeyboardReport_Data_t curr_hid_report;
+  // First, save the current report & previous report's modifiers
+  memcpy(&curr_hid_report, &Keyboard.keyReport, sizeof(curr_hid_report));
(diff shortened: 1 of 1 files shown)


gedankenexperimenter:
For the record, I do not like this change, but it's in a section of code that I already think needs to be redone as soon as there are changes merged that will allow it, and it merely increases the extent to which the encapsulation is already broken slightly. And it fixes the bug…

obra:
Thanks so much for the interim fix. Merging.

PR #621, opened 2019-04-04
Linear Addressing - Follow-Up

noseglasses:
This is a cleaned up and ready-to-merge version of the code formerly submitted as PR #589.
The code has been modified and subdivided in individual commits according to the reviewers' comments.


noseglasses:
There is one tiny issue that I wasn't able to solve, which is the weak function that the tap dance plugin employs to be user-configurable.

With such a weak-function based design there is no reasonable way to add deprecation warnings. This is because the deprecation mechanism requires some deprecated stuff to be called in user space (sketch or user plugin). Here, we have to call the function that is actually to be deprecated from within the tap dance plugins implementation. Thus, if we would flag the `tapDanceAction` function as `DEPRECATED(...)`, this deprecation warning would always be emitted whenever the plugin's cpp-file is compiled. This would only cause useless warnings that unnecessarily pollute the output of the smoke-examples tests.

Instead, i added comments in the tap dance plugin's header file that declare the function being deprecated. But that's about all we can do. So I guess, we will have to accept the additional`tapDanceAction` function as a permanent reminiscence.

As a conclusion, I strongly advice against using weak pseudo-callbacks like that in future implementations. We'd rather want to use proper callback registration instead. Here's an example that would enable indirectly deprecating callbacks by actually deprecating their registration function.

```cpp
class SomePlugin
{
   public:

      typedef void (*OldCallbackType)(uint8_t row, uint8_t col);
      typedef void (*NewCallbackType)(KeyAddr key_addr);

      DEPRECATED(OLD_CALLBACK) 
      void setCallback(OldCallbackType cb) { old_callback_ = cb; }

      void setCallback(NewCallbackType cb) { new_callback_ = cb; }
 
      void doSomething() {
         if(new_callback_) {
            (*new_callback_)(KeyAddr()); }
         else {
            (*old_callback_)(255, 255); }
         }
      }
      
   private:

      OldCallbackType old_callback_ = nullptr; // soon to be removed...
      NewCallbackType new_callback_ = nullptr;
};
```
Such an i

There are some places in Kaleidoscope where a proper callback pointer is already used. The layer class e.g. supports setting a `getKey` function pointer. Unfortunately not through an accessor function. Such accessors come for free as they are easy to be optimized out by the compiler.

src/kaleidoscope/hardware/kbdfans/KBD4x.h
```diff
@@ -45,6 +45,11 @@ class KBD4x: public kaleidoscope::hardware::ATMegaKeyboard {
 
   static constexpr int8_t led_count = 0;
 
+  typedef MatrixAddr<matrix_rows, matrix_columns> KeyAddr;
+  static constexpr int8_t numKeys() {
+    return matrix_columns * matrix_rows;
+  }
+
```
obra: Can this get lifted up to ATMegaKeyboard? Part of the point of having the abstraction there is to reduce the boilerplate code for each implementation.

src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -25,7 +25,7 @@
 // Maximum length of the pending queue
 #define QUKEYS_QUEUE_MAX 8
 // Total number of keys on the keyboard (assuming full grid)
-#define TOTAL_KEYS ROWS * COLS
+#define TOTAL_KEYS KeyboardHardware.numKeys()
```
obra: It feels like in this case, we should just be replacing this macro

src/kaleidoscope/plugin/LEDControl.cpp
```diff
@@ -112,21 +112,24 @@ void LEDControl::set_all_leds_to(uint8_t r, uint8_t g, uint8_t b) {
 }
 
 void LEDControl::set_all_leds_to(cRGB color) {
-  for (int8_t i = 0; i < LED_COUNT; i++) {
-    setCrgbAt(i, color);
+  for (int8_t led_index = 0; led_index < LED_COUNT; led_index++) {
+    setCrgbAt(led_index, color);
```
obra: Please don't mix variable renamings in with implementation changes.

src/kaleidoscope/plugin/LED-Wavepool.cpp
```diff
@@ -179,32 +178,29 @@ void WavepoolEffect::update(void) {
...
-      uint8_t offset = (r * COLS) + c;
-      int8_t height = oldpg[pgm_read_byte(rc2pos + offset)];
+  for (auto key_addr : allKeyAddr()) {
+    int8_t height = oldpg[pgm_read_byte(rc2pos + key_addr.toInt())];
 #ifdef INTERPOLATE
-      if (now & 1) {  // odd frames only
-        // average height with other frame
-        height = ((int16_t)height + newpg[pgm_read_byte(rc2pos + offset)]) >> 1;
-      }
+    if (now & 1) {  // odd frames only
```
obra: Relatively minor, but the whitespace changes shouldn't be mixed in here.

src/kaleidoscope/Kaleidoscope.h
```diff
@@ -42,6 +42,10 @@ extern HARDWARE_IMPLEMENTATION KeyboardHardware;
 
 typedef HARDWARE_IMPLEMENTATION::KeyAddr KeyAddr;
 
+inline constexpr KeyAddr allKeyAddr() {
```
obra: Having seen how we're using this in later commits, I think a name like allKeys might end up more readable and understandable than allKeyAddr

gedankenexperimenter:
I'm trying this out, and finding that `ShiftToLayer(FUNCTION)` keys (palm keys on default sketch) cause the keyboard to become non-functional. Can anyone else reproduce this problem?

src/kaleidoscope/layers.cpp
```diff
@@ -84,45 +84,43 @@ void Layer_::handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState) {
...
+        Key mappedKey = (*getKey)(layer, key_addr);
+
+        if (mappedKey != Key_Transparent) {
+          active_layers_[key_addr.toInt()] = layer;
+          break;
         }
-        layer--;
       }
     }
+    layer--;
```
gedankenexperimenter: This decrement needs to be inside the `while` loop, or it never terminates, and the keyboard hangs whenever a layer key is pressed.

gedankenexperimenter:
Als, rollover from a key with a modifier flag to one without is broken in this branch somehow; the modifier flags from the held key are not masked when the next key is pressed. For example, rollover from `Key_LeftCurlyBracket` to `Key_Q` (easy to reproduce on the default firmware by holding <kbd>fn</kbd>) produces `{Q`, but it should be `{q`.

src/kaleidoscope/Kaleidoscope.h
gedankenexperimenter: I think `allKeyAddrs()` (or even `eachKeyAddr()`) would be better than `allKeys()`. When looking at this:
```c++
for (auto x : allKeys()) { ... }
```
…I would be expecting the `x` was of type `Key`, not `KeyAddr`.

More important, this is the best place for a comment describing how the iterator works, and how it's used. Someone skimming through the code and looking at this could be easily baffled, and someone who is confused by any of the range-based for loops that use it would first look here, not in `MatrixAddr.h` — that's two more searches away, and even then it's buried deep in the file without any good clues directing the reader to the relevant comment.

gedankenexperimenter: …and I know I've said this before, but why use `auto` in those for loops and keep the type unclear, when you could remove all uncertainty and just use the actual type?

obra: I'd say that if we have KeyAddr in the thing we're iterating over, that's "Type spelled out nearby"

obra: allKeyAddrs() feels ok to me in a way that allKeyAddr() doesn't.

src/kaleidoscope/hardware/kbdfans/KBD4x.h
noseglasses: Not yet. Currently the base class `ATMegaKeyboard` is unaware of the dimensions of the derived keyboards' key matrices. 

But I think that's something, @algernon is currently working on.

src/kaleidoscope/plugin/Qukeys.h
noseglasses: Alright, I will do so.

src/kaleidoscope/plugin/LEDControl.cpp
noseglasses: That slipped my final check. Will be reverted.

src/kaleidoscope/plugin/LED-Wavepool.cpp
noseglasses: Ok.

src/kaleidoscope/Kaleidoscope.h
noseglasses: Ok, let's go for `allKeyAddrs()`.

@gedankenexperimenter, there's a very informative article about the [AAA Style (Almost Always Auto)](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/), written by Herb Sutter.

src/kaleidoscope/layers.cpp
noseglasses: Thanks.

src/kaleidoscope/plugin/LED-Wavepool.cpp
noseglasses: Just noticed that this is done by astyle because we got rid of one for loop, the indentation is automatically adapted. wontfix.

src/kaleidoscope/Kaleidoscope.h
gedankenexperimenter: @noseglasses — I'll read it, but I'm very skeptical. Thanks for sharing the argument, in any case.

@obra — To some people, myself included, unless the actual type name `KeyAddr` appears somewhere nearby, that definitely is not "type spelled out nearby". If it were a class method (e.g. `KeyAddr::all()` or `KeyAddr::each()` or `KeyAddr::allKeyAddrs()`), then I would _almost_ agree with you, because `KeyAddr` at least appears as a type name in that case, rather than as part of a function name string. I would certainly guess correctly, but I would still find it impossible to just assume that it would do what I expect, and I with the code as it is, I would have a long odyssey to get to the comment in MatrixAddr.h where it is explained.

noseglasses:
i finally went with a mixture of several solution that came up during the review. The range based for loop now reads 
```cpp 
for(auto key_addr : KeyAddr::all()) { ... }
```
This is both readable and from an implementation standpoint a better solution as it does not require an additional global function like `allKeyAddr()` that would need to be implemented for every template instance of `MatrixAddr` separately. Also by using a nested `Range` class in `MatrixAddr` (which is an implementation detail) we are now on the orthodox side when it comes to containers and iterators. This will help those people that are interested in studying the firmware code as a whole.

obra:
> This is both readable and from an implementation standpoint a better solution as it does not require an additional global function like `allKeyAddr()` that would need to be implemented for every template instance of `MatrixAddr` separately. Also by using a nested `Range` class in `MatrixAddr` (which is an implementation detail) we are now on the orthodox side when it comes to containers and iterators. This will help those people that are interested in studying the firmware code as a whole.

I like it. Thank you!

src/kaleidoscope/Kaleidoscope.h
obra: You've argued your case. I'm happy with what Florian's got here. I'm resolving this conversation.

src/kaleidoscope/hardware/kbdfans/KBD4x.h
obra: Is that the changes in the bootloader PR? If so, I'd love to get that merged in and this updated for that.

algernon: Kiiiiinda. #611 makes the derived keyboard accessible to the base class, so yeah, it will be - or can be made - aware of the keyboard dimensions.

obra:
@algernon  - #611 has enough in it that it may make sense to split it into
smaller PRs
ᐧ

noseglasses:
Rebased and force pushed to consider changes in master.

noseglasses:
Resolved conflicts with current master. Rebased an force-pushed.

obra:
Well, it took far too long, but this is merged down. Thanks so much for all the hard work @noseglasses

PR #625, opened 2019-04-11
Fixed Kaleidoscope Builder for Arduino paths containing whitespaces

noseglasses:
This PR adds what is needed to build Kaleidoscope with msys2 when Arduino is installed in its Windows standard location that contains spaces "C:\Program Files (x86)\Arduino".

it basically adds missing quotes to variable evaluations.


obra:
Thanks! Merged.

PR #632, opened 2019-04-26
Smoother mouse acceleration function

gedankenexperimenter:
The cubic acceleration function used to approximate a sine wave produced some noticeable jumps when beginning to accelerate the mouse cursor movement (several 1s, followed by several 4s, then 7s, with no intermediate values). This parabolic function produces smoother mouse cursor motion, without any sudden jumps in speed.

Curiously, the size of the binary is exactly the same.

### DIFF
--- a/src/kaleidoscope/plugin/MouseKeys/MouseWrapper.cpp
+++ b/src/kaleidoscope/plugin/MouseKeys/MouseWrapper.cpp
@@ -114,22 +114,17 @@ void MouseWrapper_::warp(uint8_t warp_cmd) {
   warp_jump(section_left, section_top, next_height, next_width);
 }
 
-// cubic wave function based on code from FastLED
-// produces a shape similar to a sine curve from 0 to 255
-// (slow growth at 0, fast growth in the middle, slow growth at 255)
-// http://www.wolframalpha.com/input/?i=((3((x)**2)%2F256)+-+((2((x)(x)(x%2F256))%2F256)))+%2B+1
+// To approximate a sine wave, this uses two parabolas. Acceleration begins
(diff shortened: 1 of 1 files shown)


obra:
Thanks for looking into this. I definitely wasn't happy with where we'd
gotten the acceleration curve to.
ᐧ

gedankenexperimenter:
There's still a tiny discontinuity right in the middle, where the value `127` gets repeated (because we need to start at `1`, not `0`, and end at `255`, not `256` (which becomes `0`). I'm not certain, but I think I might have been able to detect it once or twice, but only when I'm looking intently for it.

gedankenexperimenter:
I just tried out a modified version of this that I like a bit better, but which isn't as clear:
```c++
uint8_t MouseWrapper_::acceleration(uint8_t cycles) {
  cycles += 12;
  if (cycles < 128) {
    uint16_t c2 = cycles * cycles;
    return (c2 >> 7);
  } else {
    uint16_t remaining_cycles = 256 - cycles;
    uint16_t c2 = remaining_cycles * remaining_cycles;
    return 256 - (c2 >> 7);
  }
}
```
This removes the discontinuity in the middle of the acceleration curve, and it starts the acceleration sooner by shifting the whole curve to the left by adding twelve at the start. The compiled size is still the same, but I don't like having that magic number in there (the first twelve numbers would otherwise be zeros).

It also has the downside of returning zero for any input higher than 232. This is guarded against in the current calling functions, but it's definitely more fragile.

I think it's worth trying out, though, because it feels more responsive than the other function.

obra:
@gedankenexperimenter Can you open an issue with the new algorithm? I'm merging the current state of the PR in for now, as it's an improvement on what we've got now. Thanks!

PR #635, opened 2019-04-27
Make opposing mouse movement keys cancel each other

### LINKED ISSUE #634: Mouse movement left & up trumps right & down
When using mouse movement keys, if I hold down both `Key_mouseL` and `Key_mouseR` simultaneously, the mouse cursor moves to the left, rather than holding in place, as I would expect. This happens because of [these lines in MouseKeys.cpp](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys.cpp#L94-102). If the left-movement bit is set, the right-movement bit never gets checked.

gedankenexperimenter:
Instead of having `Key_mouseL` & `Key_mouseUp` override `Key_mouseR` & `Key_mouseDn`, respectively, stop mouse movement on a given axis when both keys are held simultaneously. Doing so will not reset acceleration, so when one of them is released, the mouse will resume movement at full speed immediately.

Closes #634.

### DIFF
--- a/src/kaleidoscope/plugin/MouseKeys.cpp
+++ b/src/kaleidoscope/plugin/MouseKeys.cpp
@@ -92,14 +92,14 @@ EventHandlerResult MouseKeys_::beforeReportingState() {
   }
 
   if (mouseMoveIntent & KEY_MOUSE_UP)
-    moveY = -speed;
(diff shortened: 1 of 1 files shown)


obra:
Thanks. Applied.

PR #637, opened 2019-05-01
Remove Qukeys dependence on KeyboardioHID

gedankenexperimenter:
This is an update to Qukeys that removes its (direct) dependence on KeyboardioHID by sending only end-of-cycle reports.

Currently, Qukeys sends reports as soon as the qukey state is determined, including in the middle of a scan cycle. To prevent removal of keycodes that are still being held, it bases these mid-cycle reports on the previous, complete report rather than the current, incomplete one. In order to do so, it was copying and modifying the reports stored in KeyboardioHID directly, but there were two problems with this approach. First, those reports ought to be private data members, and second, this only worked for NKRO reports. See #568 for more information.

I had planned to address this with a different fix (#570), but then it became clear that even this wouldn't solve all the rollover problems because the HID report doesn't store all of the information needed about the state of the keyboard (See #619 & #620 for details).

This fix is a major rewrite of Qukeys that should fix both the encapsulation and rollover issues. It is also problematic, but only because of the code complexity required. Instead of flushing the queue immediately, all reports are sent at the end of the cycle, and at most one key's report(s) will be sent each cycle.

This PR depends on #621, because I don't want to rewrite it again after that one gets merged.


gedankenexperimenter:
This PR is now obsolete (superseded by #640).

PR #642, opened 2019-05-07
Standardize plugin timers

gedankenexperimenter:
This PR is built on #641, and makes use of it to provide consistency and reliability of timers in the core Kaleidoscope plugins. It also substantially reduces PROGMEM usage, in part by reducing the number of bytes needed to store timestamps where possible.

Each plugin's changes are in a separate commit, mainly to make debugging easier. I have not yet thoroughly tested them all to make sure that I didn't make any errors.

### Plugin testing status
- [x] LEDControl
- [x] LED-AlphaSquare
- [x] LED-Stalker
- [x] LED-Wavepool
- [x] LED-BootAnimation
- [x] LED-BootGreeting
- [x] LEDEffect-Breathe
- [x] LEDEffect-Chase
- [x] LEDEffect-Rainbow
- [x] IdleLEDs
- [x] CycleTimeReport
- [x] GhostInTheFirmware
- [x] Heatmap
- [x] Leader
- [x] MagicCombo
- [x] MouseKeys
- [x] OneShot
- [x] SpaceCadet
- [x] TapDance
- [x] TypingBreaks

### DIFF
--- a/src/kaleidoscope/plugin/SpaceCadet.h
+++ b/src/kaleidoscope/plugin/SpaceCadet.h
@@ -52,7 +52,7 @@ class SpaceCadet : public kaleidoscope::Plugin {
     //The flag (set to 0)
     bool flagged = false;
     //the start time for this key press
-    uint32_t start_time = 0;
+    uint16_t start_time = 0;
   };
 
   SpaceCadet(void);
--- a/src/kaleidoscope/plugin/CycleTimeReport.cpp
+++ b/src/kaleidoscope/plugin/CycleTimeReport.cpp
@@ -20,12 +20,12 @@
 
 namespace kaleidoscope {
 namespace plugin {
-uint32_t CycleTimeReport::next_report_time_;
+uint16_t CycleTimeReport::last_report_time_;
 uint32_t CycleTimeReport::loop_start_time_;
 uint32_t CycleTimeReport::average_loop_time;
 
 EventHandlerResult CycleTimeReport::onSetup() {
-  next_report_time_ = millis() + 1000;
+  last_report_time_ = Kaleidoscope.millisAtCycleStart();
   return EventHandlerResult::OK;
 }
 
@@ -42,11 +42,11 @@ EventHandlerResult CycleTimeReport::afterEachCycle() {
   else
     average_loop_time = loop_time;
 
-  if (millis() >= next_report_time_) {
+  if (Kaleidoscope.hasTimeExpired(last_report_time_, uint16_t(1000))) {
     cycleTimeReport();
 
     average_loop_time = 0;
-    next_report_time_ = millis() + 1000;
+    last_report_time_ = Kaleidoscope.millisAtCycleStart();
   }
 
   return EventHandlerResult::OK;
--- a/src/kaleidoscope/plugin/CycleTimeReport.h
+++ b/src/kaleidoscope/plugin/CycleTimeReport.h
@@ -32,7 +32,7 @@ class CycleTimeReport : public kaleidoscope::Plugin {
   static uint32_t average_loop_time;
 
  private:
-  static uint32_t next_report_time_;
+  static uint16_t last_report_time_;
   static uint32_t loop_start_time_;
 };
 }
--- a/src/kaleidoscope/plugin/GhostInTheFirmware.cpp
+++ b/src/kaleidoscope/plugin/GhostInTheFirmware.cpp
@@ -46,11 +46,11 @@ EventHandlerResult GhostInTheFirmware::beforeReportingState() {
       return EventHandlerResult::OK;
     }
     is_pressed_ = true;
-    start_time_ = millis();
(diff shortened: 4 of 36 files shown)


src/kaleidoscope/plugin/SpaceCadet.h
```diff
@@ -52,7 +52,7 @@ class SpaceCadet : public kaleidoscope::Plugin {
     //The flag (set to 0)
     bool flagged = false;
     //the start time for this key press
-    uint32_t start_time = 0;
+    uint16_t start_time = 0;
```
gedankenexperimenter: We definitely don't need this timeout to be longer than one minute.

src/kaleidoscope/Kaleidoscope.h
```diff
@@ -132,6 +132,28 @@ class Kaleidoscope_ {
...
+   * - start_time: A timestamp when the timer started, which should be set by
+   *      calling `Kaleidoscope.millisAtCycleStart()`. It can be any integer
+   *      type.
+   *
+   * - ttl: The timeout value or interval to check (ttl = "time to live"). The
+   *      timer expires (and `hasTimeExpired()` returns true) when the time that
+   *      has elapsed since `start_time` exceeds this value. It must be and
+   *      integer type that is no bigger than the type of `start_time`.
+   */
+  template <typename _Timestamp = uint16_t, typename _Timeout = uint16_t>
```
noseglasses: That's about the first time that I see default values for a template function.

Default values for template parameters make sense when defining template classes. The only place where I can see them to be useful with template functions is when taking a pointer to the respective function. There you could write 
```cpp
auto fncPtr = &hasTimeExpired<>;
```
Is that intented? Or what is the reason for the default parameters here?

noseglasses:
Great PR! Standardizing such things makes it easier and less confusing for newbies to find the correct way to do things with timing by taking a look at the core code.

src/kaleidoscope/Kaleidoscope.h
gedankenexperimenter: You give me more credit than I deserve in this case; my inexperience with C++ templates is showing. Now that you've pointed it out, it does seem pointless to use default values, though I should add something more suggesting that it's best to use the smallest size that will work (and that, because integer overflow is handled properly, it will often work to use relatively small sizes).

gedankenexperimenter:
The most important plugins that I haven't tested yet are Leader, MagicCombo, and TapDance. I don't use any of them, so I don't feel comfortable declaring them to be working properly.

src/kaleidoscope/Kaleidoscope.h
gedankenexperimenter: See #641 for an update.

gedankenexperimenter:
I just checked LED-Chase, which appears to be broken; it starts out very fast, then slows down to normal speed. I think it probably resumes normal speed when it overflows the 16-bit integer. It shouldn't be hard to fix.

gedankenexperimenter:
I forgot to mention that I intentionally left Qukeys out of this PR because it already does the correct computation, and because #640 is a total rewrite of Qukeys that would result in merge conflicts with this one that I don't want to bother dealing with.

obra:
@gedankenexperimenter At this point, are the plugins you've updated all considered safe to merge?

gedankenexperimenter:
I still haven't tested them all; only the ones with checkmarks. I can try testing the rest, but I'm not very familiar with those ones, so I'm not certain I'll be able to tell if there's a problem. I did take care when making the code changes to get it right, but I can't be totally confident.

I'll update this PR to non-draft status and rebase it (I think there's at least one other PR that will result in a merge conflict). If I find the time, I'll do some more testing, too.

obra:
awesome. if you felt like splitting out the tested ones into a separate PR, I’d be happy to merge that asap

gedankenexperimenter:
> if you felt like splitting out the tested ones into a separate PR, I’d be happy to merge that asap

My initial thought was, "That sounds like it might be a lot of work". Then I realized that I already split the plugin changes into separate commits. Thanks, past me!

gedankenexperimenter:
I decided to go ahead and test the plugins I hadn't bothered to look at yet. Everything appears to work correctly, insofar as I detected no difference in output between this and the master branch.

obra:
In general, this looks good to me. @algernon, @noseglasses: any concerns?

noseglasses:
Looks good to me and is definitely an improvement, both in terms of performance and readability. Although I would like to see a `Time` utility class instead of cluttering `Kaleidoscope` with more and more unrelated methods and features.

While I was looking at `Kaleidoscope::hasTimeExpired(...)` i noticed a nasty little error that could cause strange behavior under certain conditions, fixed in https://github.com/keyboardio/Kaleidoscope/pull/659.

gedankenexperimenter:
Please explain what you think the "nasty little error" is. #659 is definitely wrong, by the way; those temporary time variables should not be `auto`; they are very intentionally the same size as the input timestamp.

PR #646, opened 2019-05-17
Added weak versions of keymaps_linear

noseglasses:
This is necessary if a plugin wants to define an alternative
layer system/keymap system/key lookup system which goes without using the
KEYMAP(...) macro.
Before this change, not using the KEYMAP(...) macro in the sketch
resulted in a linker error due to a missing symbol.


noseglasses:
Sure.

PR #656, opened 2019-06-07
Fix typo in Consumer_AC_FullScreenView

tpope:
I left the old definition in place for the extremely unlikely case that someone is relying on it. Let me know if another strategy is preferred.


noseglasses:
Nice catch!

What do you thing about defining it less reduntantly like
```cpp
#define Consumer_AC_FullScreenView CONSUMER_KEY(HID_CONSUMER_AC_FULL_SCREEN_VIEW, KEY_FLAGS |  HID_TYPE_SEL)
#define Consumer_AC_FullSCreenView Consumer_AC_FullScreenView
```

tpope:
Fine by me, just force pushed a change.

PR #659, opened 2019-06-14
Comments and `Kaleidoscope::hasTimeExpired(...)` static

noseglasses:
This PR does two things (in separate commits).

* turns `Kaleidoscope::hasTimeExpired(...)` static
* adds additional comments about potential pitfalls with too large values of `ttl`

### DIFF
--- a/src/kaleidoscope/Kaleidoscope.h
+++ b/src/kaleidoscope/Kaleidoscope.h
@@ -154,6 +154,17 @@ class Kaleidoscope_ {
    *      timer expires (and `hasTimeExpired()` returns true) when the time that
    *      has elapsed since `start_time` exceeds this value. It must be an
    *      integer type that is no bigger than the type of `start_time`.
+   *
+   * Warning: When using hasTimeExpired, make sure that the value of ttl
+   *          is at least by the size of the call interval smaller than
+   *          the maximum value that the datatype of timeout can store.
+   *          Otherwise false negatives are likely to occur
+   *          when checking (elapsed_time > ttl).
+   *
+   * Example: When both arguments of hasTimeExpired are of type uint16_t and
+   *          the function is called in every cycle with a cycle duration
+   *          of x ms, the value of ttl must not be larger than
+   *          std::numeric_limits<uint16_t>::max() - x.
    */
   template <typename _Timestamp, typename _Timeout>
   bool hasTimeExpired(_Timestamp start_time, _Timeout ttl) {
--- a/src/kaleidoscope/Kaleidoscope.h
+++ b/src/kaleidoscope/Kaleidoscope.h
@@ -167,7 +167,7 @@ class Kaleidoscope_ {
    *          std::numeric_limits<uint16_t>::max() - x.
    */
   template <typename _Timestamp, typename _Timeout>
-  bool hasTimeExpired(_Timestamp start_time, _Timeout ttl) {
+  static bool hasTimeExpired(_Timestamp start_time, _Timeout ttl) {
     _Timestamp current_time = millis_at_cycle_start_;
     _Timestamp elapsed_time = current_time - start_time;
     return (elapsed_time > ttl);


gedankenexperimenter:
Also, why should this method be `static`? You say this fixes something, but what's the problem?

noseglasses:
The narrowing would cause a truncation after the keyboard ran for around one minute if `_Timestamp` is `uint16_t` and after a quarter of a second if `uint8_t`. There is a good reason why it is supposed to be `uint32_t`, no?

The temporary `current_time` that is left after my change can safely be removed.

A non-static method requires the objects `this` pointer being pushed on the stack or being passed in a register (compiler and architecture dependent). That's an extra-overhead that is only justified if a method accesses non-static class inventory, which is not the case here.

gedankenexperimenter:
The truncation is intentional, because not all timers need to be concerned about times greater than 65 seconds (or, in a few cases, 256ms). Overflow is handled by first calculating `elapsed_time`.

With this change, after about a minute, you'll get problems for timers that use less than 4 bytes for efficiency, because now you'll be subtracting a value that has already been truncated from one that has not.

gedankenexperimenter:
If it's more efficient to make this function `static`, by all means go ahead with that change, but making those temporary variables `auto` is both less efficient and incorrect.

noseglasses:
The following program demonstrates what can go wrong. It's meant to be run on x86 but the behavior will be the same on 8 bit avr.

https://gist.github.com/noseglasses/9d72017cb6c9f1338061712cf91bdbaa

When dealing with templates, stay away from explicit types and asumptions about what users will pass as template parameters. A template function must work for any types if not protected by `static_asserts` and type traits.

noseglasses:
i have the feeling that it would already be safer to at least use the same template type for `start_time` and `ttl`. Then the user cannot accidentally use too large values for timeout and the original implementation with truncation could work. Not sure about signed values, though. Mind you, this is a template, so users are free to pass signed values, too.

it would be the safes to have a dedicated timer class that wraps everything and has no corner cases.

noseglasses:
Sorry, I accidentally closed this RP.
I pushed a version that does what i proposed in my last post.

gedankenexperimenter:
Yes, it would be safer, but now you would be prohibiting timeout values that are limited to 8 bits from being used with timestamps of 16 bits.

The one general problem with this whole timeout-checking algorithm is that it doesn't necessarily run every millisecond. If it only gets checked every X milliseconds, and the timeout value is within X milliseconds of its maximum value, it could fail to time out when it should. There are at least a few places where a timeout of up to 255ms is allowed, but higher values don't make sense. In order to guarantee no false negatives, we actually need to use a larger integer type for the timestamp variable than that used for the ttl.

gedankenexperimenter:
Put another way, if the timeout variable is one byte, and might be set to its maximum value, we need a two-byte timestamp to guarantee safety because a cycle can take more than one millisecond.

noseglasses:
What about the following (instead of using static_asserts and traits within the template to inform the user what we accept and what not).
```cpp
...
public:
   bool hasTimeExpired(uint16_t start_time, uint8_t timeout) {
      return hasTimeExpired_(start_time, timeout);
   } 

   bool hasTimeExpired(uint32_t start_time, uint16_t timeout) {
      return hasTimeExpired_(start_time, timeout);
   }
   
private:
  template <typename _Timestamp, typename _Timeout>
  bool hasTimeExpired_(_Timestamp start_time, _Timeout ttl) {
    _Timestamp current_time = millis_at_cycle_start_;
    _Timestamp elapsed_time = current_time - start_time;
    return (elapsed_time > ttl);
  }
...
```

What's your opinion about that?

noseglasses:
`CycleTimeReport` doesn't like this fix. @gedankenexperimenter, any ideas?

gedankenexperimenter:
It's pretty common for plugins to have timeout values of a few seconds. As a plugin author, who wants to be frugal with the meager resources on an 8-bit MCU, I don't want to use a 4-byte integer for timestamps when I know that two bytes is plenty for any reasonable use-case.

Debugging gets tricky, too, because I don't know which version gets called if both arguments are type `uint16_t`.

I really think it's best to leave this one up to the plugin developers to handle any guarantees about timeout value ranges and timestamp types. I'd rather leave it as it is than try to take away the plugin developer's responsibility and options.

If this wasn't meant for such a memory-impoverished device, I wouldn't take this position, but this is a compromise that I think is appropriate.

noseglasses:
> I really think it's best to leave this one up to the plugin developers to handle any guarantees about timeout value ranges and timestamp types. I'd rather leave it as it is than try to take away the plugin developer's responsibility and options.

Fair enough. I amended the comment of `hasTimeExpired` in a way that warns for potential pitfalls with too large values of `ttl`.

PR #668, opened 2019-07-20
Color names and tweak to SolidColors to support cRGB

jdlien:
colors.h provides a list of named cRGB definitions that (mostly) match those in CSS, with a few minor tweaks to account for the hardware of the Model 01. Additionally, colors.h contains a simple function, dim() to allow changing the brightness of a color without having to recalculate the values.

dimLimit() is used by dim() to make colors as dim as possible without shutting off color components completely, as tends to happen on the Model 01 at low brightness levels. This prevents the colors from changing too dramatically at low brightness values.

I have made an additional constructor for LEDEffect-SolidColors so that it can also be instantiated with cRGB values, which allows it to become a simple demo for the use of colors.h

Note that I have put color names in their own namespace - kaleidoscope::colors. (Not in plugins -- I'm not sure if this is desired or not, but this makes sense to me).


obra:
Thanks so much for working on this. Would you mind splitting the color stuff and the keygroup stuff into separate PRs? I expect that the keygroup work is going to require more workshopping to be mergeable. (In particular, I'm thinking about how it will interact with non-US keyboard layouts.

jdlien:
Hi @obra, As per your request I have removed keygroups.h from this PR. Is that okay, or did you want a completely new PR? I will create a separate one for the keygroups addition.

obra:
Splitting apart is great. No need to blow away the old PR. You did good :)

jdlien:
Thanks, I look forward to this being integrated, hopefully, although I understand if you want to scrutinize it or make some adjustments to make it fit into the firmware better.

I'll note that one change I made to the original was that I had a reserved color set called "nocolor" in FunctionalColor that I used to specify that no color was going to be assigned to a key since all I could really assign to it was a color and needed some way to say "don't assign any color". That way another broader grouping of keys wouldn't be overridden by it.

I don't know if something like this might be useful to have defined more generally. The color I chose was CRGB(1, 1, 2) which is completely black/off, so it's a useless color on the Model01, and even on my iMac Pro I can't find any way to distinguish this from black at full brightness.. I guess it seemed like an unlikely color for someone to choose, moreso even than the darkest gray 1,1,1.

Nonetheless, I have removed this from this set of colors as it probably doesn't make sense for most applications.

gedankenexperimenter:
I think it's worth noting that the color values in Kaleidoscope are hardware-dependent and substantially non-linear. On the Model01 (and anything else that uses KeyboardioScanner) any `CRGB` component value up to 27 gets changed to a 0 before it reaches the LED itself, so a value of `CRGB(1, 1, 2)` is actually the same as `CRGB(0, 0, 0)` to the hardware. The rest of the curve makes sense, because of how the physical LED responds, but I have long wondered if it might be better not to have so many zero values, so a 1 would be "minimum brightness" rather than "off".

jdlien:
Oh, I wasn’t aware that this was an intentional design decision. Knowing of the effect, I created the dim functions with dimLimit() to try to avoid shutting off the LEDs at low brightness levels.

obra:
We use a pretty standard LED Gamma table.  It's intentional but hardly
unique

obra:
looking more at this specific chunk of code, it feels like it might be better cast as a kaleidoscope plugin that provides the named colors, rather than something baked straight into the core.

jdlien:
Probably sensible. Any suggestions on naming?

obra:
Kaleidoscope-Plugin-CSSColors?

jdlien:
Okay, works for me. I can update later today after work.

algernon:
I'm loving the color names, and would like to merge this. I'm happy to turn the PR into a plugin, too, if you want me to go ahead with it.

gedankenexperimenter:
Perhaps the two-word color names should have underscores, *à la* the coding style guide?

jdlien:
Porque no los dos?

I could alias them so you could use whichever you prefer. I do like the idea of being able to use the ones everyone already knows from CSS, but maybe I’m biased being a web developer.

algernon:
`deep_sky_blue` is what I'd prefer too (after hyphens and Kaleidoscope written in Lisp, but lets stay realistic for now :))

jdlien:
Out of idle curiousity, was this or anything like this set of color names ever added to Kaleidoscope?

Renaming with underscores or underscore alternatives seemed to be the consensus for the names, but as @gedankenexperimenter pointed out, a bigger issue is that the exact rendering of the colors is quite hardware-dependent and I'm quite sure this would differ between the Model 01, Model 100 and so forth.

PR #669, opened 2019-07-20
Add Keygroups functions

jdlien:
I have added a series of functions in keygroups.h that are useful for determining which of several logical groupings a particular key might be in. This was integral to the way FunctionalColors works, and I think it might be useful to other plugin developers. I tried to be quite exhausive in coming up with these groups to include as many keys as I could find defined in Kaleidoscope, but perhaps there are other international keys I missed or something.

The functions I have come up with are the following:
keygroups. I have pored over the key definitions in Kaleidoscope, and models of various keyboards to be comprehensive in coming up with groups that make sense. (It turns out that there are a lot of media keys, but I only added the fairly common ones.)

Keygroups is integral to the way FunctionalColors works, as my intended use was to color keys of various types/groups differently, but not typically each individual key.

The functions I have come up with are the following
- isAlpha
- isNumber
- isPunctuation
- isFunction
- isNavigation
- isArrow
- isKeypad
- isSystem
- isMedia
- isLang
- isModifier
- isShift
- isControl
- isGui
- isAlt
- isMouseMove
- isMouseWheel
- isMouseButton
- isMouseWarp
- isMouse
- isMouseMove

Note that I have not created a separate namespace for these functions, so that may be a desirable change depending on how it fits with the current firmware design.


jdlien:
I'm not sure why I'm getting Travis build errors - probably some kind of scope or namespace issue.
```
Kaleidoscope/src/kaleidoscope/key_events.h:75:42: error: 'KeyAddr' has not been declared"
```
and so on.

obra:
Have you given thought to how this interacts with non-US-ANSI key layouts? You're going to run into situations where a key that's an alphanumeric key in a US-ANSI layout is punctuation in a non-US layout and vice-versa.

jdlien:
Yes, I have considered this somewhat, although I have limited exposure to such layouts. For example I am taking the Brazilian forward slash/question mark key and the language keys used in Japanese keyboards. Mostly these come from references to such keys in Kaleidoscope itself. I tried to be thorough.

I’d love more input from international users who use very different layouts.

jdlien:
Do you have specific examples of where a key gets remapped to a different kind of key? (Or can you point to a reference for these cases?)

obra:
I desperately wish I had a good reference for this -- then we could
generate proper mapping tables for all the different local/national
keyboard layouts.

Dvorak is one. (Semicolon becomes s, Q becomes ', etc)
French AZERTY is another (M becomes ,)

jdlien:
Well... I'm using the key names assigned by Keyboardio for this. If you define a Dvorak layout in firmware using the appropriate key names, this should work, correct? But if you define a normal QWERTY layout and let the OS remap everything, then that's when semicolon gets changed to S, etc, right?

In that case it seems reasonable to rely on the key names... because how would you know if the OS has remapped a bunch of scan codes? I guess you'd have to make a different version of keygroups for each layout (or possibly pass a parameter that will adjust it appropriately). That sounds very hard to create and maintain for every possible remapping scheme.

obra:
It is indeed a very, very complicated problem. And that's part of why I'm a
little bit hesitant about this as a core feature. There have been a couple
efforts to build lists of key definitions for non-English languages, but
they work by defining a layer on top of the USB HID definitions.

To be clear, your end goal is laudable and something I want. I just want to
be careful not to box ourselves into a corner with something that's going
to break for non-English on-device layouts.

PR #672, opened 2019-08-10
inserted missing ```sh

Plotnus:
Cleaned up the readme a bit. Was missing some opening ```
I also think there was a dropped `#` for the clone the hardware definitions header.


Plotnus:
This is not code so can we waive the DCO stuff on this one?

obra:
yup. done. thanks!

PR #684, opened 2019-09-13
Prevent rollover to a qukey from causing unintended repeats

### LINKED ISSUE #682: Erroneous key repetition with Qukeys
I'm having a problem with Qukeys where the last key _before_ I press a qukey is repeated anywhere from 2-8 times (roughly, here's an uncorrected example just frommmm writingg this issue). The specific situation is that I'm trying out a small 36 key layout, and I have a number/symbol layer that I access via a qukey on the space key. I've tried setting the overlap threshold to many values from 0 to 100, and it doesn't seem to make a big difference. Is there any way to mitigate this? Anything I can test? Something I just don't understand about qukeys?

gedankenexperimenter:
When rollover occurs from a non-modifier key to a qukey, if we delay the release event of that key until after the qukey's state is resolved, and if the hold timeout is set to a fairly large value (on the order of 500ms), unintended repeats would occur for a key that was actually only tapped. To prevent this, if there's only one event in the queue (the press of the qukey), and we see a release of a non-modifier key that's not the qukey, it's okay to allow that release event to skip the queue and simply proceed as if it had been released before the qukey was pressed.

This might fix #682, but I doubt it. @neh, please give it a try and see if it helps.

### DIFF
--- a/src/kaleidoscope/plugin/Qukeys.cpp
+++ b/src/kaleidoscope/plugin/Qukeys.cpp
@@ -51,6 +51,14 @@ EventHandlerResult Qukeys::onKeyswitchEvent(Key& key, KeyAddr k, uint8_t key_sta
 
   // Deal with keyswitch state changes.
   if (keyToggledOn(key_state) || keyToggledOff(key_state)) {
+    // If the user rolled over from a non-modifier key to a qukey, let the
+    // release event for that key skip the queue. This prevents unintended
+    // repeat characters for the tapped key, which would otherwise have its
+    // release event delayed.
+    if (keyToggledOff(key_state) && event_queue_.length() == 1 &&
+        k != event_queue_.addr(0) && !isModifierKey(key)) {
+      return EventHandlerResult::OK;
+    }
     // If we can't trivially ignore the event, just add it to the queue.
     event_queue_.append(k, key_state);
     // In order to prevent overflowing the queue, process it now.


neh:
I'm testing this out with both a high hold timeout (500) and overlap threshold (95) and it is much improved, but I still get occasional repetitions. Will put more testing time into it and update soon.

neh:
I've tested this out more, and I think it reduces the problem a bit, but doesn't eliminate it.

src/kaleidoscope/plugin/Qukeys.cpp
```diff
@@ -50,6 +50,14 @@ EventHandlerResult Qukeys::onKeyswitchEvent(Key& key, KeyAddr k, uint8_t key_sta
 
   // Deal with keyswitch state changes.
   if (keyToggledOn(key_state) || keyToggledOff(key_state)) {
+    // If the user rolled over from a non-modifier key to a qukey, let the
+    // release event for that key skip the queue. This prevents unintended
+    // repeat characters for the tapped key, which would otherwise have its
+    // release event delayed.
+    if (keyToggledOff(key_state) && event_queue_.length() == 1 &&
+        k != event_queue_.addr(0) && isModifierKey(key)) {
```
gedankenexperimenter, line 58: Past me was sloppy. The last condition here is wrong, and should be `! isModifierKey(key)`…

gedankenexperimenter:
So, I screwed up the fix for #682 here, and only just noticed it a few days ago. I've improved my fix here, and I think it will actually fix the problem, or at least one of the problems that was being observed by a few people.

obra:
Absent regression tests, I'd love to get a positive report on this one before merging.

gedankenexperimenter:
Me, too. That's why I've left it as a draft for now. I was never able to reproduce part of #682 (the repetition while typing), so while I can test the hold issue myself, I'm still in some doubt about this being the right solution.

gedankenexperimenter:
I have tested this out, and I can confirm that it works as expected now. Without this change, when rolling over from a non-modifier key to a qukey, if you hold the qukey down, and the hold timeout is long enough, you'll end up with the previous key's release event delayed until the qukey times out, even though that original key was only tapped briefly. And if the qukey press rolls over to the next key, and the overlap threshold is less than 100%, the delay will last even past the release of the qukey. With this change, the tapped key's release won't be delayed, and it will appear to the host that there was no rollover. The events will get re-ordered, but only slightly, and not in a harmful way.

I _think_ this should fix the problem @neh reported in #682. It definitely fixes one of the sub-problems, and I think it should be merged. I would really like someone who could reproduce the issue to test it out, though, to see if the new, non-botched version of this PR really fixes the whole problem.

neh:
I'm not using the Model 01 any more, but I have an Atreus on the way... would testing it there be helpful?

gedankenexperimenter:
> I'm not using the Model 01 any more, but I have an Atreus on the way... would testing it there be helpful?

Sure, especially if you can reproduce the problem without this change!

neh:
I've only had time for a quick test so far, but have been unable to reproduce the original problem on the Atreus. I should have some time this weekend to take another run at it.

obra:
So sorry this hung out for so insanely long. Merged.

PR #689, opened 2019-09-20
Force explicit type conversion for MatrixAddr

noseglasses:
**Caution: This PR if merged will cause build errors for all keyboards that are based on AtmegaKeyboard. The reason for this is that the design of the inheritance of keyboard hardware from AtmegaKeyboard is fundamentally broken in a sense that the class `AtmegaKeyboard` defines methods (e.g. `AtmegaKeyboard::isKeyMasked`) with an `KeyAddr` argument. This type is inherited from `kaleidoscope::Hardware` and evaluates to `MatrixAddr<0, 0>`. The actually correct template instantiation of `MatrixAddr<ROWS, COLS>` is only known to the derived hardware classes, e.g. `Atreus`, etc.**

**The changes in this PR reveal this problem. To fix it, a redesign of the hardware stuff is probably inevitable. After those problems have been fixed, this PR can safely be merged.**

---

By forcing an explicit type conversion between
two template class instances of template
MatrixAddr<...>, we prevent undesired implicit
construction of the wrong MatrixAddr types.

Before this change, the following would have been possible

```cpp
typedef MatrixAddr<5, 5> KeyAddr;

void f(KeyAddr k) {} // uses MatrixAddr<5, 5>

void g() {
   typedef MatrixAddr<0, 0> KeyAddr; // Stupid but possible
   f(KeyAddr(1, 12)); // Would instantiate MatrixAddr<0, 0> and
                      // implicitly convert it to MatrixAddr<5, 5>
}
```

With this commit, the compiler will emit an error and explicit type
conversion is required.

```cpp
typedef MatrixAddr<5, 5> KeyAddr1;
typedef MatrixAddr<2, 2> KeyAddr2;

void f(KeyAddr1 k) {} // uses MatrixAddr<5, 5>

void g() {
   f(KeyAddr1(KeyAddr2(1, 1)); // Now an explicit type conversion is
                               // required.
}
```

This commit also introduces a compile time check that prevents
conversion from a matrix type with greater extension to one with
smaller extension.

### DIFF
--- a/src/kaleidoscope/MatrixAddr.h
+++ b/src/kaleidoscope/MatrixAddr.h
@@ -66,8 +66,15 @@ class MatrixAddr {
   ThisType &operator=(ThisType &&) = default;
 
   template<typename MatrixAddr__>
+  explicit
   constexpr MatrixAddr(const MatrixAddr__ &other)
-    : MatrixAddr(other.row(), other.col()) {}
+    : MatrixAddr(other.row(), other.col()) 
+  {
+     static_assert(MatrixAddr__::rows <= ThisType::rows,
+        "Matrix type conversion failed. Source type must not have greater row size than target type");
+     static_assert(MatrixAddr__::cols <= ThisType::cols,
+        "Matrix type conversion failed. Source type must not have greater col size than target type");
+  }
 
   constexpr uint8_t row() const {
     return offset_ / cols;


obra:
It could be argued that at the time ATMegaKeyboard was created, it wasn't misdesigned, as the change to using KeyAddrs was a refactor that arrived later.

Do you have thoughts about how to redesign the hardware plugins to make this a non-issue?

noseglasses:
Entirely correct. It is broken in its current state. Although `ATMegaKeyboard` was actually an `ATMegaKeyboardWithSpecificMatrixLayout` from the very start, due to its use of `ROWS` and `COLS`. So it was already a bit of a hack and not, as the name suggests, a general base class.

I proposed an approach to an hardware class redesign a while ago: #665

noseglasses:
As a temporary fix, one could move the definition of `KeyAddr` to global scope and define it at the top any specized hardware header at a point before anything else is included. That way, `ATMegaKeyboard` would see the right `KeyAddr` type.

noseglasses:
I submitted a PR that is a potential (temporal) fix for the problems discussed above: https://github.com/keyboardio/Kaleidoscope/pull/690

The submitted PR is definitely not a nice solution but one that does not require to touch any derived hardware headers.

The problems introduced by the transition to `KeyAddr` single byte addressing could have easily been detected by simulator based integration testing that would have tested all relevant features of all `ATMegaKeyboard` derived keyboards. `ATMegaKeyboard::actOnMatrixScan` becoming dysfunctional would not have gone unnoticed.

I would still be happy to spend more work on the simulator and testing stuff. But currently further work depends on providing the hardware classes with the ability to define and make their properties available to simulator code and thus on making something like #665 available.

obra:
Thanks. I hear you about the simulator. And I'm really looking forward to
being able to have it in place. @algernon has had an implementation of #665
on his plate for a bit now and says that he expects to have it pushed up
for review today or tomorrow.
ᐧ

PR #691, opened 2019-09-24
Arduino 1.8.10 broke the build

obra:
arduino-builder changed in 1.8.10, which caused our build to break. these two changes work around the failures. Building with 1.8.9 still works


obra:
https://github.com/arduino/arduino-builder/issues/336 is tracking the ctags issue. It does sound like Arduino wants us to stop using arduino-builder going forward. joy.

PR #720, opened 2019-11-12
Key union converted to a proper class

noseglasses:
Important: This PR introduces a breaking change. All code that accesses the `Key::raw` member directory will break.

Travis fails because there are two more PRs that need to be merged along with this to
allow the entire firmware to build.

https://github.com/keyboardio/Kaleidoscope-HIDAdaptor-KeyboardioHID/pull/21
https://github.com/keyboardio/Model01-Firmware/pull/92

Unions are a C-reminiscense that are better avoided in modern C++.
They cause specific problems due to their nature of representing
independent types. The way they are used in Kaleidoscope, they
can easily be replaced by a class.

This enables it to properly work with Key objects in constexpr context
where with the old union-based implementation the compiler reported
errors when one Key was constructed based on a key_code/flags pair and
another one through raw-data. In such a case, the compiler assumes that
both Key instances represent something entirely different. This is
because unions were never meant for type conversions and the C++
standard considers their use for that purpose as undefined behavior.

The new class provides accessor methods for raw-data access and for
key_code/flags-data access.

This is a breaking change as it is is not possible to replace direct
member access patterns like

key.raw = 0xFFFF;

based on the raw-accessors.

All direct access via .raw, .keyCode and .flags have been replaced
throughout Kaleidoscope.

### DIFF
--- a/src/kaleidoscope_internal/deprecations.h
+++ b/src/kaleidoscope_internal/deprecations.h
@@ -51,5 +51,23 @@
   "`Kaleidoscope.device().led_count` instead."
 
 #define _DEPRECATED_MESSAGE_HARDWARE_BASE_CLASS                 \
-  "The `Hardware` base class is deprecated. Please use\n" __NL_ \
+  "The `Hardware` base class is deprecated. Please use\n" __NL__ \
   "the new APIs based on `kaleidoscope::device::Base`."
+
+#define _DEPRECATED_MESSAGE_DIRECT_KEY_MEMBER_ACCESS                           \
+  "Direct access to `Key` class' data members is deprecated.\n"                \
+  "Please use `Key::setKeyCode()`/`Key::getKeyCode()` or\n"                    \
+  "`Key::setFlags()`/`Key::getFlags()` instead.\n"                             \
+  "\n"                                                                         \
+  "For further information and examples on how to do that, \n"                 \
+  "please see UPGRADING.md."
+
+#define _DEPRECATED_MESSAGE_KEY_MEMBER_RAW_ACCESS                              \
+  "The member variable `raw` of class Key had to be removed. Please \n"        \
+  "use `Key::setRaw()`/`Key::getRaw()` to set and get raw data.\n"             \
+  "\n"                                                                         \
+  "Important: This is not a deprecation. Your code will compile but fail\n"    \
+  "           to link until all access to `Key::raw` has been replaced.\n"     \
+  "\n"                                                                         \
+  "For further information and examples on how to do that, \n"                 \
+  "please see UPGRADING.md."
--- a/NEWS.md
+++ b/NEWS.md
@@ -156,6 +156,22 @@ The [FirmwareDump](doc/plugin/FirmwareDump.md) plugin makes it possible to dump 
 
 ## Breaking changes
 
+### Implementation of type Key internally changed from C++ union to class
+
+Type `Key` was originally implemented as a C++ union. For technical reasons
+it had to be converted to a C++ class. This implies that the double usage
+of the original union, holding either raw data (member `raw`) or key code/key flags
+data (members `keyCode` and `flags`) is no more possible. 
+
+Direct use of member `raw` will 
+emit a diagnostic compiler message but will cause the firmware linking 
+process to fail. For a deprecation
+periode `keyCode` and `flags` keep on being supported but will cause 
+deprecation warnings during compile. 
+
+Please see the [relevant upgrade notes](UPGRADING.md##implementation-of-type-key-internally-changed-from-union-to-class)
+for information about how to upgrade legacy code.
+
 ### The `NumPad` plugin no longer toggles `NumLock`
 
 The `NumPad` plugin used to toggle `NumLock` when switching to the NumPad layer. This caused issues on OSX where `NumLock` is interpreted as `Clear`. For this reason, the plugin no longer does this. As a consequence, everyone's encouraged to update their keymap so that the numpad layer uses normal number keys instead of the keypad numbers. See [Model01-Firmware#79](https://github.com/keyboardio/Model01-Firmware/pull/79) for an example about how to do this.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -12,6 +12,7 @@ If any of this does not make sense to you, or you have trouble updating your .in
     - [Bidirectional communication for plugins](#bidirectional-communication-for-plugins)
     - [Consistent timing](#consistent-timing)
   + [Breaking changes](#breaking-changes)
+    - [Implementation of type Key internally changed from C++ union to class](#implementation-of-type-key-internally-changed-from-union-to-class)
     - [The `RxCy` macros and peeking into the keyswitch state](#the-rxcy-macros-and-peeking-into-the-keyswitch-state)
     - [HostOS](#hostos)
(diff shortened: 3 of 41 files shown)


obra:
This being a breaking change that touches user sketches without any deprecation period, warning, or back compat makes me fairly nervous.

Is it -possible- to turn the previous union into a class and still use the undefined-but-supported-by-gcc behavior inside to expose the old  raw,  flags, and keyCode properties for some period of time?

noseglasses:
> This being a breaking change that touches user sketches without any deprecation period, warning, or back compat makes me fairly nervous.

I definitely expected that. Very understandable.

> Is it -possible- to turn the previous union into a class and still use the undefined-but-supported-by-gcc behavior inside to expose the old raw, flags, and keyCode properties for some period of time?
 
This is something that gave me headaches for a long time. Still I haven't found a good solution.

This is not about giving the `Key` data structure accessors or the actual conversion. It is about the compiler throwing errors as soon as two constexpr  `Key` instances are compared, one having been initialized through `raw` data and the other one through `keyCode` and `flags`. For the compiler these are entirely independent data structures. This happens only at compile time and that's why it did not cause too much trouble up to now. Only when I started working on host_keymaps that heavily rely on constexpr, it turned out that it is impossible with the union-based `Key`. So technically it's not a undefined-but-supported-by-gcc behavior, depending on what you do with `Key`.

I am agonizing over a solution that gives us both, the backward compatible behavior and still the ability to do fancy compile time things, yet without success.

noseglasses:
There's definitely no way to replace a union by a class without breaking anything.

But at a closer look on the use cases of class `Key`, it turns out that there is a much higher probability for the `.keyCode` and `.flags` members being used in user code than the `.raw` member.

This suggests that we could make the effects of the union->class transition for `Key` less drastically by keeping `.keyCode` and `.flags` and only sacrificing `.raw`.

Code that uses the `.raw` member directly will still break.

i force-pushed a change to this PR that replaces `Key`s data members `keyCode` and `flags` with proxy objects that allow for emitting deprecation warnings only in those places where `.keyCode` and `.flags` are used from outside of the `Key` class. This solution is less noisy than just deprecating the class data members directly.

As in contrast to my first approach this solution requires storing `uint8_t` values for key codes and flags separately instead of packing it into one `uint16_t`, this means that when reading a `Key` from PROGMEM, both values have to be read separately. In all places where `Key`s used to loaded from PROGMEM, a convenient member function `Key Key::readFromProgmem()` wraps PROGMEM access nicely.

@obra, I am sorry that I cannot reduce your level of nervousity to 0%. But maybe the proposed solution is acceptable, nonetheless. The perspective of obtaining the ability to do all sorts of fancy things with `Key` at compile time (which the compiler disallows with unions) IMHO is definitely worth it.

noseglasses:
Factored out an unrelated change and moved it over to #721. Force pushed changes.

noseglasses:
Heureka! Last night I had an epiphany. 

There **is** a way to support `Key::keyCode`, `Key::flags` **and** `Key::raw` event when moving away from a union to a class.

The trick is to make `Key::raw` a static member and emit a deprecation message that informs the user about all access to `Key::raw` having to be replaced by `Key::getRaw()` and `Key::setRaw()`.

To be on the safe side, no instance for `Key::raw` will be provided and the code will fail to link until all access to `Key::raw` has been properly replaced.

Access to `Key::keyCode` and `Key::raw` will trigger a normal deprecation warning.

Now we have a solution where users are properly warned and we still finally have a class `Key`.

obra:
That sounds acceptable to me.
ᐧ

obra:
(I'm reviewing the code now)

obra:
I'd love to get @algernon's take on this before merge.

algernon:
I'll go over this first thing tomorrow. Already started, but won't be able to finish before I fall asleep.

algernon:
Right... I promised a review, but today didn't quite go as originally planned, so all I can offer is a first impression. I definitely like the direction this is taking, but it _is_ a breaking change, even with the `flags` and `keyCode` deprecations. Though, at least the breakage is descriptive, but it is still one. Like @obra, I can live with that. I feel like we're nearing a point where we should consider a flag day, but that's beyond this PR.

What I _love_ about this PR, is `Key::readFromProgmem()`.  `DataProxy` is also very educational to read. It's also very nice to be able to use a constructor for Keys.

All in all, my first impression is to stamp my approval. We'll need to update `NEWS.md` and `UPGRADING.md` too, but I can do that step (but if I don't have to, that's even better!).

src/kaleidoscope_internal/deprecations.h
```diff
@@ -51,5 +51,17 @@
...
 
 #define _DEPRECATED_MESSAGE_HARDWARE_BASE_CLASS                 \
-  "The `Hardware` base class is deprecated. Please use\n" __NL_ \
+  "The `Hardware` base class is deprecated. Please use\n" __NL__ \
   "the new APIs based on `kaleidoscope::device::Base`."
+
+#define _DEPRECATED_MESSAGE_DIRECT_KEY_MEMBER_ACCESS                           \
+  "Direct access to `Key` class' data members is deprecated.\n"                \
+  "Please use `Key::setKeyCode()`/`Key::getKeyCode()` or\n"                    \
+  "`Key::setFlags()`/`Key::getFlags()` instead."
```
algernon: I think this should be reworded a bit. It's clear to anyone who knows C++, but for someone who's just copy & pasting code, this might as well be Klingon. I think a better way would be to point them to `UPGRADING.md`, where we can provide a few examples of how to get from the old style to the new one.

To be honest, just slap `For further information and examples on how to do that, please see UPGRADING.md`.

src/kaleidoscope_internal/deprecations.h
```diff
@@ -51,5 +51,17 @@
...
+
+#define _DEPRECATED_MESSAGE_KEY_MEMBER_RAW_ACCESS                              \
+  "The member variable `raw` of class Key had to be removed. Please \n"        \
+  "use `Key::setRaw()`/`Key::getRaw()` to set and get raw data.\n"             \
+  "\n"                                                                         \
+  "Important: This is not a deprecation. Your code will compile but fail\n"    \
+  "           to link until all access to `Key::raw` has been replaced."
```
algernon: Same as above, slapping the same text at the end of this should do the trick, and point people to the right direction even better.

src/kaleidoscope_internal/deprecations.h
noseglasses: Modified `deprecations.h`, now pointing to `UPGRADING.md`.

src/kaleidoscope_internal/deprecations.h
noseglasses: Modified `deprecations.h`, now pointing to `UPGRADING.md`.

src/kaleidoscope_internal/deprecations.h
noseglasses: Now, as even Duolingo supports Klingon, one can expect more and more Klingon speaking users around :smirk:

noseglasses:
Thanks @algernon for the review and the positive comments. I totally forgot about `UPGRADING.md` and `NEWS.md`. Updated both and fixed the deprecation warnings accordingly.

obra:
I'm +1 to merge once @algernon signs off.

Issue #726, opened 2019-11-21
Device API: What belongs where (props/implementation)


noseglasses:
This is not an actual issue but rather a question that IMHO could do with some discussion and clarification.

The new device API introduced `...Props` classes that describe device and drivers. AFAIK, it is currently not exactly defined what parts of information belong in `...Props` classes and what in the corresponding implementation classes.

An example that demonstrates why I bring up this topic is the method 
```cpp
uint8_t getLedIndex(uint8_t key_offset)
``` 
that is currently part of the LED driver implementation. During my work on the virtual device it turned out that the simulator needs to have access to this method as it has to map a key offset to the matching LED index. This is e.g. useful when during future simulation based integration testing the actual color value of a LED is meant to be compared with a set value. But the physical devices driver classes are not build (excluded by `#ifndef KALEIDOSCOPE_VIRTUAL_BUILD` clauses). That's why the virtual driver classes cannot ask their physical siblings but have to ask their `...Props` classes instead.

My solution is to move `getLedIndex` to the LED drivers properties class.

I understand the Props classes to contain anything that is necessary to describe the implementation. Thus, they may contain any type of stateless, hardware agnostic information in terms of constants and (not necessarily `constexpr`) static methods. This perspective considers mappings through functions as descriptive information, too.

In other words, only implementation details that relate to actual hardware and that cannot be build for the virtual architecture, as well as anything that is stateful **must** be part of the implementation classes.

@obra and @algernon, what's your opinion about this? 

In any case, we should document the answer to this question somewhere in the developer relevant parts of the docs (unless its not already there and I just missed it).

algernon:
I think things like `getLEDIndex()` should indeed be moved to `Props`, because they're just lookups, really. Ideally, the led index table would be a (private) member of the `Props` struct too. In a sense, mapping is just a way to access a property, the LED index corresponding to a key. As such, Props is a good place for it.

I do feel we may need better wording, but in principle, I agree with you.

(You didn't miss it, we do not currently have decent docs for the new API. That's something I plan to fix in the near future, but didn't get there yet.)

obra:
Hrm. Part of the point of naming it "Props" was that it was my
understanding that it was "just" a data structure and wouldn't have code in
it.
ᐧ

noseglasses:
> Hrm. Part of the point of naming it "Props" was that it was my
understanding that it was "just" a data structure and wouldn't have code in
it.

[...]

PR #729, opened 2019-11-23
Key offset to LED index map made LED driver property

noseglasses:
This is an implementation of https://github.com/keyboardio/Kaleidoscope/issues/726#issuecomment-557780011

LED driver properties now can re-define an array
for their individual mapping from key offsets to LED indices.

This array is both constexpr (can be used at compiletime) and
stored in PROGMEM. The latter is used by the LED driver base
class to map key offsets to LED ids at runtime.

I gave this a WIP, because I do not have my Model01 around for testing. Will do that ASAP. Submitted this to ask for feedback.

### DIFF
--- a/src/kaleidoscope/driver/led/Base.h
+++ b/src/kaleidoscope/driver/led/Base.h
@@ -25,8 +25,14 @@ namespace kaleidoscope {
 namespace driver {
 namespace led {
 
+constexpr uint8_t no_led = uint8_t(-1);
+
 struct BaseProps {
   static constexpr uint8_t led_count = 0;
+
+  // C++ does not allow empty constexpr arrays
+  //
+  static constexpr uint8_t key_led_map[] PROGMEM = { no_led };
 };
 
 template <typename _LEDDriverProps>
@@ -43,8 +49,20 @@ class Base {
     };
     return c;
   }
-  uint8_t getLedIndex(uint8_t key_offset) {
-    return 0;
+  static uint8_t getLedIndex(uint8_t key_offset) {
+
+    // Give the compiler the oportunity to optimize
+    // for boards without LEDs.
+    //
+    if (_LEDDriverProps::led_count == 0) {
+      return no_led;
+    }
+
+    if (key_offset >= sizeof(_LEDDriverProps::key_led_map)) {
+      return no_led;
+    }
+
+    return pgm_read_byte(&_LEDDriverProps::key_led_map[key_offset]);
   }
 
   static class LEDs {
--- a/src/kaleidoscope/device/keyboardio/Imago.cpp
+++ b/src/kaleidoscope/device/keyboardio/Imago.cpp
@@ -31,6 +31,8 @@ namespace kaleidoscope {
 namespace device {
 namespace keyboardio {
 
+constexpr uint8_t ImagoLEDDriverProps::key_led_map[] PROGMEM;
+
 static constexpr uint8_t CMD_SET_REGISTER = 0xFD;
 static constexpr uint8_t CMD_WRITE_ENABLE = 0xFE;
 static constexpr uint8_t WRITE_ENABLE_ONCE = 0b11000101;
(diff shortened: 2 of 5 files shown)


noseglasses:
@obra, please confirm whether it is ok to convert the 2D key index mapping array of the Imago to a 1D version?

src/kaleidoscope/driver/led/Base.h
```diff
@@ -25,8 +25,14 @@ namespace kaleidoscope {
 namespace driver {
 namespace led {
 
+constexpr uint8_t no_led = 254;
```
algernon: At the moment, the Imago uses 254, while the Raise uses 255. We should figure out if this needs to be device specific, or can be generic, and if the latter, what value should it have? It's (currently) not checked by anything as far as I see. We should probably fix that too. (Adding checks is a different PR, just mentioning here)

I'd also suggest a different name, to make it clear what this is used for: `no_led_at_key`. Though, that might be a bit on the too long side of things.

obra:
I think so. The goal is that the imago implementation eventually gets lifted out and standardized. I’m planning to use that architecture in future devices, but nothing about it is set in stone yet.

src/kaleidoscope/driver/led/Base.h
```diff
@@ -43,8 +49,25 @@ class Base {
...
     return c;
   }
-  uint8_t getLedIndex(uint8_t key_offset) {
-    return 0;
+  static uint8_t getLedIndex(uint8_t key_offset) {
+
+    static_assert(sizeof(_LEDDriverProps::key_led_map) >=
+                  _LEDDriverProps::led_count,
+                  "The number of LEDs in the LED map must be greater or "
+                  "equal to the number of LEDs");
```
algernon: I do not think this assertion is valid. For example, the Dygma Raise has _less_ in the map than the total number of LEDs, because it uses some of the LEDs for underglow.

src/kaleidoscope/driver/led/Base.h
obra: Is it "safe" to declare it as -1 globally, as a shorthand for "last value in the unsigned datatype" or will the compiler be smart and stop us from doing that?

src/kaleidoscope/driver/led/Base.h
noseglasses: Good point. Will fix that.

src/kaleidoscope/driver/led/Base.h
noseglasses: `uint8_t(-1)` should work. That's a common idiom for what you intend.

noseglasses: @obra, what's your opinion about the name of the no_led constant?

src/kaleidoscope/driver/led/Base.h
noseglasses: Fixed and pushed an update.

src/kaleidoscope/driver/led/Base.h
noseglasses: Pushed an update that implements the -1 solution.

obra:
> On Nov 23, 2019, at 2:49 PM, noseglasses <notifications@github.com> wrote:
> 
> ﻿
> @noseglasses commented on this pull request.
> 
> In src/kaleidoscope/driver/led/Base.h:
> 
> > @@ -25,8 +25,14 @@ namespace kaleidoscope {
>  namespace driver {
>  namespace led {
>  
> +constexpr uint8_t no_led = 254;
> @obra, what's your opinion about the name of the no_led constant?
> 
no strong opinion

obra:
@algernon What's your take on this one? It seems fairly straightforward.

algernon:
I'm still not 100% sold on the `no_led` name, but otherwise it has my stamp of approval. I do not feel strongly about the name, either, so if @obra's fine with it, then lets merge this!

noseglasses:
Please wait with merging this until it's tested on hardware (that's why its still WIP). Will be tomorrow unless someone else is in the mood for testing.

noseglasses:
Tested on Model01. Works as expected.
Rebased on top of master and force pushed an update.
Ready to merge.

algernon:
I was looking into rebasing the Raise plugin (#722) on top of this, when I ran into something I forgot: the Raise is an unconventional device in that there's  two variants of it: one with an ANSI, and another with an ISO layout, which differ in the right-hand side only, and both are powered by the very same firmware. When the sides are "connected", the firmware checks if the right half if ANSI or ISO, and will act accordingly from that point onward.

This also affects the key<->LED mapping. The Raises a 2d array, with the first dimension being ANSI/ISO. As such, it will also require a custom `getLEDIndex()`.

Note that we can't just move `getLEDIndex()` to props, because the ANSI/ISO setting is something that may change during run-time, when one plugs in a different right side. In other words, Raise's `getLEDIndex()` is not stateless.

algernon:
Note: the Raise thing should _not_ be a blocker for getting this merged. Rather, it's something to figure out afterwards. This PR is a nice improvement, and should be merged regardless of how we deal with the Raise's peculiarities.

algernon:
Actually, scratch that. My caffeine levels were low, and I was looking at another map (the led map, not the key->led map). The key->led map for the Raise is common between ISO and ANSI, and as such, `getLedIndex()` is stateless, and we can use the generic one.

Sorry for the noise!

PR #762, opened 2019-12-09
Using include-what-you-use to reduce the amount of unnecessary includes

noseglasses:
As proposed [here](https://github.com/keyboardio/Kaleidoscope/pull/737#issuecomment-562363991), this is a follow up to https://github.com/keyboardio/Kaleidoscope/pull/737.


src/kaleidoscope/plugin/Leader.h
```diff
@@ -17,15 +17,19 @@
...
+#include "kaleidoscope/plugin.h"                // for Plugin
+#include "Kaleidoscope-Ranges.h"
 
 #define LEADER_MAX_SEQUENCE_LENGTH 4
 
 #define LEAD(n) Key(kaleidoscope::ranges::LEAD_FIRST + n)
 
 #define LEADER_SEQ(...) { __VA_ARGS__, Key_NoKey }
-#define LEADER_DICT(...) { __VA_ARGS__, {{Key_NoKey}, NULL} }
+#define LEADER_DICT(...) { __VA_ARGS__, {{Key_NoKey}, nullptr} }
```
obra: I'd rather code changes don't get mixed in with what got docced as just iwyu cleaning up our includes

src/kaleidoscope/plugin/NumPad.h
```diff
@@ -16,7 +16,12 @@
 
 #pragma once
 
-#include "Kaleidoscope-LEDControl.h"
+#include <stdint.h>                             // for uint8_t
+#include "kaleidoscope/KeyAddr.h"               // for KeyAddr
+#include "kaleidoscope/event_handler_result.h"  // for EventHandlerResult
+#include "kaleidoscope/plugin.h"                // for Plugin
+
+struct cRGB;
```
obra: Code change rather than header change here.

src/kaleidoscope/plugin/WinKeyToggle.h
```diff
@@ -17,7 +17,14 @@
...
 
-#include "kaleidoscope/Runtime.h"
+#include <stdint.h>                             // for uint8_t
+#include "kaleidoscope/KeyAddr.h"               // for KeyAddr
+#include "kaleidoscope/event_handler_result.h"  // for EventHandlerResult
+#include "kaleidoscope/plugin.h"                // for Plugin
+
+namespace kaleidoscope {
+class Key;
+}
```
obra: Do we really not offer a header which can provide this forward declaration? (I presume that's what's going on)

noseglasses: Forward declarations are a way to avoid includes. That's why iwyu suggests them.

src/kaleidoscope/plugin/NumPad.h
noseglasses: Where?

src/kaleidoscope/plugin/Leader.h
noseglasses: That replacement avoids a header include as `NULL` is a macro.

src/kaleidoscope/plugin/NumPad.h
noseglasses: Ah, you probably mean the forward declaration? 
The distinction between code change and header change makes no sense IMHO. Adding and removing headers changes the code.

src/kaleidoscope/plugin/Leader.h
noseglasses: I can surely put that into a separate commit.

noseglasses:
Fixed issues with Travis/shellcheck.

noseglasses:
Rebased due to conflicts.

Strangely, github lists travis tests as pending but travis reports them as successfully terminated.

This PR is prone to get in conflict state again soon as it is touching so many files' include sections.

PR #769, opened 2019-12-12
Fixed builds on msys2

noseglasses:
This adds some quotes to various paths used in Kaleidoscope's build system.
This fixes builds on msys2 that failed due to whitespaces in system paths.


noseglasses:
I forgot to mention that this is not actually a fix but enables the build with msys2 on some Arduino versions. On Arduino 1.8.10 builds will still fail due to other reasons: https://github.com/keyboardio/Kaleidoscope/issues/770

PR #774, opened 2019-12-13
WIP: Fixed build warnings with gcc 7 in Layer_::getKeyFromPROGMEM(...)

noseglasses:
This fixes https://github.com/keyboardio/Kaleidoscope/issues/773

The method has been moved to the sketch footer to allow the
compiler to see the correct array dimensions of keymap_linear.

### DIFF
--- a/bin/kaleidoscope-builder
+++ b/bin/kaleidoscope-builder
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU General Public License along with
 # this program. If not, see <http://www.gnu.org/licenses/>.
 
-set -e
+#set -e
 
 ######
 ###### Build and output configuration
--- a/bin/kaleidoscope-builder
+++ b/bin/kaleidoscope-builder
@@ -312,6 +312,10 @@ compile () {
         ${ARDUINO_VERBOSE} \
 	      ${ARDUINO_AVR_GCC_PREFIX_PARAM} \
 	      "${SKETCH_DIR}/${SKETCH}.ino"
+	      
+    preprocessed_sketch="${BUILD_PATH}/sketch/${SKETCH}.ino.cpp"
+    echo "Dumping preprocessed sketch ${preprocessed_sketch}"
+    cat "${preprocessed_sketch}"
 
     cp "${BUILD_PATH}/${SKETCH}.ino.hex" "${HEX_FILE_PATH}"
     cp "${BUILD_PATH}/${SKETCH}.ino.elf" "${ELF_FILE_PATH}"
--- a/src/kaleidoscope/layers.cpp
+++ b/src/kaleidoscope/layers.cpp
@@ -103,10 +103,6 @@ Key Layer_::eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState) {
   return Key_NoKey;
 }
 
-Key Layer_::getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr) {
-  return keyFromKeymap(layer, key_addr);
-}
-
 void Layer_::updateLiveCompositeKeymap(KeyAddr key_addr) {
   int8_t layer = active_layers_[key_addr.toInt()];
   live_composite_keymap_[key_addr.toInt()] = (*getKey)(layer, key_addr);
--- a/src/kaleidoscope/layers.h
+++ b/src/kaleidoscope/layers.h
@@ -119,6 +119,9 @@ class Layer_ {
   typedef Key(*GetKeyFunction)(uint8_t layer, KeyAddr key_addr);
   static GetKeyFunction getKey;
 
+  // Due to an issue with gcc 7, getKeyFromPROGMEM(...) is currently implemented
+  // in kaleidoscope_internal/sketch_preprocessing/sketch_footer.h.
+  //
   static Key getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr);
   DEPRECATED(ROW_COL_FUNC) static Key getKeyFromPROGMEM(uint8_t layer, byte row, byte col) {
     return getKeyFromPROGMEM(layer, KeyAddr(row, col));
--- a/src/kaleidoscope_internal/sketch_preprocessing/sketch_footer.h
+++ b/src/kaleidoscope_internal/sketch_preprocessing/sketch_footer.h
@@ -1,1 +1,17 @@
 // Any code that appears here is added to the bottom of the preprocessed sketch file.
+
+#include "kaleidoscope/layers.h"
+
+namespace kaleidoscope {
+
+// gcc 7 (Arduino > 1.8.10) has a problem with the layer dimension of
+// keymap_linear, as used by keyFromKeymap(...) being undefined.
+// By moving the implemenation here,
+// we can be sure that the array dimensions are known and no spurious
+// warnings will result.
+//
+Key Layer_::getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr) {
+  return keyFromKeymap(layer, key_addr);
+}
+
+} // namespace kaleidoscope


noseglasses:
It's worth being mentioned that supressing the array-bound warnings through pragmas does not work in this case. It's a known deficiency of gcc that some warnings can only be suppressed by command line `-W...` flags.

noseglasses:
Travis fails because the bundle has not been updated after merging the header/footer PR.

algernon:
I have updated the bundle last night. Travis is failing at the linking stage.

noseglasses:
I see. My assumption was wrong anyway.
The proplem may be due to travis caching ccache. That may cause problems with the new sketch header/footer stuff.
I had tested the PR with the latest bundle and Arduino 1.8.10 and it definitely worked.

algernon:
Yeah, I was suspicious about the cache too, and have emptied it and restarted the build, still fails to link. Works fine locally with 1.8.10, will check with 1.8.8 and 1.8.9 later too.

obra:
I cleared all the travis caches, pulled the bundle up to current and... it still explodes on build :/

Does it work for you with Arduino 1.8.8?

noseglasses:
I only checked with Arduino 1.8.10. I am currently traveling and therefore cannot check. Might take some days till I can give it a try.

algernon:
Works for me with 1.8.8, 1.8.9, and 1.8.10 too.

noseglasses:
Are we really sure that this is tested with a version of the bundle that comprises the sketch header/footer change? 

Otherwise the linker error would make perfekt sense as the footer is not used at all.

obra:
Unless something is really weird with Travis, I do believe those have been merged and pulled up. I did a fresh check out of master for the bundle  and the commits were there. I’m flying to China today and can’t dig further yet, but am happy to do so.

noseglasses:
It would be nice to merge this as those build warnings with gcc 7 are a bit annoying. 
Travis failing is still unrelated.

algernon:
Well, current bundle master works with Travis, I restarted the build of this PR so that it picks up the fixed bundle, but it still fails the same way, even though it is visibly using the new bundle :/

(Can't reproduce the fail locally, either, not even when I clear my ccache and /tmp :/)

noseglasses:
This must be related to the sketch footer not being considered.
I wish one could access Travis' build artifacts easily. Then, debugging this problem would not be too difficult.

obra:
You could have the build process cat the artifact you care about...
ᐧ

noseglasses:
I'd like to check what's in the prepocessed sketch and in the sketch object. The former could be solved with cat.
@obra, to save me some time digging into the (scripted?) Travis configuration of the bundle repo, maybe you already have an idea where to place the `cat` command?

noseglasses:
I'm giving up for now as I am not even sure which version of the `kaleidoscope-builder` bash script is used for testing. Is it the one that comes with the tested repo (here Kaleidoscope) or with the bundle that is used for testing? Seems like the latter.

I tried to add a `cat` command directly after the invocation of `arduino-builder` in `kaleidoscope-builder` (see the last commits of this PR) but Travis does not output anything.

On the other hand output works during a normal local test run e.g. with `make smoke-examples`.

Any hints appreciated.

algernon:
I think the problem here is that we're using _some_ parts of the Kaleidoscope in the bundle, yes. I have an idea how to fix that, and will test it shortly.

algernon:
Hm, interesting. I managed to get an entirely different failure when moving the hardware dir outside of `pwd`.

noseglasses:
@algernon, thanks for working on this.

algernon:
Ooof. I have absolutely no idea what's happening now. At least I managed to make it use the new Kaleidoscope, so that's something.

algernon:
THERE WE GO!

Quick fix: don't just remove the `library.properties` of the bundle's kaleidoscope, but replace the whole dir with ours.

I think the problem here is that we were using kaleidoscope-builder from the bundle, and that adds stuff onto the include path based on its own path. So we need to use the kaleidoscope-builder from the current Kaleidoscope. That'll be the proper fix.

Thankfully, this is completely unrelated to the the PR. We just happened to not trip over the problem before.

algernon:
I'll prep a PR with the fixes, rebase this on top of master once that's merged, and Travis should be happy.

obra:
It looks like travis is no longer segfaulting, but -is- showing other (possibly legitimate?) build errors.

> Building examples/Basic/Basic 0.0.1 into /home/travis/build/keyboardio/Kaleidoscope/.kaleidoscope-build-cache/sketch/1191366-Basic.ino/output...
/home/travis/build/keyboardio/current-libraries/Kaleidoscope/src/kaleidoscope/device/keyboardio/twi.c: In function '__vector_36':
/home/travis/build/keyboardio/current-libraries/Kaleidoscope/src/kaleidoscope/device/keyboardio/twi.c:447:47: warning: this statement may fall through [-Wimplicit-fallthrough=]
     twi_masterBuffer[twi_masterBufferIndex++] = TWDR;
/home/travis/build/keyboardio/current-libraries/Kaleidoscope/src/kaleidoscope/device/keyboardio/twi.c:449:3: note: here
   case TW_MR_SLA_ACK:  // address sent, ack received
   ^~~~
/home/travis/build/keyboardio/Kaleidoscope/hardware/keyboardio/avr/libraries/KeyboardioScanner/twi.c: In function '__vector_36':
/home/travis/build/keyboardio/Kaleidoscope/hardware/keyboardio/avr/libraries/KeyboardioScanner/twi.c:445:47: warning: this statement may fall through [-Wimplicit-fallthrough=]
     twi_masterBuffer[twi_masterBufferIndex++] = TWDR;
/home/travis/build/keyboardio/Kaleidoscope/hardware/keyboardio/avr/libraries/KeyboardioScanner/twi.c:447:3: note: here
   case TW_MR_SLA_ACK:  // address sent, ack received
   ^~~~
/home/travis/build/keyboardio/Kaleidoscope/arduino-1.8.10/hardware/arduino/avr/cores/arduino/new.cpp:29:6: warning: the program should also define 'void operator delete(void*, unsigned int)' [-Wsized-deallocation]
 void operator delete(void * ptr) {
      ^~~~~~~~
/home/travis/build/keyboardio/Kaleidoscope/arduino-1.8.10/hardware/arduino/avr/cores/arduino/new.cpp:33:6: warning: the program should also define 'void operator delete [](void*, unsigned int)' [-Wsized-deallocation]
 void operator delete[](void * ptr) {
      ^~~~~~~~
/home/travis/build/keyboardio/Kaleidoscope/.kaleidoscope-build-cache/sketch/1191366-Basic.ino/build/libraries/Kaleidoscope/Kaleidoscope.a(layers.cpp.o):(.data._ZN12kaleidoscope6Layer_6getKeyE+0x0): undefined reference to `kaleidoscope::Layer_::getKeyFromPROGMEM(unsigned char, kaleidoscope::MatrixAddr<(unsigned char)4, (unsigned char)16>)'
exit status 1
make: *** [travis-smoke-examples] Error 1
The command "make travis-test KALEIDOSCOPE_TEMP_PATH=$(pwd)/.kaleidoscope-build-cache BOARD_HARDWARE_PATH=$(pwd)/hardware" exited with 2.

algernon:
Once this PR is rebased on top of #807, it should no longer fail with the linker error you saw. The segfaults only occurred with #807, and did not without. I've restarted that build, and we'll see how it goes.

obra:
@noseglasses I...seem to have been able to use pragmas to suppress the warnings.

Can you have a look at https://github.com/keyboardio/Kaleidoscope/commit/d5c6bf28dae8e6e3c86a7c73c892b13ce84d48dd and tell me if this change seems like a reasonable solution to this issue?

noseglasses:
> Can you have a look at d5c6bf2 and tell me if this change seems like a reasonable solution to this issue?

@obra, seems reasonable. 

What really surprises me is that wrapping the accessors with pragmas fixes the problem. I understand the compiler warnings being about the array containing the `Key` objects rather than class `Key` itself. To solve this riddle one would probably need to dive into the compiler's implementation. But that's definitely not worth the hassle.

obra:
Fixed by d5c6bf28dae8e6e3c86a7c73c892b13ce84d48dd

PR #776, opened 2019-12-19
Added a compiler/Arduino version check

noseglasses:
Kaleidoscope fails to build with gcc version 4.8 that comes with Arduino 1.8.5. Arduino 1.8.6 comes with gcc 5.4 and builds fine.

This PR adds a compiler version check. The build fails if the avr-gcc version is too old.
A verbose error message reports Arduino upgrade information.

This is meant to prevent issues like https://github.com/keyboardio/Kaleidoscope/issues/773.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ If you're just getting started with the Keyboardio Model 01, the [introductory d
 
 ## Setup the Arduino IDE
 
-Setup the Arduino IDE on your system. Make sure you install at least version 1.6, since older version may not support all required features.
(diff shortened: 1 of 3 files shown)


obra:
On the one hand, I really don't like breaking things that used to work or forcing much newer dependencies unintentionally. 

On the other hand, GCC 4.9 is ancient and I'm thrilled Arduino has finally gotten to a modern compiler and the upgrade story for users is "upgrade a single app"

Approved, with thanks.

PR #778, opened 2019-12-21
Tweak blazing trail decay

tremby:
This patch alters the LED-Stalker plugin.

- Rewrite decay code to fade brightness smoothly with an x^4
  relationship
- Fade hue linearly from orange to red rather than for a subset of
  time from red to orange

I can't find it any more, but when I originally wrote this someone pointed out that this requires floating point logic, which is in some way expensive.

I'm submitting it again in case that's no longer an issue for whatever reason.

In my opinion the effect looks *much* nicer with these tweaks! Maybe it's worth it? ☺

### DIFF
--- a/src/kaleidoscope/plugin/LED-Stalker.cpp
+++ b/src/kaleidoscope/plugin/LED-Stalker.cpp
@@ -111,14 +111,18 @@ uint8_t hue_end = 0;
 cRGB BlazingTrail::compute(uint8_t *step) {
   cRGB color;
 
-  // Get a float from 0 to 1 for our position in the animation
-  float pos = (0xff - *step) / (float) 0xff;
+  // Get the position in the animation, where 0 is start and 0xff is end
+  uint8_t pos255 = 0xff - *step;
 
   // Fade hue linearly from orange to red
-  uint8_t hue = hue_start + pos * (hue_end - hue_start);
+  uint8_t hue = hue_start + (pos255 * (hue_end - hue_start) >> 8);
 
-  // Fade value from full following a -x^4+1 curve
-  uint8_t val = 0xff * (1 - pos * pos * pos * pos);
+  // Fade value from full following a 1-x^4 curve
+  uint8_t val =
+    0xff // Maximum brightness
+    - ((uint32_t) pos255 * pos255 * pos255 * pos255 // Animation position to 4th power
+       >> 24) // ...pulled down to 8-bit range (but this has a maximum of 0xfc rather than 0xff)
+    - pos255 / (0x100 / 4); // Correction to bring the end result into a full 0 to 0xff range
 
   color = hsvToRgb(hue, 0xff, val);
 
--- a/src/kaleidoscope/plugin/LED-Stalker.cpp
+++ b/src/kaleidoscope/plugin/LED-Stalker.cpp
@@ -105,8 +105,8 @@ cRGB Haunt::compute(uint8_t *step) {
 // BlazingTrail
 BlazingTrail::BlazingTrail(void) {
 }
-uint8_t hue_start = 50.0 / 360 * 0xff;
-uint8_t hue_end = 0;
+constexpr uint8_t hue_start = 50.0 / 360 * 0xff;
+constexpr uint8_t hue_end = 0;
 
 cRGB BlazingTrail::compute(uint8_t *step) {
   cRGB color;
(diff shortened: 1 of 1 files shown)


noseglasses:
I had some fun playing with the code of this PR regarding the question how much impact floats have against integers on AVR. The answer is yes, floating point logic is quite expensive there. That's because everything needs to be emulated AFAIK.

In many cases, including this one, floating point math can be carefully replaced by integer logic.

Below I paste two versions of the hue/val computation, one with and one without floating point math. I had to add a linear correction function to the integer version to really generate a zero `val` for `step==0`. 

I [godbolted](https://www.godbolt.org/z/Z_GiUJ) the code. This demonstrates what a big difference in terms of assembly instructions it makes (78 vs. 31 lines). Also the float version requires 11 additional invocations of floating point auxiliary functions while the integer version goes with only one additional call to an auxiliary function. Those auxiliary functions would not be part of the firmware binary for the integer version. And that's what probably makes the greatest difference in terms of PROGMEM consumption.

```cpp
constexpr uint8_t hue_start = 50.0 / 360 * 0xff;
constexpr uint8_t hue_end = 0;

void h(uint8_t step, uint8_t &hue, uint8_t &val) {
    float pos = (0xff - step) / (float) 0xff;
    hue = hue_start + pos * (hue_end - hue_start);
    val = 0xff * (1 - pos * pos * pos * pos);
}

void hInt(uint8_t step, uint8_t &hue, uint8_t &val) {
    
    uint8_t a = 0xff - step;
    
    uint16_t b = a*(hue_end - hue_start);
    hue = hue_start + (b >> 8);

    uint32_t c = a*a;

   // The next two lines could be merged to save some instructions
    uint8_t cor = 3 - (uint16_t(3*step) >> 8);
    
    val = 0xff - (uint8_t((c*c) >> 24)) - cor; 
}
```

I suggest this PR to b

tremby:
That is very illuminating; thank you!

I'm going to spend a little time now reading it over in more detail and I'll comment again soon.

One question, though -- what if other plugins or other code elsewhere in Kaleidoscope is using floating point? Does that mitigate some of the extra bloat, if the emulation logic is shared? Is there really nowhere else that uses floating point currently? (I realize that still doesn't help with the extra processing required each time they're used.)

algernon:
If anything else would be using floating point, that'd mitigate the problem, yes. The only other plugin that uses floating point is Heatmap, which is kinda rarely used and is a wasteful thing as it is. Nothing else is using floating point.

(And at some point, I'd love if Heatmap wouldn't use it either)

tremby:
Understood. Trying to wrap myself around how this int maths is working. I'm going to flash one of my other M01s with the int code to compare :)

tremby:
Just realized I had hue in a uint8_t, but integer hues can go up to 359. Oops!

tremby:
Come to think of it, it doesn't support the circular nature of hue as is, and the start and end points are hardcoded, so it doesn't actually matter.

tremby:
@noseglasses I'm not sure exactly what was wrong but your code was causing the light to fade out for a while and then back in again, before cutting out.

I reimplemented it and got it looking identical (to my eye) to my float-based code.

I'm not sure exactly what the change that fixed the issue was. 

I changed how the correction works -- just divide pos255 (which you had named `a`) by a quarter of 255. Seems to do the job fine -- or is there a reason you did the correction the way you did? I think it's essentially the same thing but your code ended up giving a value range of 254 through 0 rather than 255 through zero -- I think you should have multiplied by 4 rather than 3.

obra:
Indeed, we have a blanket prohibition on floating point in the kaleidoscope core because of the amount of overhead it adds

tremby:
Ah, that makes it much less surprising!

Anyway, this code is ready for review. I could give a video of before vs
after if that would help.

noseglasses:
@tremby, I could only run the two versions of the code on x86 and compare the values because I did not have my M01 with me.
Glad you figured out how to fix it. Must be some subtile differences between the two compilers/architectures.
I was aware of the range 0.…254 but hoped it wouldn't matter too much visually. In general I guess, the closer you want to get with the integer version to the original the more instructions you have to spend.

noseglasses:
@tremby, `hue_start` and `hue_end` are constant and only used in one place. It might be worth turning them `constexpr` to allow the compiler to optimize.

tremby:
Done; thanks.

algernon:
Ooooh, this is *much much* smoother and nicer indeed. And the cost ain't too high either (-22 progmem, +23 ram), and the code is not only more readable, it is better commented too.

Thank you a lot for making the animation amazing! <3

(And big thanks to @noseglasses too, for the help in de-floating it!)

PR #782, opened 2020-01-02
Revamp the Atreus2 layout.

technomancy:
In particular, having backtick and backslash on the fn layer wasn't
helpful since they have dedicated keys now.

Also for the most part, the shift+number punctuation keys are in
sequential order, which makes them a lot easier to learn. (The
exception is the # key, which is moved down to make room for the arrow
cluster.)

The parens are moved so they're surrounding the arrows on the home
row, and the curly brackets are moved to mirror the square brackets.

A redundant backspace is added so it can be hit with either hand now.

https://atreus.technomancy.us/i/layout-new.png


obra:
Thanks. Merging this in.

PR #814, opened 2020-02-24
Add Atreus to udev rules file.

technomancy:


obra:
Thanks! Merging now. I may tweak the symlink when I tweak the name of the sketch to match the product name we're showing customers. (Since it's getting sold as the Keyboardio Atreus rather than the Atreus 2)

PR #818, opened 2020-02-25
macOS device cleanup

indirect:
Testing out the new, much nicer macOS device detection (and automatic bootloading!) I noticed a few environment variables weren't getting set, and cleaned up a couple of warnings.


obra:
<3

PR #824, opened 2020-03-21
Feature/flash qmk

WildlyManifest:
I made these changes in order to flash my Atreus, which had QMK firmware and the Caterina bootloader that came with the AStar. 

The QMK firmware did not expose any serial ports until the keyboard was reset, so it could not be reset by sending a command from the script. A different method was necessary. 

The other changes were some improvements to the port detection logic for macos.

To flash an Atreus or another keyboard that has QMK, the QMK_FLASH flag needs to be set. In addition, the bootloader pid and vid needs to be set to the values defined by the bootloader that is currently installed. E.g. for my device:
```
QMK_FLASH=1
BOOTLOADER_VID="0x1ffb"
BOOTLOADER_PID="0x0101"
```

### DIFF
--- a/bin/find-device-port-macos
+++ b/bin/find-device-port-macos
@@ -66,8 +66,9 @@ sub try_for_location_id {
     my $location_id = shift;
 
     # Here, also, the final character is always a "1", so if macOS ever stops doing that, this
-    # will need an update, as well.
-    my $loc = substr($location_id, 2, 3);
+    # will need an update, as well. 
+    my $len = index($location_id, "0", 2) - 2;
+    my $loc = substr($location_id, 2, $len);
     exit_with_port_if_exists("/dev/cu.usbmodem" . $loc . 1);
 }
 
--- a/bin/find-device-port-macos
+++ b/bin/find-device-port-macos
@@ -10,6 +10,8 @@ use strict;
 my $vid = shift @ARGV;
 my $pid = shift @ARGV;
 
+exit 0 unless defined $vid && defined $pid;
+
 # ioreg might be more machine-readable than system_profiler, but I haven't been able to
 # get it to produce useful output
 my @output = qx(/usr/sbin/system_profiler SPUSBDataType 2> /dev/null);
@@ -52,14 +54,13 @@ sub try_for_raw_serialnum {
     # filename. I'm not sure why, but system_profiler has a serial number ending in "E",
     # whereas the device filename ends in "1". In fact, when I change HID.getShortName()
     # to return "kbio02", the final character is replaced with a "1", so we should do the
-    # same here. If the serial number doesn't end in a digit, however, we may want to append
-    # rather than replace the last character with a 1.
-    if ($serial_port_name =~ /\d$/) {
-        $serial_port_name =~ s/.$/1/;
-    } else {
-        $serial_port_name .= "1";
-    }
+    # same here. 
+    exit_with_port_if_exists($serial_port_name . "1");
+    
+    # now try with the last character chop-ped off 
+    chop $serial_port_name;
     exit_with_port_if_exists($serial_port_name);
+    exit_with_port_if_exists($serial_port_name . "1");
 }
 
 sub try_for_location_id {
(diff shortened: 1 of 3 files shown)


bin/kaleidoscope-builder
```diff
@@ -160,37 +159,65 @@ prompt_before_flashing () {
...
+            find_bootloader_ports
+        else
+            # This is defined in the (optional) user config.
+            # shellcheck disable=SC2154
+            ${preFlash_HOOKS}
+        
+            # Keyboards with QMK firmware must be reset manually to enter bootloader mode.
+            echo "Reset your keyboard now to initiate flashing."
+            wait_for_bootloader_port
+        fi
```
algernon: So, the basic difference here is the manual reset, right? I'd suggest not tieing the name to QMK, because a manual reset can be useful even with Kaleidoscope. So how about we call it `MANUAL_RESET`? And narrow the paths a little, somewhat like his:

```suggestion
        prompt_before_flashing
        # This is defined in the (optional) user config.
        # shellcheck disable=SC2154
        ${preFlash_HOOKS}

        if [ -z "${MANUAL_RESET}" ]; then          
           reset_device
           sleep 2
           find_bootloader_ports
        else
           wait_for_bootloader_port
        fi
```

I do not think we need the extra text, because the pre-flashing prompt already tells us to put the keyboard into flash mode. A manual reset does that.

bin/find-device-port-macos
```diff
@@ -52,32 +54,33 @@ sub try_for_raw_serialnum {
...
 }
 
 sub try_for_location_id {
     my $location_id = shift;
 
     # Here, also, the final character is always a "1", so if macOS ever stops doing that, this
-    # will need an update, as well.
-    my $loc = substr($location_id, 2, 3);
-    exit_with_port_if_exists("/dev/cu.usbmodem" . $loc . 1);
+    # will need an update, as well.
```
obra, line 70: I don't understand what this code change does or why. Can you add some more comments?

bin/find-device-port-macos
```diff
@@ -52,32 +54,33 @@ sub try_for_raw_serialnum {
...
 sub try_for_location_id {
     my $location_id = shift;
 
     # Here, also, the final character is always a "1", so if macOS ever stops doing that, this
-    # will need an update, as well.
-    my $loc = substr($location_id, 2, 3);
-    exit_with_port_if_exists("/dev/cu.usbmodem" . $loc . 1);
+    # will need an update, as well. 
+    if ($location_id =~ /0x(\d+?)0*\b/) { 
+        exit_with_port_if_exists("/dev/cu.usbmodem" . $1 . 1);
```
obra, line 72: using $1 directly there can be a little bit brittle and not assigning it to a variable misses an opportunity for a descriptive name to help make the code more readable.

What problem was this change fixing?

bin/find-device-port-macos
```diff
@@ -52,32 +54,33 @@ sub try_for_raw_serialnum {
...
+    }
 }
 
 sub try_for_sn_prefix {
     my $sn = shift;
+    chop $sn if length $sn > 1;
```
obra, line 78: A bit of doc here, please

bin/find-device-port-macos
```diff
@@ -10,6 +10,8 @@ use strict;
 my $vid = shift @ARGV;
 my $pid = shift @ARGV;
 
+exit 0 unless defined $vid && defined $pid;
```
obra, line 13: 0 is a success condition for exiting. This should probably be something like 


```
if (!defined $vid || !defined $pid) {
  die "$0 has two required parameters, VID and PID";
}
```

obra:
Sorry it took a bit to review this.

We should clean up the history of this PR to have separate logical commits for each logical change. 

Whitespace changes, if any, should be in their own commits. 

For example, https://github.com/keyboardio/Kaleidoscope/pull/824/commits/1d1c18d65b186241a14ba7eb797fa967ca884c3b mixes a bunch of seemingly unrelated changes and whitespace changes.

I don't understand the reasons behind some of the changes to the OSX port detection logic. If there are specific issues being solved, ideally doc them inline. Failing that, explain them in the commit message.

Thanks!

bin/find-device-port-macos
WildlyManifest: From my limited experimentation, macOs does truncate the last character and replace it with '1' even if it is not a digit. This way, all possibilities are covered.

bin/find-device-port-macos
WildlyManifest: The location string is not necessarily 3 chars long.

WildlyManifest:
Hi Jesse, 
I think I have addressed your review comments. The only thing I am not sure about is cleaning up the commits; I dont really know how to go about doing that.

Jason Koh

obra:
Hi Jason,

Thanks so much for the updates!
Don't worry about that. I'm happy to take care of it :)

For the future, "git rebase -i origin/master" is the command that kicks off
that kind of workflow -
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History is a tutorial
that might help some, but this is very much one of the bits of git
that's....well, I don't want to say that it's not user friendly, but it's
much pickier about who its friends are.

Best,
Jesse

obra:
@WildlyManifest Wanna have a look at #841 - I took your branch, redid the history and cleaned up a couple more things. I should be able to test it on a real mac later today. I'm closing this PR, so discussion should move over there.  Thanks so much for all the hacking!

PR #827, opened 2020-04-07
Feature/optional filename paths

akerl:
I have a CI build that creates the .hex files, and then a helper script to call kaleidoscope-builder to flash later. Without this PR, flash picks a temporary dir and tries to rebuild the firmware. This allows overriding the paths, similarly to other paths in the builder script.


obra:
Thanks! Merged

PR #834, opened 2020-04-27
System control keys on mac

obra:
System control keys on macOS were not functional, since we were deferring the press until immediately before release, which doesn't work for how Apple handles OSC System Control keys. Instead, we should send one press event at press time and one release event at release time.

### DIFF
--- a/src/kaleidoscope/key_events.cpp
+++ b/src/kaleidoscope/key_events.cpp
@@ -32,9 +32,9 @@ static bool handleSyntheticKeyswitchEvent(Key mappedKey, uint8_t keyState) {
     if (keyIsPressed(keyState))
       Runtime.hid().keyboard().pressConsumerControl(mappedKey);
   } else if (mappedKey.getFlags() & IS_SYSCTL) {
-    if (keyIsPressed(keyState)) {
-    } else if (keyWasPressed(keyState)) {
+    if (keyToggledOn(keyState)) {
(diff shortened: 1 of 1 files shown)


algernon:
Looks like this is already on master?

Looks sane to me nevertheless.

obra:
Ugh. Unintentional. Glad it got approved, then ;)

gedankenexperimenter:
Looking at the code, I believe there's still a bug. If the user rolls over from one system control key to another, the second one will get released as soon as the first one gets released.

1. Key `A` toggles on -> `A` report sent
2. Key `B` toggles on -> `B` report sent, clobbering `A` (correctly)
3. Key `A` toggles off -> empty report sent (incorrectly indicating that `B` was released)
4. Key `B` toggles off -> another empty report sent

When a system control key is released, it shouldn't unconditionally send the empty report. Instead, it should check to verify that the released key is the same as the last one that was sent. To do this right, it's necessary to store that state, probably in KeyboardioHID.

obra:
Hm. That's an interesting corner case. I'll open up a new, separate
issue for it.  I agree that it should be fixed, but it's also
exceedingly unlikely that anyone will run into it. As far as I know,
there just aren't that many system-control keys that are actually
supported by host operating systems.

obra:
Opened #840 to track the issue @gendankenexperiementer noticed.

Issue #845, opened 2020-05-15
Map Shift+LED to previous LED effect


xerus2000:
As fas as I'm aware, Shift+LED currently does nothing, and I don't see a way to map it from Kaleidoscope. Couldn't this be mapped to go to the previous LED effect by default?

obra:
That shouldn’t it be possible to do

alfalfasprossen:
Technically `LEDControl` could check the state of the shift modifiers when interpreting the `Key_LEDEffectNext` / `Key_LEDEffectPrevious` key press. But that may not work reliably when plugins are involved that handle shift modifiers like `OneShot` - and having LEDControl access those plugins is unlikely to be manageable or desirable.

You can achieve this in your sketch using a Macro though. If you use OneShot, you may need to check the state of shift with it as well, but I didn't try it.

```c++
#include "Kaleidoscope.h"
#include "Kaleidoscope-Macros.h"
#include "Kaleidoscope-LEDControl.h"
#include "Kaleidoscope-LEDEffect-SolidColor.h"

enum { MACRO_NEXT_PREV_LED_MODE };

// *INDENT-OFF*
KEYMAPS(
  [0] = KEYMAP_STACKED
  (Key_NoKey, Key_1, Key_2, Key_3, Key_4, Key_5, M(MACRO_NEXT_PREV_LED_MODE),
   Key_Backtick,      Key_Q, Key_W, Key_E, Key_R, Key_T, Key_Tab,
   Key_PageUp,        Key_A, Key_S, Key_D, Key_F, Key_G,
   Key_PageDown,      Key_Z, Key_X, Key_C, Key_V, Key_B, Key_Escape,

   Key_LeftControl, Key_Backspace, Key_LeftGui, Key_LeftShift,
   XXX,

   Key_skip,  Key_6, Key_7, Key_8,     Key_9,      Key_0,         Key_skip,
   Key_Enter, Key_Y, Key_U, Key_I,     Key_O,      Key_P,         Key_Equals,
              Key_H, Key_J, Key_K,     Key_L,      Key_Semicolon, Key_Quote,
   Key_skip,  Key_N, Key_M, Key_Comma, Key_Period, Key_Slash,     Key_Minus,

   Key_RightShift, Key_RightAlt, Key_Spacebar, Key_RightControl,
   XXX),
)
// *INDENT-ON*


static void nextPrevLEDModeMacro(uint8_t keyState) {
  if (keyToggledOn(keyState)) {
[...]

PR #849, opened 2020-06-08
Use shift state to invert next/prev LED effect key direction

alfalfasprossen:
Handling of the respective keys is moved to `beforeReportingState`
as the state of shift is only then available. Usage together with
OneShot requires OneShot to be initialized before LEDControl.

A simple sketch can be found here https://github.com/alfalfasprossen/Kaleidoscope-Sketches/blob/master/experiments/shiftledtest/shiftledtest.ino

Related issue: #845


obra:
Thanks!

We...really ought to have a better affordance for whether modifiers are active, but that's not your responsibility.

The only thing this needs for my +1 is a "make astyle" pass to fix an indent issue picked up by the automated linter. If you don't get to it today, I ought to be able to tomorrow.

obra:
I touched up the indentation and merged. Thanks!

PR #856, opened 2020-06-18
SpaceCadet:  Fix key injection when many pressed at once

cscorley:
Hi team,

I ran into an issue recently where having the following mapping for SpaceCadet was causing a peculiar issue at work.

```
  static kaleidoscope::plugin::SpaceCadet::KeyBinding spacecadetmap[] = {
      {Key_LeftShift, Key_LeftParen},
      {Key_RightShift, Key_RightParen},
      {Key_LeftGui, Key_LeftCurlyBracket, 250},
      {Key_RightAlt, Key_RightCurlyBracket},
      {Key_LeftAlt, Key_RightCurlyBracket},
      {Key_LeftControl, Key_LeftBracket},
      {Key_RightControl, Key_RightBracket},
      SPACECADET_MAP_END};
```

These were great but also do not allow for quick combos like `Ctrl + Shift + B`.  Doing this would produce `Ctrl+B` and the `Shift` would be dropped.  Swapping the order of the mapping so that the Ctrl mappings came first would produce a `Shift+B` and the `Ctrl` would be dropped.  I found myself having to do `Ctrl + Shift + wait a full second + B`.  

This fixes the issue by pre-checking the existing map flag statuses and aborting the timeouts if two keys are found to have been pressed.  This causes any combination of any mapped keys (e.g., `Ctrl+Shift`) to produce just those inputs again.  I am not sure it would make sense to me to want anything else, but I understand that this might be a backwards-incompatible change and am happy to put this behind a flag that must be opted-in to use.

(Amended commit to sign-off)

### DIFF
--- a/src/kaleidoscope/plugin/SpaceCadet.cpp
+++ b/src/kaleidoscope/plugin/SpaceCadet.cpp
@@ -119,7 +119,7 @@ EventHandlerResult SpaceCadet::onKeyswitchEvent(Key &mapped_key, KeyAddr key_add
     ) {
 
       if (map[i].flagged
-        && map[i].input != mapped_key) {
+          && map[i].input != mapped_key) {
         other_mapped_key_flagged = true;
         break;
       }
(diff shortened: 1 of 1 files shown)


cscorley:
This is almost certainly a duplicate of #506 

That's what I get for not looking at the current PR list

obra:
Thanks! merged.

PR #858, opened 2020-06-18
Guard against nullptr access in LEDControl

### LINKED ISSUE #855: LEDControl with no LEDMode crashes firmware
I've come across a problem that when using `LEDControl` in a sketch without using any `LEDMode`, the firmware will crash, making the keyboard unresponsive.

This is the case with some of the examples like [LED-ActiveModColor.ino](https://github.com/keyboardio/Kaleidoscope/blob/master/examples/LEDs/LED-ActiveModColor/LED-ActiveModColor.ino)

### Possible Cause
The reason, I assume, is that `LEDControl::cur_led_mode_` is accessed, but is an uninitialized pointer. When LEDControl is initialized, we end up at https://github.com/keyboardio/Kaleidoscope/blob/master/src/kaleidoscope/plugin/LEDControl.h#L44
`beforeReportingState()` calls `update()` after the sync delay time has passed.

If we put a check `if (num_led_modes_ > 0)` to guard against accessing `cur_led_mode_` this works. An alternative would be to initialize the pointer to `nullptr / NULL` and do a check for that. 

However there are plugins that may call other functions of `LEDControl` which might again access the pointer. There is also `LEDControl::get_mode()` - at which point checking for a valid pointer would be up to that plugin. I think this is fine though. If you use pointers, you gotta be aware of the risks. All other functions where `cur_led_mode_` is accessed inside `LEDControl` itself could be guarded 🤔 

Another option might be to force adding something like `LEDOff` mode to be present somehow, at least if there are no other modes initialized - I didn't look much into that, I think that this may be more complicated than it sounds, especially if it should be optional to have that effect in a custom firmware.

I think the probability of someone actually using a sketch with LEDControl and no LEDMode is quite low, but there are a few example sketches that do exactly that.

### WARNING
Trying to reproduce this by compiling a respective sketch and flashing it to a Model 01 will cause the device to stop working and you will have to *hold the Prog key while connecting* the device to be able to upload another firmware.

alfalfasprossen:
If LEDControl is used without any LEDMode plugin, `cur_led_mode_` is
an unitialized pointer. This initializes it to `nullptr` and guards
against accessing it.

Fixes #855

### DIFF
--- a/src/kaleidoscope/plugin/LEDControl.cpp
+++ b/src/kaleidoscope/plugin/LEDControl.cpp
@@ -28,7 +28,7 @@ static constexpr uint8_t uninitialized_mode_id = 255;
 
 uint8_t LEDControl::mode_id = uninitialized_mode_id;
 uint8_t LEDControl::num_led_modes_ = LEDModeManager::numLEDModes();
-LEDMode *LEDControl::cur_led_mode_;
+LEDMode *LEDControl::cur_led_mode_ = nullptr;
 uint8_t LEDControl::syncDelay = 32;
 uint16_t LEDControl::syncTimer = 0;
 bool LEDControl::enabled_ = true;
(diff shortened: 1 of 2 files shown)


obra:
Lgtm. @algernon  Does this seem reasonable to you?

PR #891, opened 2020-09-03
Add core plugin doc section for Qukeys

gedankenexperimenter:
I happened to notice that the core plugins overview in the documentation didn't have a section for Qukeys, so I added one. It's probably not great, but it's better than nothing.


obra:
Thanks! Merged.

gedankenexperimenter:
I happened to look at the [ReadTheDocs page](https://kaleidoscope.readthedocs.io/en/latest/customization/core-plugins.html) today, and it hasn't changed; when does that get updated?

obra:
The build of docs broke earlier in the week. Thanks. Should be ok now.

PR #893, opened 2020-09-05
Major revisions to OneShot

gedankenexperimenter:
There were a few minor problems and inefficiencies in the existing OneShot plugin. This PR comprises a number of changes to OneShot that should make the plugin more reliable, while also reducing its footprint (-432 bytes PROGMEM, -17 bytes RAM) and hopefully improving the readability of the code.


obra:
That's a heck of a potential improvement. Since we've got @paniag working on test suites, would you be up for laying out some of the problem cases this should solve?

gedankenexperimenter:
Yes. I intended to enumerate them (there weren't a lot, and they were corner cases), but got in a little deep without taking all the notes I should first, so I might not have a complete list. It might be easier to describe the tests that ought to be done (and the expected behaviour) than it is to catch all the fiddly bugs that were present. I will do my best to make a list of the concrete improvements in behaviour, though.

I have a couple more changes to make before I'm done, though.

gedankenexperimenter:
I've got it working well now, and the only remaining bug I'm aware of is not fixable without a major change to Kaleidoscope proper. The current unfair comparison is:
```
  PROGMEM:  -452 bytes
  RAM:       -18 bytes
```
That's unfair because this should really be coupled with #890, which contains some substantial improvements to ActiveModColor. This PR on its own doesn't work quite as well with ActiveModColor as the current master does (though the difference is relatively small; neither is great). With that PR included as well:
```
  PROGMEM:  -210 bytes
  RAM:       -27 bytes
```
Still a substantial improvement, and with just a bit more code we can use different highlight colours for the three effective states of a OneShot key (sticky, one-shot, normal modifier).

src/kaleidoscope/plugin/OneShot.h
```diff
@@ -68,45 +62,85 @@ class OneShot : public kaleidoscope::Plugin {
...
-
-  void inject(Key mapped_key, uint8_t key_state);
 
  private:
+
+  // --------------------------------------------------------------------------
+  // Constants
   static constexpr uint8_t ONESHOT_KEY_COUNT = 16;
-  typedef struct {
+  static constexpr uint8_t ONESHOT_MOD_COUNT = 8;
```
gedankenexperimenter, line 111: Instead of doing math (e.g. `ONESHOT_KEY_COUNT / 2`), I think it's clearer and more robust to use two constants. Iteration indices can start and end at whichever constant is appropriate, and if we wanted to change the number of supported OneShot layers, it wouldn't require changing anything but `ONESHOT_KEY_COUNT`.

src/kaleidoscope/plugin/Escape-OneShot.cpp
```diff
@@ -19,26 +19,30 @@
...
-EventHandlerResult EscapeOneShot::onKeyswitchEvent(Key &mapped_key, KeyAddr key_addr, uint8_t keyState) {
-  if (mapped_key != Key_Escape || (keyState & INJECTED))
-    return EventHandlerResult::OK;
-
-  if (did_escape_)
-    mapped_key = Key_NoKey;
-  did_escape_ = !keyToggledOff(keyState);
-
-  if ((!::OneShot.isActive() || ::OneShot.isPressed()) && !::OneShot.isSticky()) {
-    return EventHandlerResult::OK;
```
gedankenexperimenter, line 37: I found this test condition very confusing for some reason.

src/kaleidoscope/plugin/Escape-OneShot.cpp
```diff
@@ -19,26 +19,30 @@
...
+EventHandlerResult EscapeOneShot::onKeyswitchEvent(Key &mapped_key, KeyAddr key_addr, uint8_t key_state) {
+  // We only act on an escape key that has just been pressed, and not
+  // generated by some other plugin. Also, only if at least one
+  // OneShot key is active and/or sticky. Last, only if there are no
+  // OneShot keys currently being held.
+  if (mapped_key == Key_Escape &&
+      keyToggledOn(key_state) &&
+      !(key_state & INJECTED) &&
+      ::OneShot.isActive() &&
+      !::OneShot.isPressed()) {
```
gedankenexperimenter, line 36: This new test condition is required because the `pressed` flag wasn't being cleared when `sticky` OneShot keys were released. With that fixed in OneShot, it needed to change here.

src/kaleidoscope/plugin/Escape-OneShot.cpp
```diff
@@ -19,26 +19,30 @@
...
+    // Cancel all OneShot keys
+    ::OneShot.cancel(true);
+    // Suppress the escape key event, and mask the key until it's released
+    Layer.updateLiveCompositeKeymap(key_addr, Key_NoKey);
```
gedankenexperimenter, line 40: This is the big change here. Rather than using the `did_escape_` boolean and blocking the pressed Escape key every cycle, we update the keymap cache so that it will find a blank key, which won't be processed at all.

src/kaleidoscope/plugin/OneShot.h
```diff
@@ -68,45 +62,85 @@ class OneShot : public kaleidoscope::Plugin {
...
+  // --------------------------------------------------------------------------
+  // Single-key test functions
+
+  static bool isOneShotKey(Key key) {
+    return (key.getRaw() >= kaleidoscope::ranges::OS_FIRST &&
+            key.getRaw() <= kaleidoscope::ranges::OS_LAST);
+  }
+  static bool isActive(Key key);
+  static bool isSticky(Key key);
+  static bool isPressed(Key key);
```
gedankenexperimenter, line 81: This new function is intended for use by ActiveModColor.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
 
-      activateOneShot(idx);
-    }
+// Is this function used by anything? It's part of the public API, but
+// none of the core plugins use it. Calling the event handler doesn't
+// seem like the right thing to do here, so perhaps the function
+// should be deprecated, then removed.
+void OneShot::inject(Key mapped_key, uint8_t key_state) {
+  onKeyswitchEvent(mapped_key, UnknownKeyswitchLocation, key_state);
+}
```
gedankenexperimenter, line 164: I can't figure out a use case for this `inject()` function. None of the core plugins uses it, and I'm suspicious of calling the event handler like this, rather than calling `handleKeyswitchEvent()`.

src/kaleidoscope/plugin/OneShot.h
```diff
@@ -68,45 +62,85 @@ class OneShot : public kaleidoscope::Plugin {
...
   static bool isStickable(Key key);
 
+  // --------------------------------------------------------------------------
+  // Utility function for other plugins to cancel OneShot keys
+  static void cancel(bool with_stickies = false);
+
+  // --------------------------------------------------------------------------
+  // Vestigial functions?
+  void inject(Key mapped_key, uint8_t key_state);
   static bool isModifierActive(Key key);
```
gedankenexperimenter, line 91: These two functions don't seem to provide anything useful. Perhaps @algernon can enlighten me?

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
 
-bool OneShot::isPressed() {
-  for (uint8_t i = 0; i < ONESHOT_KEY_COUNT; i++) {
-    if (state_[i].pressed)
+// ----------------------------------------------------------------------------
+// State variables
+
+OneShot::OneShotKeyState OneShot::state_[OneShot::ONESHOT_KEY_COUNT];
+uint16_t OneShot::start_time_ = 0;
+KeyAddr OneShot::prev_key_addr_ = UnknownKeyswitchLocation;
```
gedankenexperimenter, line 37: Instead of keeping track of the `Key` value of the last key pressed, I changed it to the address of the key, which is more certainly unique, and needs only one byte of RAM. This does mean that two quick taps (or even rollover) of a pair of identically-mapped OneShot keys will no longer result in that OneShot modifier/layer becoming sticky. This is worth considering, even though it's not mentioned in the documentation of the behaviour, because there might be users out there who make an OSL sticky by, for example, simultaneously tapping both palm keys on a Model01. That could become officially supported as a second way of making a OneShot sticky, but I have not done so yet.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
+uint8_t OneShot::release_countdown_ = 0;
```
gedankenexperimenter, line 38: Rather than use the boolean value `should_cancel_`, I use this `release_countdown_` value to delay the release of OneShot keys by one cycle so that they will affect the keypress they're intended to.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
+      prev_key_addr_ = key_addr;
+      if (!(mapped_key >= Key_LeftControl && mapped_key <= Key_RightGui) &&
+          !(mapped_key.getFlags() == (KEY_FLAGS | SYNTHETIC | SWITCH_TO_KEYMAP))) {
+        // We can't immediately send the release event(s) for active
+        // OneShot keys here, because that would release those keys
+        // before the report in which the keycodes they are meant to
+        // modify gets sent. Instead, we use this variable, which will
+        // get bit-shifted each cycle, and trigger the release when it
+        // gets to 1 (which is why it's written as a bit-shifted 1 here).
+        release_countdown_ = 1 << 1;
```
gedankenexperimenter, line 205: I'm writing the value `2` here as `1 << 1` because it's going to be bit-shifted down once per cycle. When only the low bit is set, it will trigger the release of OneShot keys. Most of the time, it will be zero.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
-
-    return EventHandlerResult::EVENT_CONSUMED;
+    return EventHandlerResult::OK;
   }
 
-  // ordinary key here, with some event
+  // We know that `mapped_key` is a OneShot key, and the call to
+  // `isOneShotKey_()` above ensures that the index is not out of
+  // bounds, so this is now valid:
+  uint8_t n = mapped_key.getRaw() - ranges::OS_FIRST;
```
gedankenexperimenter, line 214: Among the cosmetic changes I made, I changed the index label `idx` to `n` (when not iterating through a for loop). I find the single-letter version more readable, and I had already altered almost every line of code, so it didn't seem like a big deal to change this, too.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
+    state_[key.getRaw() - ranges::OS_FIRST].stickable = true;
+}
+
+void OneShot::disableStickability(Key key) {
+  if (key >= ranges::OS_FIRST && key <= ranges::OS_LAST)
+    state_[key.getRaw() - ranges::OS_FIRST].stickable = false;
+}
+
+void OneShot::enableStickabilityForModifiers() {
+  for (uint8_t i{0}; i < ONESHOT_MOD_COUNT; ++i) {
```
gedankenexperimenter, line 58: I changed all the for loops to this style, using brace initialization and the prefix increment operator. I had already written several like that before I noticed that the existing code used a different style. Either of these elements can be changed back from `i{0}` to `i = 0` and `++i` to `i++` if preferred; the compiled binary should be the same either way.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
+}
+
+void OneShot::disableStickabilityForModifiers() {
+  for (uint8_t i{0}; i < ONESHOT_MOD_COUNT; ++i) {
+    state_[i].stickable = false;
+  }
+}
+
+void OneShot::enableStickabilityForLayers() {
+  for (uint8_t i{ONESHOT_MOD_COUNT}; i < ONESHOT_KEY_COUNT; ++i) {
```
gedankenexperimenter, line 70: The existing version of this function (and the one below) used `ONESHOT_KEY_COUNT / 2`, taking advantage of what is essentially a coincidence (there happen to be the same number of modifiers and supported OSL values). I think this illustrates more clearly what's going on, and why.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
+
+// For any active OneShot modifier keys, keep those modifiers active
+// in the keyboard HID report.
 EventHandlerResult OneShot::beforeReportingState() {
-  for (uint8_t i = 0; i < ONESHOT_KEY_COUNT / 2; i++) {
-    if (state_[i].active) {
-      activateOneShot(i);
+  for (uint8_t i{0}; i < ONESHOT_KEY_COUNT; ++i) {
+    if (state_[i].active || state_[i].pressed) {
+      sustainOneShot(i);
```
gedankenexperimenter, line 290: Here are two subtle, but important changes. First, I decided to include the OneShot layer keys as well as the modifiers. It's not necessary, but it more closely emulates what happens if the user holds down a normal layer shift key.
Second, when a OneShot key has been cancelled, it's `active` flag will be cleared, but its `pressed` flag will still be set, so I needed to add the second condition.
Third, `activateOneShot()` (both the old and new version) sends a "key toggled on" event, but in this new version, the toggle-on gets sent by the event handler, and `beforeReportingState()` only needs to send the "key held" event, so I'm using the new `sustainOneShot()` function instead.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
-
-  if (is_cancelled) {
-    should_cancel_ = false;
-    should_cancel_stickies_ = false;
-  }
+  // Also, advance the counter for OneShot keys that have been
+  // cancelled by the press of a non-OneShot, non-modifier key. An
+  // unconditional bit shift should be more efficient than checking
+  // for zero to avoid underflow.
+  release_countdown_ >>= 1;
```
gedankenexperimenter, line 316: This is perhaps too "clever" to be easily understood. I hope the comment I wrote is adequate.

src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -22,266 +22,339 @@
...
+      } else {
+        // If this OneShot key is not "stickable", we deactivate
+        // it. It still remains in effect until it is released.
+        state_[n].active = false;
+      }
+    }
+    prev_key_addr_ = key_addr;
+
+  } else if (keyToggledOff(key_state)) {
+    state_[n].pressed = false;
```
gedankenexperimenter, line 262: I'm now guaranteeing that `pressed` will be true when a OneShot key is pressed, and false when it's not…unless there are two of the same OneShot key pressed at the same time, in which case it will get unset when the first one gets released. This is an unfortunate side effect of tracking `Key` values instead of `KeyAddr` values. The problem could be fixed by searching through the whole keymap for held OneShot keys any time one is released, but I don't think it's worth the trouble.

gedankenexperimenter:
Fixes #894

gedankenexperimenter:
Fixes #895

gedankenexperimenter:
This rewrite does not address #896 (just to be clear).

gedankenexperimenter:
Here's a partial list of testcases for OneShot:

# OneShot Test Cases

## With default configuration

- [ ] tap OSM, tap normal key twice
- [ ] tap OSM, tap two subsequent normal keys
- [ ] tap OSM, roll over two normal keys
- [ ] tap OSM, press normal modifier, check that the OSM stays active
- [ ] tap OSM, press layer shift, check that the OSM stays active
- [ ] tap OSM `X`, tap OSM `Y`, tap `X` again (confirm cancel of `X` but not `Y`) 
- [ ] double-tap OSM, tap multiple normal keys (confirm stickiness)
- [ ] double-tap OSM, tap again to cancel (also with intervening keys)
- [ ] tap OSM, wait for timeout
- [ ] tap multiple OSM keys, wait for timeout (timeout should be based on the last one pressed, and they should all time out together)
- [ ] tap multiple OSM keys, tap normal key, confirm all modifiers applied, and that OSMs are all released
- [ ] make OSM `X` sticky, tap OSM `Y`, tap another key twice, confirm that `Y` is cancelled after the first press, and `X` remains sticky
- [ ] hold OSM to check hold timeout
- [ ] hold OSM past `hold_time_out`, quickly release and tap again
- [ ] hold OSM, tap multiple keys, confirm that the modifier still applies, even after the timeout would expire
- [ ] repeat above tests with OSL keys
- [ ] combine OSM & OSL keys

## With `double_tap_time_out` (< `time_out`) configured

- [ ] tap OSM twice far enough apart to not trigger stickiness, confirm that the key is cancelled by the second press (before timeout)
- [ ] double-tap OSM, confirm that it becomes sticky

## With two identical OSM keys in the keymap

- [ ] tap one OSM, tap the second (behaviour currently undocumented)

gedankenexperimenter:
I converted this back to a draft PR, because I've now rewritten OneShot (along with a small change to Kaleidoscope itself), which fixes all the problems I've discovered, and also allows for the possibility of future enhancements (turning any key into a OneShot key, not just modifiers and layer shifts). I think we ought to discuss what to do, with at least two options. I'll submit that other rewrite as a separate PR, and link it to this one when it's in a better state.

gedankenexperimenter:
I'm closing this because #905 offers substantial functional advantages that this version can't provide.

PR #897, opened 2020-09-10
Add Consumer keys to adjust display brightness

gedankenexperimenter:
This adds `Consumer_BrightnessUp` and `Consumer_BrightnessDown`, which are not included in the USB HID Usage Tables (1.12v2), but which are supported by Linux, macOS, and Windows (according to QMK docs).

References:
- [QMK documentation](https://beta.docs.qmk.fm/using-qmk/simple-keycodes/keycodes)
- [Linux kernel HID input source code](https://github.com/torvalds/linux/blob/7fe10096c1508c7f033d34d0741809f8eecc1ed4/drivers/hid/hid-input.c#L903)

I haven't had a chance to test these keys on Linux or Windows, but they work on my iMac running Catalina.

### DIFF
--- a/src/kaleidoscope/key_defs_consumerctl.h
+++ b/src/kaleidoscope/key_defs_consumerctl.h
@@ -397,3 +397,8 @@
 #define Consumer_AC_Split CONSUMER_KEY(HID_CONSUMER_AC_SPLIT, KEY_FLAGS |  HID_TYPE_SEL)
 #define Consumer_AC_Distribute_Horizontally CONSUMER_KEY(HID_CONSUMER_AC_DISTRIBUTE_HORIZONTALLY, KEY_FLAGS |  HID_TYPE_SEL)
 #define Consumer_AC_Distribute_Vertically CONSUMER_KEY(HID_CONSUMER_AC_DISTRIBUTE_VERTICALLY, KEY_FLAGS |  HID_TYPE_SEL)
+
+// These keys are not documented in the USB HID Usage Tables, but are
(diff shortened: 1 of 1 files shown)


gedankenexperimenter:
Note: These could also be included in [HIDtables.h](https://github.com/keyboardio/Kaleidoscope/blob/c01c55fac4d24ee1b83623607883628cde9e2e2d/src/kaleidoscope/HIDTables.h#L348), but since they don't appear in the HUT document, I didn't include them there. If we do add them to that file, should they be inserted in order (they appear in a "reserved" range—see the link above), or should they be added at the bottom, since they're not from the same source document?

obra:
To my surprise and delight, it turns out that there are a number of (possibly private) errata to the HID tables.

More pleasantly, the first update to the whole document in something like 15 years was published last month.

https://www.usb.org/sites/default/files/hut1_2.pdf

Rather than add stuff piecemeal, I'd much rather update the whole thing starting from KeyboardioHID. (Though we should absolutely be doing that update to KeyboardioHID, Kaleidoscope, and Chrysalis)

gedankenexperimenter:
Aha! I did not realize that the HUT docs had gotten an update. That's great news!

One section of interest is 15.22: Keyboard Backlight Controls…

gedankenexperimenter:
I've started to work on a comprehensive update based on version 1.2 of the HUT document, so I converted this PR to a draft.

gedankenexperimenter:
#901 makes this PR obsolete, so I'm closing it now.

PR #901, opened 2020-09-12
Update USB HID Usage Tables with new System & Consumer keys

gedankenexperimenter:
There has been an update to the [USB HID Usage Tables (v1.2)](https://www.usb.org/sites/default/files/hut1_2.pdf), and it includes several new System Control and Consumer Control keys, some of which are being used already by major operating systems. This PR updates Kaleidoscope's copy of the keycodes table, and adds the new `Key` definitions. I also did some reformatting and added comments to hopefully make the table more readable and make it easier to spot errors.

In addition, I have included changes to make the definition of System Control and Consumer Control keys simpler, and more robust.

This PR includes (and replaces) #899 & #897.


src/kaleidoscope/key_defs.h
```diff
@@ -225,20 +225,6 @@ typedef kaleidoscope::Key Key_;
...
-   We need to keep the bottom two bits clear for defining the keys as sysctl / consumerctl
-*/
-
-#define HID_TYPE_CL    B00000000
-#define HID_TYPE_LC    B00000100
-#define HID_TYPE_NARY  B00001000
-#define HID_TYPE_OOC   B00001100
-#define HID_TYPE_OSC   B00010000
-#define HID_TYPE_RTC   B00010100
-#define HID_TYPE_SEL   B00011000
```
gedankenexperimenter, line 240: These `HID_TYPE_*` constants were being combined with others in the `flags` byte of both Consumer and System Control `Key` definitions, but this was incorrect. These bits overlap with the `IS_CONSUMER` and `SWITCH_TO_KEYMAP` bits, which are used to detect ConsumerControl and Layer keys, respectively.

For example `HID_TYPE_RTC` would be detected in [`Layer.eventHandler()`](https://github.com/keyboardio/Kaleidoscope/blob/49064d679e187f1ca26a4f15cbed05f4309d7d29/src/kaleidoscope/layers.cpp#L103) as `SWITCH_TO_KEYMAP`, and since all of these `Key` types have the `SYNTHETIC` bit set, this would result in any System or Consumer keys with usage type RTC triggering a call to `Layer.handleKeymapKeyswitchEvent()` (which probably wouldn't do anything, because the `keycode` byte would be out of range, but it's still incorrect). Likewise, any System key with `HID_TYPE_OOC` would be misidentified as a Consumer key in [`handleSyntheticKeyswitchEvent()`](https://github.com/keyboardio/Kaleidoscope/blob/49064d679e187f1ca26a4f15cbed05f4309d7d29/src/kaleidoscope/key_events.cpp#L31), and the System key's HID report would never get sent.

src/kaleidoscope/key_defs.h
```diff
@@ -255,14 +241,15 @@ typedef kaleidoscope::Key Key_;
...
+#define SYSTEM_KEY(code) Key(code, SYNTHETIC | IS_SYSCTL)
 
 /* Most Consumer keys are more then 8bit, the highest Consumer hid code
    uses 10bit. By using the 11bit as flag to indicate a consumer keys was activate we can
    use the 10 lsb as the HID Consumer code. If you need to get the keycode of a Consumer key
    use the CONSUMER(key) macro this will return the 10bit keycode.
 */
 #define CONSUMER(key) (key.getRaw() & 0x03FF)
-#define CONSUMER_KEY(code, flags) Key((code) | ((flags | SYNTHETIC|IS_CONSUMER) << 8))
+#define CONSUMER_KEY(code) Key((code & 0x03FF) | ((SYNTHETIC | IS_CONSUMER) << 8))
```
gedankenexperimenter, line 252: This is a simpler and more correct version of `CONSUMER_KEY()`. The high six bits should always be the same on any Consumer key, regardless of HID usage type.

Really, I think this should be a `static constexpr` function inside the `Key` class: `Key::consumer(uint16_t code)`, rather than a macro, but that requires more extensive code reorganization.

src/kaleidoscope/key_defs.h
```diff
@@ -255,14 +241,15 @@ typedef kaleidoscope::Key Key_;
 #define KEY_LEFT_FN2 0xff
 #define Key_LFN2 Key(KEY_LEFT_FN2, KEY_FLAGS)
 
+#define SYSTEM_KEY(code) Key(code, SYNTHETIC | IS_SYSCTL)
```
gedankenexperimenter, line 244: To simplify the definition of System Control keys, I added this macro as the counterpart to the existing `CONSUMER_KEY()` macro below.

src/kaleidoscope/key_defs.h
obra: Now that we're using that magic 0x03FF constant more than once, It should get defined as a named constant. Really it should have been one from the beginning.

obra:
I'm going to start to break this apart. The data update to the HID table is a no-brainer, but mixing in functionality changes into the same PR makes me a little uneasy.  (I'd probably have done the changes to the Consumer and System key macros each as its own PR)

obra:
All the changes to HIDTables.h are now on master as 9404dcc048401b315bca19cb676d499f8d8199d3.

src/kaleidoscope/key_defs.h
gedankenexperimenter: Excellent point. I'll get on that.

gedankenexperimenter:
> I'm going to start to break this apart. The data update to the HID table is a no-brainer, but mixing in functionality changes into the same PR makes me a little uneasy. (I'd probably have done the changes to the Consumer and System key macros each as its own PR)

I did try to keep the commits separate to make it easy to reconfigure things, since I figured you might want adjustments made.

src/kaleidoscope/key_defs.h
obra: That indeed sounds like a bug, but one to be fixed, rather than just throwing away the key type data.

src/kaleidoscope/key_defs_sysctl.h
```diff
@@ -18,41 +18,50 @@
...
-#define System_PowerDown Key(HID_SYSTEM_POWER_DOWN, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_Sleep Key(HID_SYSTEM_SLEEP, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_WakeUp Key(HID_SYSTEM_WAKE_UP, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_ContextMenu Key(HID_SYSTEM_CONTEXT_MENU, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_MainMenu Key(HID_SYSTEM_MAIN_MENU, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_AppMenu Key(HID_SYSTEM_APP_MENU, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_MenuHelp Key(HID_SYSTEM_MENU_HELP, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_MenuExit Key(HID_SYSTEM_MENU_EXIT, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_MenuSelect Key(HID_SYSTEM_MENU_SELECT, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_OSC)
-#define System_MenuRight Key(HID_SYSTEM_MENU_RIGHT, KEY_FLAGS | SYNTHETIC|IS_SYSCTL | HID_TYPE_RTC)
```
obra, line 30: Not all of the System control keys -are- OSC keys.  Even if we're not yet using this data, I really don't want to just discard it from the tables.

gedankenexperimenter: The problem is that the `HID_TYPE_*` bits overlap with other bits that are in use in the `flags` byte of a `Key` object. If you include `HID_TYPE_OOC`, you're also including `IS_CONSUMER` along with it. `handleSyntheticKeyswitchEvent()` could be changed to check for System keys before Consumer keys to solve that problem, but Consumer keys can't use those `HID_TYPE_*` constants.

src/kaleidoscope/key_defs.h
gedankenexperimenter: I don't think there are enough bits to spare.

src/kaleidoscope/key_defs_sysctl.h
obra: I want @algernon's input before I sign off on it, but I think I'd be ok with a solution where, for now, we define all the HID_TYPE_XX constants to 0, rather than just removing them from the key definitions.

gedankenexperimenter: Correction: we can't solve it by changing `handleSyntheticKeyswitchEvent()`, because Consumer keys might have the `IS_SYSCTL` bit set, depending on the value.

gedankenexperimenter: I think we could call out four HID types (using the two bits above `IS_CONSUMER` and below `SYNTHETIC`). Or a default, and three that get special handling.

obra: I'd rather defer the exact bit math until we understand our needs.

gedankenexperimenter:
@obra – At this point, would you like me to rebase this PR onto the master branch?

obra:
Sure. Or break it into a pair of PRs to handle consumer and system keys

gedankenexperimenter:
I've done a rebase in my working repository. I'm checking it now to make sure I didn't screw anything up in the process, then I'll force-push the new version.

src/kaleidoscope/key_defs.h
```diff
@@ -255,14 +260,19 @@ typedef kaleidoscope::Key Key_;
...
+  Key(code, SYNTHETIC | IS_SYSCTL | (hid_type & HID_TYPE_MASK))
 
 /* Most Consumer keys are more then 8bit, the highest Consumer hid code
    uses 10bit. By using the 11bit as flag to indicate a consumer keys was activate we can
    use the 10 lsb as the HID Consumer code. If you need to get the keycode of a Consumer key
    use the CONSUMER(key) macro this will return the 10bit keycode.
 */
-#define CONSUMER(key) (key.getRaw() & 0x03FF)
-#define CONSUMER_KEY(code, flags) Key((code) | ((flags | SYNTHETIC|IS_CONSUMER) << 8))
+constexpr uint16_t CONSUMER_KEYCODE_MASK = 0x03FF;
```
gedankenexperimenter, line 271: Magic number replaced with defined constant. The official code style guide says it should be named `kConsumerKeycodeMask`, but that style isn't used anyplace else in Kaleidoscope that I can find, so I'm leaving it in all caps for the moment.

src/kaleidoscope/key_defs.h
```diff
@@ -225,19 +225,24 @@ typedef kaleidoscope::Key Key_;
...
+#define HID_TYPE_LC    B00000000
+#define HID_TYPE_MC    B00000000
+#define HID_TYPE_NARY  B00000000
+#define HID_TYPE_OOC   B00000000
+#define HID_TYPE_OSC   B00000000
+#define HID_TYPE_RTC   B00000000
+#define HID_TYPE_SEL   B00000000
+#define HID_TYPE_SV    B00000000
+// Mask defining the allowed usage type flag bits:
+#define HID_TYPE_MASK  B00110000
```
gedankenexperimenter, line 245: These are the only two non-conflicting bits available. Enough for a base `00` default, and three "special" usage types (`01`, `10`, `11`), to be defined later.

src/kaleidoscope/key_defs.h
```diff
@@ -255,14 +260,19 @@ typedef kaleidoscope::Key Key_;
 #define KEY_LEFT_FN2 0xff
 #define Key_LFN2 Key(KEY_LEFT_FN2, KEY_FLAGS)
 
+#define SYSTEM_KEY(code, hid_type) \
+  Key(code, SYNTHETIC | IS_SYSCTL | (hid_type & HID_TYPE_MASK))
```
gedankenexperimenter, line 264: I've added back in the HID Usage Type parameter. I'm using the `HID_TYPE_MASK` to ensure that only the allowed bits can be set in the `flags` byte.

src/kaleidoscope/key_defs.h
```diff
@@ -255,14 +260,19 @@ typedef kaleidoscope::Key Key_;
...
    use the 10 lsb as the HID Consumer code. If you need to get the keycode of a Consumer key
    use the CONSUMER(key) macro this will return the 10bit keycode.
 */
-#define CONSUMER(key) (key.getRaw() & 0x03FF)
-#define CONSUMER_KEY(code, flags) Key((code) | ((flags | SYNTHETIC|IS_CONSUMER) << 8))
+constexpr uint16_t CONSUMER_KEYCODE_MASK = 0x03FF;
+#define CONSUMER(key) (key.getRaw() & CONSUMER_KEYCODE_MASK)
+#define CONSUMER_KEY(code, hid_type) \
+  Key((code & CONSUMER_KEYCODE_MASK) | \
+      ((SYNTHETIC | IS_CONSUMER | (hid_type & HID_TYPE_MASK)) << 8))
```
gedankenexperimenter, line 275: HID Usage Type parameter also added here.

obra:
I've just spent a couple hours working through this PR, rewriting the history to contain separable logical changes, since the flow got a little hard to follow with the rework.

Closing in favor of #915

Issue #896, opened 2020-09-08
Double-tapping a OneShot layer key doesn't always make it sticky


gedankenexperimenter:
If a OneShot layer key activates a higher layer that has a non-transparent key mapped on the OneShot key, it will stay active only until the next keypress. That works just fine, as long as the next key pressed is anything other than that OneShot key. Because Kaleidoscope does a lookup in the keymap when a key toggles on, it won't find the OneShot layer key, so it's impossible to double-tap such a key to make it sticky.

I have a solution for this problem, but it involves a major change to the interface between the event handler and the keymap, and might require changes to several plugins to adapt to it. I'll submit another issue and/or a draft PR to discuss it further. I don't think it would be worth it just to solve this rather questionable "problem", but I think it has the potential to simplify multiple keystroke-handling plugins.

PR #904, opened 2020-09-15
Add a full keyboard state array

gedankenexperimenter:
## Summary

This PR introduces a new data structure that represents the full current state of the keyboard. It is implemented as an array of `Key` objects, indexed by `KeyAddr`, one per physical keyswitch on the keyboard. For each key position, the value will either be `Key_Transparent` for a key that is "off" or "idle", or it will contain the current `Key` value that the key is set to, most likely looked up from the keymap, but possibly overridden by a plugin.

Without any active plugins, the keyboard state array will mirror the state of the physical keyswitches. When a keyswitch toggles on, its entry in the array will take on that key's value from the keymap. When it toggles off, it will get the value `Key_Transparent`.

Active plugins will be able to use this structure to make the _logical_ representation of the keyboard state differ from the _physical_ state of the keyswitches. In particular, this will make it easier for a plugin to delay the activation of a keyswitch that toggles on, and to keep a key active after its keyswitch toggles off.

## Event-driven firmware (phase one)

This is the first step toward a version of Kaleidoscope that is truly event-driven, and I think that it is important to understand it in that context. Ultimately, the plan is to make extensive changes to Kaleidoscope's main loop that will improve performance and both simplicity and reliability of plugins.

A full keyswitch scan will be done each cycle, as before, but rather than calling the full set of plugin "event" handlers for each active keyswitch in sequence and building a new HID report along the way, then calling the pre-report handlers and sending the resulting report, it will respond to each keyswitch state change. Each time one toggles on or off, it will call plugin event handlers for that event. If a plugin needs information about other keys that are active, it can then consult the keyboard state array, which—unlike the HID report—is always complete and accurate. Once the event handlers are done, a new HID report will be constructed from the information in the keyboard state array (newly updated as a result of the event processing), and then the pre-report handlers called, and the report sent.

At that point, the scan will continue, so that if a second event is detected in the same cycle, it will be handled independently, and a second report sent without the need for plugins to be concerned about any corner case bugs that could be triggered by "simultaneous" keyswitch events.

This new main loop would obviate the need for re-processing keyswitches that are still held every cycle, and result in a substantial reduction in unnecessary rebuilding of the same report, only to have it suppressed by the HID module after all the work is done.

It will also make it much clearer which handler function should be used for any given task that a plugin needs to accomplish. Right now, there are quite a few plugins that use the `beforeReportingState()` handler in order to do things like check timeouts, because this function gets called once per cycle. If it gets changed to once per HID report (attempt), it's more obviously decoupled from the scan cycle, as it should be.

## A HID report is inadequate for recording state

Until this PR, Kaleidoscope has used the keyboard HID report as a mechanism for recording the state of the keyboard, and this has been problematic at best. First of all, though that report does accurately reflect the state of a non-programmable keyboard (with very minor exceptions), where each key sends one and only one unique keycode, it is inadequate to record the full state of a Kaleidoscope-powered keyboard, even if we disregard layer keys and other special `Key` values not used directly in the keyboard HID report. Keys with modifier flags complicate things quite a bit, as do duplicate keys (modifiers are probably most common, but not the only one: Redial is the most likely example I can think of, but there may well be keymaps out there with multiple <kbd>space</kbd> or <kbd>E</kbd> keys.) This is, of course, why the HID report gets cleared and re-constructed every cycle.

With a full keyboard state array, it becomes much easier to track the information that a plugin needs. For example, if the left shift keycode is active in the HID report (and a plugin would need to check the _old_ HID report, of course), it's not possible for a plugin to know from that whether that `shift` came from a key with a modifier flag (e.g. `LSHIFT(Key_X)`) or from a modifier-only key (e.g. `Key_LeftShift` or `Key_Meh`) — or both. The keyboard state array makes this straightforward to check, simplifying plugin design, and reducing the likelihood of bugs.

## The old keymap cache

The existing live composite keymap is a cache of values from the keymap, so that a full lookup down the layer stack isn't necessary for every active keyswitch in every cycle. Since the keyboard state array must be checked before the keymap, in case a plugin has altered the value of an active key, the live composite keymap becomes redundant in the vast majority of calls, and would only save lookups on key toggle on events for keys that have already been pressed once since the last layer change. These are rare enough that it not longer makes sense to keep this cache.

[Note: When I originally created this PR (and, indeed, when I first got the idea during the design stage for writing Kaleidoglyph), I had realized that it would make sense to convert the live composite keymap into a representation of the keyboard state, and so I thought about it as a re-purposing of an existing structure. I think this has caused a great deal of confusion, so I'm trying to avoid that trap again. Both the keymap cache and the keyboard state array have the same basic structure, and even similar contents and update times and mechanisms, but it's vital that we not think of them as being that closely related. The keyboard state array is _not_ a keymap cache.]

---

### Original PR description

~~This converts `Layer.live_composite_keymap_[]` from a simple cache to a representation of the keyboard's current state. Barring plugin activity, an idle key will have the value `Key_Transparent`, and a pressed key will have the value of whatever key it's currently mapped to in the keymap. A value of `Key_NoKey` will mask that key until it is released.~~

~~This change will allow plugins to deliberately alter the keymap cache by calling either `Layer.updateLiveCompositeKeymap()` directly or by calling `handleKeyswitchEvent()` with a specified `Key` value other than `Key_NoKey` or `Key_Transparent`. It also means that a plugin can abort the clearing of a cache entry for a released key by returning `EVENT_CONSUMED`. This will allow OneShot layer shift keys to become sticky even if there is another key mapped to that address on the target layer, for example.~~

~~I'm submitting this as a separate PR from the rewrite I just did of OneShot in order to call more attention to this change, which might impact other keystroke-handling plugins (other than OneShot, Escape-OneShot, Qukeys, and ActiveModColor). TapDance and Leader come to mind as plugins that would need some checking, and there may be others that I'm even less familiar with. The main thing to look out for is places where `EVENT_CONSUMED` is returned for keys that toggled off, and calls to `handleKeyswitchEvent()` that supply both (non-blank) `Key` and (valid) `KeyAddr` arguments.~~

### DIFF
--- a/src/kaleidoscope/event_handler_result.h
+++ b/src/kaleidoscope/event_handler_result.h
@@ -18,9 +18,33 @@
 
 namespace kaleidoscope {
 
+// This is the set of return values for event handlers. Event handlers for
+// plugins are called in sequence by the corresponding hook function, in plugin
+// initialization order. The interpretation of these return values can vary
+// based on the needs of the hook function, but should be as follows:
+//
+// - OK: Continue processing the event. The calling hook function should
+//       continue calling next event handler in the sequence. If all event
+//       handlers return `OK`, finish processing the event.
+//
+// - EVENT_CONSUMED: Stop processing event handlers. The calling hook function
+//       should not call any further handlers, but may continue to take some
+//       actions to finish processing the event. This should be used to indicate
+//       that the event has been successfully handled.
+//
+// - ABORT: Ignore the event. The calling hook function should not call any
+//       further handlers, and should treat the event as if it didn't
+//       happen. This should be used by plugin handlers that need to either
+//       suppress an event or queue the event in order to delay it.
+//
+// - ERROR: Undefined error. The calling hook function should not call any
+//       further handlers. There is currently no specification for what should
+//       happen if this is returned.
+
 enum class EventHandlerResult {
   OK,
   EVENT_CONSUMED,
+  ABORT,
   ERROR,
 };
 
--- a/src/kaleidoscope/layers.cpp
+++ b/src/kaleidoscope/layers.cpp
@@ -1,5 +1,5 @@
 /* Kaleidoscope - Firmware for computer input devices
- * Copyright (C) 2013-2018  Keyboard.io, Inc.
+ * Copyright (C) 2013-2021  Keyboard.io, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
@@ -14,7 +14,8 @@
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "kaleidoscope/Runtime.h"
+#include "kaleidoscope_internal/device.h"
+#include "kaleidoscope/hooks.h"
 #include "kaleidoscope/layers.h"
 #include "kaleidoscope/keyswitch_state.h"
 
@@ -42,19 +43,15 @@ uint32_t Layer_::layer_state_;
 uint8_t Layer_::active_layer_count_ = 1;
 int8_t Layer_::active_layers_[31];
 
-Key Layer_::live_composite_keymap_[Runtime.device().numKeys()];
-uint8_t Layer_::active_layer_keymap_[Runtime.device().numKeys()];
+uint8_t Layer_::active_layer_keymap_[kaleidoscope_internal::device.numKeys()];
 Layer_::GetKeyFunction Layer_::getKey = &Layer_::getKeyFromPROGMEM;
 
 void Layer_::setup() {
   // Explicitly set layer 0's state to 1
   bitSet(layer_state_, 0);
 
-  // Update the keymap cache, so we start with a non-empty state.
+  // Update the active layer cache (every entry will be `0` to start)
   Layer.updateActiveLayers();
-  for (auto key_addr : KeyAddr::all()) {
-    Layer.updateLiveCompositeKeymap(key_addr);
-  }
 }
 
 void Layer_::handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState) {
@@ -113,24 +110,29 @@ void Layer_::handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState) {
 }
 
 Key Layer_::eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState) {
-  if (mappedKey.getFlags() != (SYNTHETIC | SWITCH_TO_KEYMAP))
-    return mappedKey;
-
-  handleKeymapKeyswitchEvent(mappedKey, keyState);
-  return Key_NoKey;
+  if (mappedKey.getFlags() == (SYNTHETIC | SWITCH_TO_KEYMAP))
+    handleKeymapKeyswitchEvent(mappedKey, keyState);
+  return mappedKey;
 }
 
 Key Layer_::getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr) {
   return keyFromKeymap(layer, key_addr);
 }
 
+// Deprecated
 void Layer_::updateLiveCompositeKeymap(KeyAddr key_addr) {
-  int8_t layer = active_layer_keymap_[key_addr.toInt()];
-  live_composite_keymap_[key_addr.toInt()] = (*getKey)(layer, key_addr);
+  // We could update the active keys cache here (as commented below), but I
+  // think that's unlikely to serve whatever purpose the caller had in
+  // mind. `Layer.lookup()` will still give the correct result, and without a
+  // `Key` value is specified, this function no longer serves a purpose.
+
+  // #include "kaleidoscope/LiveKeys.h"
+  // int8_t layer = active_layer_keymap_[key_addr.toInt()];
+  // live_keys.activate(key_addr, (*getKey)(layer, key_addr));
 }
 
 void Layer_::updateActiveLayers(void) {
-  memset(active_layer_keymap_, 0, Runtime.device().numKeys());
+  memset(active_layer_keymap_, 0, kaleidoscope_internal::device.numKeys());
   for (auto key_addr : KeyAddr::all()) {
     int8_t layer_index = active_layer_count_;
     while (layer_index > 0) {
--- a/src/kaleidoscope/layers.h
+++ b/src/kaleidoscope/layers.h
@@ -111,6 +111,9 @@ class Layer_ {
   }
   static boolean isActive(uint8_t layer);
 
+  static void handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState);
+
+  DEPRECATED(LAYER_EVENTHANDLER)
   static Key eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState);
 
   typedef Key(*GetKeyFunction)(uint8_t layer, KeyAddr key_addr);
@@ -137,8 +140,6 @@ class Layer_ {
   static uint8_t active_layer_count_;
   static int8_t active_layers_[31];
   static uint8_t active_layer_keymap_[kaleidoscope_internal::device.numKeys()];
-
-  static void handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState);
 };
 }
 
--- a/src/kaleidoscope/layers.h
+++ b/src/kaleidoscope/layers.h
@@ -1,5 +1,5 @@
 /* Kaleidoscope - Firmware for computer input devices
- * Copyright (C) 2013-2018  Keyboard.io, Inc.
+ * Copyright (C) 2013-2021  Keyboard.io, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
@@ -20,6 +20,7 @@
 #include "kaleidoscope/key_defs.h"
 #include "kaleidoscope/keymaps.h"
 #include "kaleidoscope/device/device.h"
+#include "kaleidoscope/LiveKeys.h"
 #include "kaleidoscope_internal/device.h"
 #include "kaleidoscope_internal/sketch_exploration/sketch_exploration.h"
 #include "kaleidoscope_internal/shortname.h"
@@ -83,7 +84,13 @@ class Layer_ {
    * `lookupOnActiveLayer`.
    */
   static Key lookup(KeyAddr key_addr) {
-    return live_composite_keymap_[key_addr.toInt()];
+    // First check the keyboard state array
+    Key key = live_keys[key_addr];
+    // If that entry is clear, look up the entry from the active keymap layers
+    if (key == Key_Transparent) {
+      key = lookupOnActiveLayer(key_addr);
+    }
+    return key;
   }
   static Key lookupOnActiveLayer(KeyAddr key_addr) {
     uint8_t layer = active_layer_keymap_[key_addr.toInt()];
@@ -111,9 +118,11 @@ class Layer_ {
 
   static Key getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr);
 
+  DEPRECATED(LAYER_UPDATELIVECOMPOSITEKEYMAP)
   static void updateLiveCompositeKeymap(KeyAddr key_addr, Key mappedKey) {
-    live_composite_keymap_[key_addr.toInt()] = mappedKey;
+    live_keys.activate(key_addr, mappedKey);
   }
+  DEPRECATED(LAYER_UPDATELIVECOMPOSITEKEYMAP)
   static void updateLiveCompositeKeymap(KeyAddr key_addr);
   static void updateActiveLayers(void);
 
@@ -127,7 +136,6 @@ class Layer_ {
   static uint32_t layer_state_;
   static uint8_t active_layer_count_;
   static int8_t active_layers_[31];
-  static Key live_composite_keymap_[kaleidoscope_internal::device.numKeys()];
   static uint8_t active_layer_keymap_[kaleidoscope_internal::device.numKeys()];
 
   static void handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState);
--- a/src/kaleidoscope_internal/deprecations.h
+++ b/src/kaleidoscope_internal/deprecations.h
@@ -27,3 +27,10 @@
 
 /* Messages */
 
+#define _DEPRECATED_MESSAGE_LAYER_UPDATELIVECOMPOSITEKEYMAP             __NL__ \
+  "`Layer.updateLiveCompositeKeymap()` is deprecated.\n"                __NL__ \
+  "The 'live composite keymap' cache has been replaced with the\n"      __NL__ \
+  "'active keys' cache, which now represents the state of the active\n" __NL__ \
+  "keys at any given time. It is probably not necessary to directly\n"  __NL__ \
+  "update that cache from a plugin, but if you need to, please use\n"   __NL__ \
+  "the `live_keys.activate(key_addr, key)` function instead."
--- a/src/kaleidoscope_internal/deprecations.h
+++ b/src/kaleidoscope_internal/deprecations.h
@@ -34,3 +34,7 @@
   "keys at any given time. It is probably not necessary to directly\n"  __NL__ \
   "update that cache from a plugin, but if you need to, please use\n"   __NL__ \
   "the `live_keys.activate(key_addr, key)` function instead."
+
+#define _DEPRECATED_MESSAGE_LAYER_EVENTHANDLER               __NL__ \
+  "`Layer.eventHandler()` is deprecated.\n"                  __NL__ \
+  "Please use `Layer.handleKeymapKeyswitchEvent()` instead."
--- a/src/kaleidoscope_internal/event_dispatch.h
+++ b/src/kaleidoscope_internal/event_dispatch.h
@@ -80,7 +80,7 @@
 
 #define _REGISTER_EVENT_HANDLER(                                                 \
     HOOK_NAME, HOOK_VERSION, DEPRECATION_TAG,                                    \
-    SHOULD_ABORT_ON_CONSUMED_EVENT,                                              \
+    SHOULD_EXIT_IF_RESULT_NOT_OK,                                                \
     TMPL_PARAM_TYPE_LIST, TMPL_PARAM_LIST, TMPL_DUMMY_ARGS_LIST,                 \
     SIGNATURE, ARGS_LIST)                                                 __NL__ \
                                                                           __NL__ \
@@ -116,8 +116,8 @@
    MAKE_TEMPLATE_SIGNATURE(UNWRAP TMPL_PARAM_TYPE_LIST)                   __NL__ \
    struct _NAME4(EventHandler_, HOOK_NAME, _v, HOOK_VERSION) {            __NL__ \
                                                                           __NL__ \
-      static bool shouldAbortOnConsumedEvent() {                          __NL__ \
-        return SHOULD_ABORT_ON_CONSUMED_EVENT;                            __NL__ \
+      static bool shouldExitIfResultNotOk() {                             __NL__ \
+        return SHOULD_EXIT_IF_RESULT_NOT_OK;                              __NL__ \
       }                                                                   __NL__ \
                                                                           __NL__ \
       template<typename Plugin__,                                         __NL__ \
@@ -166,8 +166,8 @@
                                                                      __NL__ \
    result = EventHandler__::call(PLUGIN, hook_args...);              __NL__ \
                                                                      __NL__ \
-   if (EventHandler__::shouldAbortOnConsumedEvent() &&               __NL__ \
-       result == kaleidoscope::EventHandlerResult::EVENT_CONSUMED) { __NL__ \
+   if (EventHandler__::shouldExitIfResultNotOk() &&                  __NL__ \
+       result != kaleidoscope::EventHandlerResult::OK) {             __NL__ \
       return result;                                                 __NL__ \
    }                                                                 __NL__
 
--- a/docs/UPGRADING.md
+++ b/docs/UPGRADING.md
@@ -34,6 +34,69 @@ any API we've included in a release. Typically, this means that any code that us
 
 ## New features
 
+### Keyboard State array
+
+The keymap cache (`Layer_::live_composite_keymap_[]`) has been replaced by a keyboard state array (`kaleidoscope::live_keys[]`). The top-level functions that handle keyswitch events have been updated to treat this new array as a representation of the current state of the keyboard, with corresponding `Key` values for any keys that are active (physically held or activated by a plugin).
+
+#### For end-users
+
+There should be no user-visible changes for anyone who simply uses core plugins. A few functions have been deprecated (`Layer.eventHandler()` & `Layer.updateLiveCompositeKeymap()`), but there are straightforward replacements for both.
+
+#### For developers
+
+The major changes are to the `handleKeyswitchEvent()` function, which has been reorganized in order to update the new keyboard state array with correct values at the appropriate times. In addition to that, two new facilities are available:
+
+##### `EventHandlerResult::ABORT`
+
+This is a new return value available to plugin event handlers, which is similar to `EVENT_CONSUMED` in that it causes the calling hook function to return early (stopping any subsequent handlers from seeing the event), but is treated differently by `handleKeyswitchEvent()`. If a handler returns `EVENT_CONSUMED`, the keyboard state array will still be updated by `handleKeyswitchEvent()`, but if it returns `ABORT`, it will not. In both cases, no further event processing will be done by the built-in event handler.
+
+##### `live_keys[key_addr]`
+
+This is the new facility for checking the value of an entry in the keyboard state array. It is indexed directly by `KeyAddr` values, without the need to convert them to integers first. For example, it could be used in a range-based `for` loop to check for values of interest:
+
+```c++
+for (KeyAddr key_addr : KeyAddr::all()) {
+  Key key = live_keys[key_addr];
+  if (key == Key_LeftShift || key == Key_RightShift) {
+    // do something special...
+  }
+}
+```
+
+Additionally, if the `KeyAddr` values are not needed, one can use the iterator from the new `KeyMap` class like so:
+
+```c++
+for (Key key : live_keys.all()) {
+  if (key == Key_X) {
+    // do something special...
+  }
+}
+```
+
+The `live_keys` object's subscript operator can also be used to set values in the keyboard state array:
+
+```c++
+live_keys[key_addr] = Key_X;
+```
+
+It also comes with several convenience functions which can be used to make the intention of the code clear:
+
+```c++
+// Set a value in the keyboard state array to a specified Key value:
+live_keys.activate(key_addr, Key_X);
+
+// Set a value to Key_Transparent, deactivating the key:
+live_keys.clear(key_addr);
+
+// Set all values in the array to Key_Transparent:
+live_keys.clear();)
+
+// Set a value to Key_NoKey, masking the key until its next release event:
+live_keys.mask(key_addr);
+```
+
+In most cases, it won't be necessary for plugins or user sketches to call any of these functions directly, as the built-in event handler functions will manage the keyboard state array automatically.
+
 ### New build system
 
 In this release, we replace kaleidoscope-builder with a new Makefile based build system that uses `arduino-cli` instead of of the full Arduino IDE. This means that you can now check out development copies of Kaliedoscope into any directory, using the `KALEIDOSCOPE_DIR` environment variable to point to your installation.
@@ -578,6 +641,14 @@ The following headers and names have changed:
 - [SpaceCadet](plugins/Kaleidoscope-SpaceCadet.md) had the `kaleidoscope::SpaceCadet::KeyBinding` type replaced by `kaleidoscope::plugin::SpaceCadet::KeyBinding`.
 - [Syster](plugins/Kaleidoscope-Syster.md) had the `kaleidoscope::Syster::action_t` type replaced by `kaleidoscope::plugin::Syster::action_t`.
 - [TapDance](plugins/Kaleidoscope-TapDance.md) had the `kaleidoscope::TapDance::ActionType` type replaced by `kaleidoscope::plugin::TapDance::ActionType`.
+
+### Live Composite Keymap Cache
+
+The live composite keymap, which contained a lazily-updated version of the current keymap, has been replaced. The `Layer.updateLiveCompositeKeymap()` functions have been deprecated, and depending on the purpose of the caller, it might be appropriate to use `Runtime.updateActiveKey()` instead.
+
+When `handleKeyswitchEvent()` is looking up a `Key` value for an event, it first checks the value in the active keys cache before calling `Layer.lookup()` to get the value from the keymap. In the vast majority of cases, it won't be necessary to call `Runtime.updateActiveKey()` manually, however, because simply changing the value of the `Key` parameter of an `onKeyswitchEvent()` handler will have the same effect.
+
+Second, the `Layer.eventHandler()` function has been deprecated. There wasn't much need for this to be available to plugins, and it's possible to call `Layer.handleKeymapKeyswitchEvent()` directly instead.
 
 # Removed APIs
 
--- a/docs/codebase/glossary.md
+++ b/docs/codebase/glossary.md
@@ -17,7 +17,7 @@ A single physical input, such as a keyswitch or other input like a knob or a sli
 
 ### Key number
 
-An integer representing a Keyswitch’s position in the “Physical Layout”
+An integer representing a Keyswitch’s position in the “Physical Layout”. Represented in the code by the `KeyAddr` type.
 
 ### Physical Layout
 
@@ -33,7 +33,7 @@ A representation of a specific behavior. Most often a representation of a specif
 
 ### Keymap
 
-A list of key bindings for all keyswitchess on the Physical Layout
+A list of key bindings for all keyswitchess on the Physical Layout. Represented in the code by the `KeyMap` type.
 
 ### Keymaps
 
@@ -47,10 +47,9 @@ An entry in that ordered list of keymaps. Each layer has a unique id number that
 
 An ordered list of all the currently-active layers, in the order they should be evaluated when figuring out what a key does.
 
-### Override Layer
+### Live keys
 
-A special layer that’s always active and evaluated before checking keys in the “Active layer stack”  
-
+A representation of the current state of the keyboard's keys, where non-transparent entries indicate keys that are active (logically—usually, but not necessarily, physically held). Represented in the code by the `LiveKeys` type (and the `live_keys` object).
 
 
 ## Keyswitch state
--- a/plugins/Kaleidoscope-Escape-OneShot/src/kaleidoscope/plugin/Escape-OneShot.cpp
+++ b/plugins/Kaleidoscope-Escape-OneShot/src/kaleidoscope/plugin/Escape-OneShot.cpp
@@ -23,21 +23,19 @@
 namespace kaleidoscope {
 namespace plugin {
 
-bool EscapeOneShot::did_escape_;
-
 EventHandlerResult EscapeOneShot::onKeyswitchEvent(Key &mapped_key, KeyAddr key_addr, uint8_t keyState) {
   if (mapped_key != Key_Escape || (keyState & INJECTED))
     return EventHandlerResult::OK;
 
-  if (did_escape_)
-    mapped_key = Key_NoKey;
-  did_escape_ = !keyToggledOff(keyState);
+  if (!keyToggledOn(keyState))
+    return EventHandlerResult::OK;
 
   if ((!::OneShot.isActive() || ::OneShot.isPressed()) && !::OneShot.isSticky()) {
     return EventHandlerResult::OK;
   }
 
   ::OneShot.cancel(true);
+  mapped_key = Key_NoKey;
   return EventHandlerResult::EVENT_CONSUMED;
 }
 
--- a/plugins/Kaleidoscope-Escape-OneShot/src/kaleidoscope/plugin/Escape-OneShot.h
+++ b/plugins/Kaleidoscope-Escape-OneShot/src/kaleidoscope/plugin/Escape-OneShot.h
@@ -27,8 +27,6 @@ class EscapeOneShot : public kaleidoscope::Plugin {
 
   EventHandlerResult onKeyswitchEvent(Key &mapped_key, KeyAddr key_addr, uint8_t keyState);
 
- private:
-  static bool did_escape_;
 };
 }
 }
--- a/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.cpp
+++ b/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.cpp
@@ -67,7 +67,7 @@ EventHandlerResult Qukeys::onKeyswitchEvent(Key& key, KeyAddr k, uint8_t key_sta
     }
     // Any event that gets added to the queue gets re-processed later, so we
     // need to abort processing now.
-    return EventHandlerResult::EVENT_CONSUMED;
+    return EventHandlerResult::ABORT;
   }
 
   // The key is being held. We need to determine if we should block it because
@@ -84,7 +84,7 @@ EventHandlerResult Qukeys::onKeyswitchEvent(Key& key, KeyAddr k, uint8_t key_sta
       }
       // Otherwise, the first matching event was a key press, so we need to
       // suppress it for now.
-      return EventHandlerResult::EVENT_CONSUMED;
+      return EventHandlerResult::ABORT;
     }
   }
 
--- a/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
+++ b/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
@@ -20,7 +20,7 @@
 
 #include "kaleidoscope/Runtime.h"
 #include <Kaleidoscope-Ranges.h>
-#include "kaleidoscope/plugin/Qukeys/KeyAddrEventQueue.h"
+#include "kaleidoscope/KeyAddrEventQueue.h"
 
 // DualUse Key definitions for Qukeys in the keymap
 #define MT(mod, key) Key(                                               \
--- a/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys/KeyAddrEventQueue.h
+++ b/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys/KeyAddrEventQueue.h
@@ -105,6 +105,19 @@ class KeyAddrEventQueue {
     release_event_bits_ >>= 1;
   }
 
+  void shift(uint8_t n) {
+    if (n >= length_) {
+      clear();
+      return;
+    }
+    length_ -= n;
+    for (uint8_t i{0}; i < length_; ++i) {
+      addrs_[i]      = addrs_[i + n];
+      timestamps_[i] = timestamps_[i + n];
+    }
+    release_event_bits_ >>= n;
+  }
+
   // Empty the queue entirely.
   void clear() {
     length_             = 0;
--- a/plugins/Kaleidoscope-Redial/src/kaleidoscope/plugin/Redial.cpp
+++ b/plugins/Kaleidoscope-Redial/src/kaleidoscope/plugin/Redial.cpp
@@ -21,27 +21,16 @@
 namespace kaleidoscope {
 namespace plugin {
 
-Key Redial::key_to_redial_;
 Key Redial::last_key_;
-bool Redial::redial_held_ = false;
 
 EventHandlerResult Redial::onKeyswitchEvent(Key &mapped_key, KeyAddr key_addr, uint8_t key_state) {
-  if (mapped_key == Key_Redial) {
-    if (keyToggledOff(key_state))
-      key_to_redial_ = last_key_;
-
-    mapped_key = key_to_redial_;
-    redial_held_ = keyIsPressed(key_state);
-
-    return EventHandlerResult::OK;
+  if (keyToggledOn(key_state)) {
+    if (mapped_key == Key_Redial) {
+      mapped_key = last_key_;
+    } else if (shouldRemember(mapped_key)) {
+      last_key_ = mapped_key;
+    }
   }
-
-  if (keyToggledOn(key_state) && shouldRemember(mapped_key)) {
-    last_key_ = mapped_key;
-    if (!redial_held_)
-      key_to_redial_ = mapped_key;
-  }
-
   return EventHandlerResult::OK;
 }
 
--- a/plugins/Kaleidoscope-Redial/src/kaleidoscope/plugin/Redial.h
+++ b/plugins/Kaleidoscope-Redial/src/kaleidoscope/plugin/Redial.h
@@ -34,9 +34,7 @@ class Redial : public kaleidoscope::Plugin {
   EventHandlerResult onKeyswitchEvent(Key &mapped_key, KeyAddr key_addr, uint8_t key_state);
 
  private:
-  static Key key_to_redial_;
   static Key last_key_;
-  static bool redial_held_;
 };
 
 }
--- a/plugins/Kaleidoscope-ShapeShifter/src/kaleidoscope/plugin/ShapeShifter.cpp
+++ b/plugins/Kaleidoscope-ShapeShifter/src/kaleidoscope/plugin/ShapeShifter.cpp
@@ -16,25 +16,21 @@
  */
 
 #include <Kaleidoscope-ShapeShifter.h>
+#include "kaleidoscope/keyswitch_state.h"
+#include "kaleidoscope/LiveKeys.h"
 
 namespace kaleidoscope {
 namespace plugin {
 
 const ShapeShifter::dictionary_t *ShapeShifter::dictionary = NULL;
-bool ShapeShifter::mod_active_;
-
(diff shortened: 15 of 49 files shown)


gedankenexperimenter:
I will submit the OneShot rewrite PR that's built on this separately, when I get the time.

algernon:
FWIW, this is pretty high in my queue. Next time I'm working closer to Kaleidoscope, this will be on top.

gedankenexperimenter:
I'll go ahead and do a round of rebasing, then. I'm looking forward to discussing it with you, @algernon.

gedankenexperimenter:
It was pointed out on [Discord](https://discord.com/channels/492408953041321984/492907083583455246/766437554110464010) that the MouseKeys plugin consumes events that it shouldn't, for this to work. I'll check for other plugins that might be problematic, or make another adjustment, if possible.

src/kaleidoscope/layers.cpp
```diff
@@ -53,7 +52,7 @@ void Layer_::setup() {
   // Update the keymap cache, so we start with a non-empty state.
```
gedankenexperimenter, line 52: This comment needs to be changed

src/kaleidoscope/layers.cpp
```diff
@@ -112,21 +111,22 @@ void Layer_::handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState) {
...
 
-Key Layer_::eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState) {
+EventHandlerResult Layer_::eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState) {
   if (mappedKey.getFlags() != (SYNTHETIC | SWITCH_TO_KEYMAP))
-    return mappedKey;
+    return EventHandlerResult::OK;
 
   handleKeymapKeyswitchEvent(mappedKey, keyState);
-  return Key_NoKey;
+  return EventHandlerResult::EVENT_CONSUMED;
```
gedankenexperimenter, line 119: This looks suspicious. Is this going to result in layer keys failing to show up in the `active_keys_[]` array? Check `handleKeyswitchEvent()`

src/kaleidoscope/layers.cpp
```diff
@@ -112,21 +111,22 @@ void Layer_::handleKeymapKeyswitchEvent(Key keymapEntry, uint8_t keyState) {
...
 }
 
 Key Layer_::getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr) {
   return keyFromKeymap(layer, key_addr);
 }
 
+// deprecate this function
```
gedankenexperimenter, line 126: I need to add a deprecation warning, and date here.

src/kaleidoscope/layers.h
```diff
@@ -82,8 +84,15 @@ class Layer_ {
    * are curious what the active layer state describes the key as, use
    * `lookupOnActiveLayer`.
    */
+  static Key getStatus(KeyAddr key_addr) {
+    return Runtime.activeKey(key_addr);
+  }
   static Key lookup(KeyAddr key_addr) {
-    return live_composite_keymap_[key_addr.toInt()];
+    Key key = getStatus(key_addr);
```
gedankenexperimenter, line 91: Change this to `Runtime.activeKey(key_addr);`

src/kaleidoscope/layers.h
```diff
@@ -114,15 +123,15 @@ class Layer_ {
...
 
-  static Key eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState);
+  static EventHandlerResult eventHandler(Key mappedKey, KeyAddr key_addr, uint8_t keyState);
 
   typedef Key(*GetKeyFunction)(uint8_t layer, KeyAddr key_addr);
   static GetKeyFunction getKey;
 
   static Key getKeyFromPROGMEM(uint8_t layer, KeyAddr key_addr);
 
   static void updateLiveCompositeKeymap(KeyAddr key_addr, Key mappedKey) {
```
gedankenexperimenter, line 133: Deprecate this function

gedankenexperimenter (commented):
Some notes on things that need updating.

gedankenexperimenter:
So, the description of this PR has gotten a bit out of date since mid-September. I think it's time for an update.

When I first submitted this one, I tried to make the changes as minimal as possible, with the idea that it would be easier to do the code review and verify that it was good. Since then, we've got the testing infrastructure working, and realized that it just doesn't make sense to try to keep the changes as small as possible, because some of them are a bit misleading that way.

## `Layer.live_composite_keymap_[]` _vs_ `Runtime.active_keys_[]`

The original purpose of the `live_composite_keymap_[]` array was to serve as a cache to speed up keymap entry lookups. At the time, this was very important because every single keyswitch would get scanned and the keymap entry looked up in every cycle so that plugin handlers could be called with the `Key` value lookup up. Now, however, we don't bother to call plugin event handlers for idle keys, so most sketches don't have a reason to retrieve the keymap entries for most of the keys in any cycle, so it's not so critical that keymap entry lookups are fast.

[Quite a long time ago, I realized that this keymap cache could be re-purposed and used as a representation of the state of the keyboard (and used it when writing Kaleidoglyph). With two special values, I could represent idle keys (`Key_Transparent`) and masked keys (`Key_NoKey`), and any active keys could have their specific `Key` values in the "cache". My primary purpose in doing it this was was that the Keyboard HID report was inadequate for representing the full state of the keyboard (mainly modifier flags, but some corner cases like two `E` keys are also failure points), but this PR is only the first step to getting us there.]

Here's how it works in this PR: when `handleKeyswitchEvent()` is called with `Key_NoKey`, it looks up the value of the key to pass to event handlers, as before, but now it first checks the `Runtime.active_keys_[]` array, and uses that if the key was not idle (`Key_Transparent`). While a key is being held, it's a tiny bit slower than before because of this check. If the key was idle, then it does a lookup in the keymap via `Layer.lookup()`, which now has more overhead than before, but since it only happens once per keypress (in most cases), that's not a big deal. In fact, that's not very different from what we do now, anyway, instead of seeding the cache completely on every layer change, so the performance hit will be negligible.

One useful way to think of it is that the active keys array is like an extra layer that's always on top of all the active keymap layers.

## `handleKeyswitchEvent()` changes

The big difference here is when the values in the cache get updated. Previously, cache entries would get updated before plugin event handlers were called; now, those cache updates are delayed until after event handlers are done. Those event handlers get the `Key` value passed as a parameter, so they don't need the cache to be updated first. The big difference is that the interpretation of `EVENT_CONSUMED` is different. In the past, it meant "I'm certain I'm done with this key, and I'm sure nobody else needs to do anything about it." That was reasonable enough, but quite limiting, because it could mean that some action was taken, and because the plugin in question took ownership of it, there was nothing further to be done (e.g. MouseKeys handling a mouse movement key), but it could also mean that the plugin specifically needed to stop other plugins from handling that event (e.g. Qukeys delaying a keypress while determining what value to assign to the key).

Now the meaning of `EVENT_CONSUMED` is less ambiguous: it means, "This keypress got aborted or delayed; nobody should do anything about it [yet]". This does mean that certain plugins (e.g. Macros, MouseKeys) get an update so that they return `OK` instead of `EVENT_CONSUMED` when they finish handling a key.

## What are the benefits?

There are a few plugins that already emulate keys being pressed at times when they are not physically held down, and _vice versa_. OneShot and Qukeys both do so already, but it's rather awkward and error-prone. As an example, there's issue #896: a OneShot layer key can't reliably be double-tapped (depending on the keymap) to make it sticky, because the second tap will cause a lookup on the newly-activated layer, so its behaviour will depend on what that keymap entry is. With the `active_keys_[]` version of the cache, OneShot can just put the value `ShiftToLayer(N)` in the cache, and it will persist through the layer change.

### OneShot

OneShot is the most obvious beneficiary of the active keys array cache. In #905, it figures out what the target key is, and changes the `Key` value of the event. When plugin handling is done, it's that value that gets stored in the cache, and used in future cycles until a key release event gets through. After the key toggles on, OneShot just keeps track of the one-shot state of each key (two bits per key is enough), but no longer needs to do any translation, or direct manipulation of the HID report. This also allows it to make arbitrary keys sticky, not just a short list of modifiers and layer-shifts.

And when a OneShot key is released, but its state is such that it should still remain active, OneShot merely needs to return `EVENT_CONSUMED` for that key to stop the release event from marking the key as idle in the active keys cache. [OneShot still needs to actively call `handleKeyswitchEvent()` each cycle to keep it alive in the HID report, but even that won't be necessary if we adopt the next stage of my event-driven Kaleidoscope changes.]

### Other plugins

Qukeys is another plugin that needs to do similar things (both temporarily suppressing handling of keyswitches that are physically pressed, and keeping them active for short periods after they're released). TapDance is another that could benefit from this, and possibly Leader, and perhaps some others that I'm not thinking of.

### Long-term benefits

In the long term, the biggest benefit will be that Kaleidoscope will be storing more state information than it is now, which will simplify certain multiple-overlapping-keys event sequences. The keyboard HID report seems like it would be sufficient, but as we've discovered with modifier flags and rollover, we've got a confusing bit of code that only mostly solves the problem. With the active keys cache (and a different event handling procedure), that complex code can be both simpler and better at handling corner cases.

## What are the risks?

There isn't much danger that doesn't already exist, but in earlier versions of the PR some helpful testers discovered keymap entries getting "stuck". This happened because of a plugin returning `EVENT_CONSUMED` on a key release event, stopping the entry in the cache from getting cleared. Later, when the key was pressed (after a layer change, perhaps), the key would still seem to be mapped to (for example) a MouseKeys key, or a Macros key. The good news is that these errors are very obvious, and easy to fix (just return `OK` instead of `EVENT_CONSUMED` from `onKeyswitchEvent()`).

gedankenexperimenter:
I still want to write a few testcases to demonstrate correctness for this PR, and make another pass through all the included plugins to be certain I didn't miss any cases where `onKeyswitchEvent()` used to return `EVENT_CONSUMED` where it should now return `OK`. When that's done, I'll mark this PR as not a draft any longer.

The other thing I think we should discuss is the names of the two new functions added to `Runtime`: `activeKey(KeyAddr)` & `updateActiveKey(KeyAddr, Key)`. The former is for retrieving a `Key` value from the `active_keys_[]` array, and the latter is for setting a value in that array. Past experience has taught me that @obra and I rarely agree on names, so I try not to put too much effort into them. If `getActiveKey()` & `setActiveKey()` are preferred, that's fine with me, and if some terminology other than "active keys" is preferred to describe the structure, I'm likewise not opposed to changing it.

gedankenexperimenter:
While going through the existing plugins to find places that might need modification, I had an idea that could simplify things. Rather than changing the meaning of `EVENT_CONSUMED`, we could introduce a new `EventHandlerResult` variant, which I would call `ABORT`. To make this work, I would change the test for abortable hook functions so that they abort (stop subsequent handlers from being called) if a handler returns any result other than `OK` (right now, they only abort if a handler returns `EVENT_CONSUMED`). This would let us treat the two values differently, probably not requiring any changes to existing plugins other than the ones that specifically want to stop automatic updating of the active keys array (e.g. Qukeys & OneShot). I think this is probably a much better solution.

gedankenexperimenter:
Right now, there are three `EventHandlerResult` variants defined: `OK`, `EVENT_CONSUMED`, & `ERROR`. The test used in `event_dispatch.h` that determines whether a hook function stops early or continues on to call the next handler is:
```c++
result == kaleidoscope::EventHandlerResult::EVENT_CONSUMED
```
I'm proposing changing it to:
```c++
result != kaleidoscope::EventHandlerResult::OK
```
I believe these are equivalent right now for all core Kaleidoscope plugins, because none of them use `ERROR`. I know almost nothing about third-party plugins right now, but it would be a bit surprising to see one using that variant. @algernon, you probably have a better idea than I do about this; do you know of any cases where `ERROR` is in use?

gedankenexperimenter:
I reorganized the commits again today, removing changes to plugins that are no longer necessary with the introduction of `EventHandlerResult::ABORT`, and separating logical changes from each other as best I can. The only plugin that currently uses `ABORT` is Qukeys, but others will use it eventually (both OneShot and TapDance, at least) if this PR gets merged.

The last thing I think this PR needs is a testcase demonstrating that keymap entries belonging to plugins that return `EVENT_CONSUMED` on key releases don't get stuck. This will require a sketch with two layers, and a Macros key (I think that's easiest) on the second layer, with the same `KeyAddr` as a plain keyboard key on the base layer. It shouldn't take long to write at least one test suite once I get the time for it. With luck, I'll get far enough to mark it ready for review sometime tomorrow.

gedankenexperimenter:
I added one testcase. Rather than using Macros, I decided to write a very simple in-sketch plugin to demonstrate how it works. @obra — I considered re-ordering the commits to put the testcase first, but hesitated because this particular one shouldn't work with current master. It's probably a good idea anyway, but I'm waffling; would you like me to force-push it with the testcase as the first commit?

gedankenexperimenter:
I went ahead and added another testcase, using Macros. It's not as clear a demonstration of the changes and how they work, but it does demonstrate that an existing plugin (in at least this one simple configuration) is unaffected by the active keys change.

src/kaleidoscope_internal/event_dispatch.h
```diff
@@ -167,7 +167,7 @@
    result = EventHandler__::call(PLUGIN, hook_args...);              __NL__ \
                                                                      __NL__ \
    if (EventHandler__::shouldAbortOnConsumedEvent() &&               __NL__ \
```
obra, line 169: Before I read this change, I was wondering if we should instead be creating a method like

```EventHandler__::shouldAbortOnResult(kaleidscope::EventHandlerResult)``` and move the logic in there.



Now that I'm looking, while I understand the intent for the change from == EVENT_CONSUMED to != OK,  it makes the conditional of 

```
if (EventHandler__::shouldAbortOnConsumedEvent() && result != kaleidoscope::EventHandlerResult::OK) {             ```
 read kind of funny, which makes me think that moving the check into a method makes sense.

src/kaleidoscope/event_handler_result.h
```diff
@@ -21,6 +21,7 @@ namespace kaleidoscope {
 enum class EventHandlerResult {
   OK,
   EVENT_CONSUMED,
+  ABORT,
```
obra, line 24: Can you add the doc from that commit message as comments on the ENUM? I want to make sure it's captured somewhere less ephemeral than git history and the changelog.

src/kaleidoscope_internal/deprecations.h
```diff
@@ -46,3 +46,10 @@
   "Layers are now in activation-order, please use" __NL__ \
   "`Layer.forEachActiveLayer()` instead."
 
+#define _DEPRECATED_MESSAGE_LAYER_UPDATELIVECOMPOSITEKEYMAP             __NL__ \
```
obra, line 49: I'm not sure this is a public enough API to -need- the deprecation song and dance. @algernon - what do you think?

src/kaleidoscope_internal/event_dispatch.h
gedankenexperimenter: I'm in complete agreement. This was a case of me trying to make the change as small as possible to make it as easy to verify as correct as possible. I don't have a sufficiently complete understanding of how the plugin hook system works to be confident in changing it, but the "consumed event" in the name definitely starts to become a lie with the change I made here.

Maybe it would be best to do that as a separate PR from this one?

src/kaleidoscope/event_handler_result.h
gedankenexperimenter: Excellent point. I will add those comments when I get a chance. Thank you.

src/kaleidoscope_internal/deprecations.h
gedankenexperimenter: I'd be a bit surprised if `Layer.eventHandler()` gets called from anyplace else. On the other hand, it can be deprecated and removed along with `updateLiveCompositeKeymap()`, so the cost is minimal. If the call were mine to make, I don't know which option I'd pick.

src/kaleidoscope/event_handler_result.h
gedankenexperimenter: I added some comments describing the return values, and their interpretation.

src/kaleidoscope_internal/deprecations.h
gedankenexperimenter: More definitive: `Layer.eventHandler()` does not get called by any existing core plugin. It probably doesn't get called by any third-party plugins, but that's not impossible.

src/kaleidoscope_internal/event_dispatch.h
gedankenexperimenter: I added a small commit changing the name `shouldAbortOnConsumedEvent()` to `shouldExitIfResultNotOk()`. I know that's not really what you had in mind, but it at least matches the new conditional better. I think it might be best to avoid use of the same term ("abort", unless you want to change `EventHandlerResult::ABORT` to something else—`CANCEL`, perhaps? (I still think "abort" is more appropriate)) in both places, and I think "exit" or "return" would be more descriptive of what the hook function will do if it gets a non-`OK` result from a handler.

gedankenexperimenter: I tried to make the change you suggested, but I get an inscrutable error in preprocessor macro expansion:
```
error: unterminated argument list invoking macro "EVAL4"
```
I've scrutinized the three lines of code I changed, and have no idea how to resolve that, so I'm giving up for now.

gedankenexperimenter:
I found another category of bug triggered by this change that I want to do a wider search for. This one came up in Redial. I fixed it, and in the processed simplified Redial and made it a bit easier to follow by taking advantage of the active keys cache.

The problem was that Redial was changing the value of `mapped_key` from `Key_Redial` to another cached value, but expecting it not to persist, so that it would make that change every cycle until it toggled off. On toggle off, it took another action (clearing another state variable). With the active keys cache, as soon as it changes the value of `mapped_key`, it stops seeing `Key_Redial` "events", and never cleared the state variable telling it that the key was held. (This variable was necessary before the active keys cache, because if rollover happened, it was necessary to not change the effective value of the redial key).

So, the thing to look out for is plugins looking for a key toggled off event for a key value that got changed on the same key toggling on. I'll do a search for this, and also add a note about it in UPGRADING.

gedankenexperimenter:
I'm converting this back to a draft; I found a couple of other plugins that probably need minor changes.

gedankenexperimenter:
For the moment, this is rebased on top of #982 (which I opened because I wanted to use those functions in some of the plugin changes I was working on here). I can adapt it without those changes, if necessary, but I'd rather not. I still have a bunch of new testcases to write for this PR, but the TapDance ones are looking good, and that was the most extensive plugin rewrite. (Note: there is still a rollover issue with TapDance keys, but it's a minor one, and was pre-existing on master. I can fix it, but that will be so much easier if Kaleidoscope becomes event-driven that I plan to wait until then.)

obra:
@algernon I'd like to merge this. Any opinions?

gedankenexperimenter:
@obra — Not that I object to merging this PR, obviously, but I was hoping for at least some discussion of the two new method names `Runtime.activeKey(KeyAddr)` and `Runtime.updateActiveKey(KeyAddr, Key)`, which I think are okay, but not obviously the best (perhaps a more obvious alternative would be `getActiveKey()` & `setActiveKey()`). I have only the most tenuous preference for the names I chose, and I'd hate for them to be accepted only for us to determine that the choice was a poor one later on.

obra:
Thanks for flagging that!

obra:
@gedankenexperimenter looking at those names, I do feel like they could be a bit more explicitly named, since they're runtime-global.   The current slate of names suggests to me that there's only one active key.

Maybe  `setActiveKeyAtAddress` and `getActiveKeyAtAddress`?

gedankenexperimenter:
That's unnecessarily wordy, given that both take a `KeyAddr` parameter. To me, it obviously it only applies to the specified entry in the array, and the redundant verbage just makes it harder to focus on the relevant code.

obra:
It's a public method in the global namespace and isn't something that will be called very frequently from most code. If it were an object method on a cache object or something, I'd be totally cool with making it very short.  
From the current naming, it's not clear that it's related to a cache, so `Runtime.activeKey(KeyAddr)` could easily be read as "make this address active" or "is this key pressed?"

gedankenexperimenter:
I'm not sure what you mean by it being "in the global namespace". Can you elaborate? It's a method on the singleton object `kaleidoscope::Runtime`, just like almost all of the plugin objects we have, except that those are one additional level down in the namespace hierarchy.


> From the current naming, it's not clear that it's related to a cache, so `Runtime.activeKey(KeyAddr)` could easily be read as "make this address active" or "is this key pressed?"

Thats a fair reason to reject the names I chose, but not a great one for adding `AtAddress` to the end of `getActiveKey` & `setActiveKey`, which I suggested as alternatives.

As an aside, with this PR, it's gone beyond being a cache, and is now a representation of the keyboard state at any given time, because each entry gets cleared when the key is released (logically, not necessarily physically).

Perhaps you would prefer `getActiveKeyValue(KeyAddr)` or just `getActiveValue(KeyAddr)`? Or `getCurrentKeyValue(KeyAddr)`? [I do think `active` is a bit better than `current` because it indicates that it's not just a keymap cache, but the value the keyboard is currently sending to the host.]

If we really want to get into it, I think the name of the `Key` class is a poor one, because it could be interpreted as so many different things (I think @noseglasses pointed this out a long time ago). It would be much clearer to call it `KeyValue` to disambiguate.

obra:
> If we really want to get into it, I think the name of the `Key` class is a poor one, because it could be interpreted as so many different things (I think @noseglasses pointed this out a long time ago). It would be much clearer to call it `KeyValue` to disambiguate.

At some point, we should absolutely get into the use of 'key' to name things in Kaleidoscope. Too much stuff gets called by the same name. But this issue is the wrong place to chat about it.


> I'm not sure what you mean by it being "in the global namespace". Can you elaborate? It's a method on the singleton object `kaleidoscope::Runtime`, just like almost all of the plugin objects we have, except that those are one additional level down in the namespace hierarchy.
>

Indeed, my wording was technically incorrect.  The kaleidoscope::Runtime singleton object is something we treat as a global namespace. In sketches, it's still known as Kaleidoscope sometimes. Things directly in that namespace can reasonably be presumed to be things that affect the whole system.  That namespace was historically a catch-all, especially with my initial horrible factoring of the system. We've pared a lot of stuff out of it over time, and ideally, I think, it's most useful as a container, with as little code as possible directly in the Runtime.

If you're working through code and you see Runtime.setActiveKey(...), Runtime.activeKey(...), Runtime.updateActiveKey(...), it's natural to presume that they're things that apply to the runtime.  

Looking at it harder, maybe the issue is more that everything got lifted up to be on the Runtime directly?  Can you write a bit about the intent of moving everything directly into Runtime? 

>As an aside, with this PR, it's gone beyond being a cache, and is now a representation of the keyboard state at any given time, because each entry gets cleared when the key is released (logically, not necessarily physically).

mm. That's a very good point.  If it's not a cache, we shouldn't call it a cache.

gedankenexperimenter:
> Looking at it harder, maybe the issue is more that everything got lifted up to be on the Runtime directly? Can you write a bit about the intent of moving everything directly into Runtime?

Sure! The "active keys" structure replaces the "live composite keymap" cache of the `Layer` object (which was, at least originally, purely a cache, and that's where the "cache" term comes into it). The `Runtime` object is the equivalent of what I named the `Controller` in Kaleidoglyph. It's the central, top-level object that drives everything from its `setup()` and `loop()` methods, and seems to me the most logical place to store the global state of the device, which is exactly what the "active keys" array represents. It's state that should be globally accessible to all the plugins, but it's not actually part of the keymap (except that it takes precedence over the keymap), so it doesn't really belong in the `Layer` object.

In short, `Runtime` is the existing object that is, in my opinion, the most appropriate container for this kind of top-level "global" (in the system-wide sense, not the lexical-scoping sense) state. If you don't see it as the appropriate place to store this state, then I suppose we can create a new globally-accessible object for it.

Incidentally, in my (not yet submitted) KeyEvent branch, I moved most of the logic of dealing with events (i.e. the replacement for `handleKeyswitchEvent()`) into `Runtime`, as well. I suppose you're not going to like that, either, though it is even more closely related to the `loop()` function. I guess I don't understand what your concept of the `Runtime` object is. That's not surprising, because I also feel like the name is very odd and unhelpful.

obra:
Runtime was the refactoring of what used to just be "Kaleidoscope", which was very much a catch-all, from before we had much in the way of object orientation or even namespacing.  I'd been writing my C++ like C and that was the result. 

We've been peeling it apart over time.

Your suggestion that this doesn't belong in 'Layer' seems reasonable to me. I think that perhaps the new logical keyboard layer state representation may then merit being an object itself, with an object accessor hanging off Runtime.

obra:
I started working on refactoring this over on the branch ' f/keymap-cache-redesign'

Over there, @gedankenexperimenter commented:

> I feel pretty strongly that the "live keymap" state should not be part of the Layer object (or whatever better name it gets). The former is the current state of the keyboard, the latter is the mapping of keys to values. It would be better to introduce a new, separate object (i.e. ActiveKeymap or LiveKeymap). The "live keymap" can be updated completely independently of the layers in the keymap, but storing this state in the Layer structure strongly indicates that they are more tightly bound than that.

I'm lifting the discussion back here, just so it's all in one place.

We agreed over on Discord that `Layer` is not the right name for this object, although we haven't yet figured out what the right name is.

The object currently called 'Layer' has, up to now, been the source of truth for the answer to the question "what does this key do?"

It should be factored better than it is today, but rather than spinning out this one bit of it now at the same time we are essentially fixing a bug that's blocking end-user functionality, we will be better served by refactoring the whole of 'Layer' systematically.

Popping it back there did also help me realize that we already had a pattern we should be matching for the getter for this functionality--multiple functions that pull values out of particular locations in keymaps named 'lookup', which pretty much settles the accessor naming for now.

gedankenexperimenter:
> Popping it back there did also help me realize that we already had a pattern we should be matching for the getter for this functionality--multiple functions that pull values out of particular locations in keymaps named 'lookup', which pretty much settles the accessor naming for now.

Would you mind clarifying what you mean by this? I'm afraid I don't understand. Meanwhile, I'll try to clarify my reason for wanting the new structure to be separate from the `Layer` object:

> The object currently called 'Layer' has, up to now, been the source of truth for the answer to the question "what does this key do?"

The main question that the current `Layer` object answers is one that I would phrase just a bit more specifically: "What will this key do if it is activated?"

The structure that I'm asking for with this PR answers a different question: "What is this key doing _now_?" From the point of view of a plugin author, that's a very different question, and possibly not even related to the keymap.

For this reason, I think the term "lookup", if used at all, should not be used for the "active keys" structure. That has a strong connotation of it being merely a cache for the keymap, but it's not. For the same reason, I'm souring a bit on the use of the term "keymap" for this structure; it's something that represents the _state_ of the keyboard (though not always the state of the physical key _switches_).

obra:
>> Popping it back there did also help me realize that we already had a pattern we should be matching for the getter for this functionality--multiple functions that pull values out of particular locations in keymaps named 'lookup', which pretty much settles the accessor naming for now.

>Would you mind clarifying what you mean by this? I'm afraid I don't understand. 

Sure. 

This was the pre-existing state of the code:

```
  static Key lookup(KeyAddr key_addr) {
    return live_composite_keymap_[key_addr.toInt()];
  }
  static Key lookupOnActiveLayer(KeyAddr key_addr) {
    uint8_t layer = active_layer_keymap_[key_addr.toInt()];
    return (*getKey)(layer, key_addr);
  }
```

This was how it looked on your branch:
```
  static Key lookup(KeyAddr key_addr) {
    // First check the active keys array
    Key key = Runtime.activeKey(key_addr);
    // If that entry is clear, look up the entry from the active keymap layers
    if (key == Key_Transparent) {
      key = lookupOnActiveLayer(key_addr);
    }
    return key;
  }
```

This is how I have it as of now:

```
  static Key lookup(KeyAddr key_addr) {
    // First check the active keys array
    Key key = lookupOnLiveKeymap(key_addr);
    // If that entry is clear, look up the entry from the active keymap layers
    if (key == Key_Transparent) {
      key = lookupOnActiveLayer(key_addr);
    }
    return key;
  }
```

The thing we're doing is called a lookup here. Hence, we should keep calling a lookup. rather than calling it 'activeKey' or 'liveKey'  - We already have an explicit method called lookupOnActiveLayer and another called just 'lookup'

This datastructure is a keymap.

Reading over the code last night and this morning, we pretty consistently use "keymap" to describe a data structure containing a view of a keyboard with one entry for each key. When users describe multiple keymaps in their sketches, the macros are "KEYMAPS". Individual keymaps get described as "KEYMAP" or "KEYMAP_STACKED". The layer change keys talk about changing between different keymaps.

Based on all of this:

A "keymap" in Kaleidoscope is a datastructure that is a map of the keys on a keyboard, with one value for each key, describing some behavior that should happen when that key is actuated. It is a map of keys. 

A "layer" is one of multiple keymaps inside an ordered datastructure. 

We've never had a coherent name for that ordered datastructure.  It's sometimes called 'layers'. It gets partially defined by the 'KEYMAPS' macro in user sketches, though is also augmented by any keymaps stored in EEPROM.

obra:
Stepping back and going back to the original description of this change, maybe I'm wildly misunderstand what's -in- this datastructure now? (more in a moment)

obra:
(For the future, it's possible to edit the first comment on a PR - If you're changing what the PR does, it's good form to change that, too.)

I've definitely misunderstood the intent of this change, based on a belief that the description still accurately explained the behavior and intent. I'm sorry about that.

Where I first went wrong was that I thought you were changing the behavior and name of the live_composite_keymap, rather than, effectively,  removing that feature and adding a fundamentally different feature in a similar place in the code, because I wasn't working from your different explanation further down in the discussion.

On top of that, I think I misunderstood what you meant by 'active' - The code in layers.cpp uses "active" to describe which layers are currently "turned on," but I think you're using  using it differently.

To date, we haven't used the terms "active" and "idle" keys much in Kaleidoscope and I think this has been central to my  misunderstanding. Inside the core of Kaleidoscope, I see some places in docs where we talk about keys being idle (and even non-idle), but as far as I can tell, we've never used 'active' to talk about keys that are currently "pressed" in the core, other than in the context of values in a HID report.


Am I right that when you write about a key being 'active', what you mean is that the firmware currently believes that the keyswitch at that position has been debounced and is indicating that there's an electrical connection and that when you write about a key being 'idle', that key is -not- indicating an electrical connection? 

You wrote that:

> One useful way to think of it is that the active keys array is like an extra layer that's always on top of all the active keymap layers.

When we were talking about names, I can see how "live" might have felt ok but not stellar to each of us for different reasons. To me, it read as the current/live state of the whole keyboard, while to you, maybe it read as an array of all of the "live"/active/pressed keys? Is that right?

If I've got this right, we're probably in a better place for figuring out the right way forward.

gedankenexperimenter:
I'm glad we're getting closer to mutual understanding on this one.

> (For the future, it's possible to edit the first comment on a PR - If you're changing what the PR does, it's good form to change that, too.)
> 
> I've definitely misunderstood the intent of this change, based on a belief that the description still accurately explained the behavior and intent. I'm sorry about that.

I'm aware that it's possible to change the initial description; sorry for not updating it. I was negligent in not correcting the details that changed from the original version, probably because the concept of what was represented never changed. I'll happily update it when I get a chance.

> On top of that, I think I misunderstood what you meant by 'active' - The code in layers.cpp uses "active" to describe which layers are currently "turned on," but I think you're using using it differently.

I think I'm using it the same way. "Active" == "turned on", whether we're talking about a keymap layer or a key. I'm not sure what distinction you have in mind.

> To date, we haven't used the terms "active" and "idle" keys much in Kaleidoscope and I think this has been central to my misunderstanding. Inside the core of Kaleidoscope, I see some places in docs where we talk about keys being idle (and even non-idle), but as far as I can tell, we've never used 'active' to talk about keys that are currently "pressed" in the core, other than in the context of values in a HID report.

I feel like I've been using the terms "active" and "idle" quite consistently and frequently in conversations about Kaleidoscope in general, and the term "idle" mainly in reference to keys, mainly with @algernon, but possibly with others. This goes way back to conversations stemming from my first working version of Kaleidoglyph, from which we got the proposal to reduce cycle times by not calling event handlers for "idle" keyswitches.

I've been using the "active keys" terminology since I first came up with this design for Kaleidoglyph ~2 years ago, so obviously it seems natural to me now, even if it didn't at first. It's difficult to know when we're making assumptions about "obvious" things that are counterintuitive to others…

> Am I right that when you write about a key being 'active', what you mean is that the firmware currently believes that the keyswitch at that position has been debounced and is indicating that there's an electrical connection and that when you write about a key being 'idle', that key is -not- indicating an electrical connection?

No, not necessarily. In my lexicon, a physical "keyswitch" is either "idle" or "active" (and the transitions between the two states are of particular interest), and the corresponding logical "key" is also either "idle" or "active". Generally, if the physical keyswitch is in one of those states, the logical key will be, too, but that's not necessarily the case. We've got a number of plugins, including two of the most commonly used (OneShot & Qukeys) that disconnect the states of the keyswitch and the key in order to function.

> When we were talking about names, I can see how "live" might have felt ok but not stellar to each of us for different reasons. To me, it read as the current/live state of the whole keyboard, while to you, maybe it read as an array of all of the "live"/active/pressed keys? Is that right?

Maybe? I would say no, but only because I think "the current/live state of the whole keyboard" and "an array of all of the live/active/pressed keys" are equivalent. I suspect that the former means something different to you than it does to me. What I'm trying to introduce is a structure that represents the "output" state of the keyboard (where the "input" state is the physical keyswitch state), where an "idle" key is represented by the value `Key_Transparent` (and has no effect on the output), and an "active" key has some other value (e.g. `Key_A`), and (in the case of HID Keyboard keys) is currently adding some value to the HID report.

[Having written that, perhaps the terms "input" and "output" might be helpful in this discussion (as opposed to "physical" and "logical")?]

In the case of OneShot (at least, the way it works in #905), it keeps an "active keys" entry "active" after the physical keyswitch is released. In the case of Qukeys, it keeps the "active keys" entry "idle" after the physical keyswitch is pressed (until it determines which value the key should have).

---

Regarding the code for `Layer.lookup()`: I think I should further explain that this PR is, to me, a stepping stone to the event-driven Kaleidoscope changes that I want to make, which will have much bigger benefits for plugin code. I would prefer it if any reference to the "active keys" array was removed from `Layer.lookup()` entirely, and a different, higher-level function called instead in most cases. I would separate out the code for the looking up the key values from a static set of layers from the code for the dynamic object that represents the keyboard state (indeed, that's the way I wrote it in Kaleidoglyph), but preserving backwards compatibility with plugins that call `Layer.lookup()` required this compromise. Having that call to `Runtime.activeKey()` (or whatever you refactor it to) is, to me an undesirable necessity, not the ideal of the design. At some point, I'd like to remove that line, but probably not until after my `KeyEvent` branch matures, reducing the need for plugins to call `Layer.lookup()` in the first place.

---

As a side note, I made liberal use of `operator[]` in Kaleidoglyph, partly as an exercise in learning modern C++ more thoroughly. It had the benefit of obviating the need to think of clever names for getter and setter functions. I wouldn't guess that you'd like the idea, but I thought I'd toss it out there as an option.

obra:
> I'm aware that it's possible to change the initial description; sorry for not updating it. I was negligent in not correcting the details that changed from the original version, probably because the concept of what was represented never changed. I'll happily update it when I get a chance.

And I failed at reading comprehension by not seeing the longer update later. 

I want to re-read-through all of this with fresh eyes tomorrow, possibly by starting to write out all this terminology into the glossary. I don't know that 'active' is a wrong term, just that it's not been used in the codebase and wasn't being used in the way I was expecting, so I was misunderstanding your intent. I want to make sure that we write down the meanings as we hammer them out.

gedankenexperimenter:
> A "keymap" in Kaleidoscope is a datastructure that is a map of the keys on a keyboard, with one value for each key, describing some behavior that should happen when that key is actuated. It is a map of keys.
> 
> A "layer" is one of multiple keymaps inside an ordered datastructure.

I'm not a fan of this terminology, which leaves us without a clear candidate for the singular thing represented by the current `Layer` object (which I named `Keymap` in Kaleidoglyph, for whatever that's worth). I used the names `KeyArray` (for the data structure you're defining as a "keymap"), `Layer` for an object that provides a mapping from a `KeyAddr` to a `Key` (and usually, but not necessarily contains a `KeyArray` to store this mapping), and `Keymap` for a set of `Layer` objects, with an ordering and on/off states. The "active keys" object in Kaleidoglyph is a `KeyArray`, for what it's worth.

I intend that as purely informational. If you want to define "keymap" in that way, I won't say you're wrong to do so, but I do want to make the case that the contents of the codebase are not really that clear, and that users have been having conversations for quite a while in which they refer to "the keymap" as the whole set of layers. I will also beg you to find a term that replaces that "keymap" before you carve it in stone as meaning something else exclusively.

obra:
Users also refer to the Model 01 as "the Keyboardio" ;)

gedankenexperimenter:
> Users also refer to the Model 01 as "the Keyboardio" ;)

Users like @algernon?

Here's a quote from a [discussion on the forum](https://community.keyboard.io/t/borrow-your-laptop/423/3):

> So here’s an easy thing to do: take the factory keymap, paste it after your own, and adjust the layers in the factory keymap, then you can create a macro that toggles your layers off, and enables FACTORY (or does the opposite).

Skimming through old posts there that use the term “keymap”, I'm struck by how frequently @obra used it to refer to a single layer, and used the plural (either “keymaps” or just “maps”) when referring to either the complete set of layers or a subset thereof. I don't think I ever noticed that before, or dismissed it as lack of precision in service to expedience. Clearly, this was not the case.

I have by no means done an exhaustive search (nor do I intend to), but I believe that both @algernon and I have consistently been using the term “layer” almost exclusively when talking to users about a single layer, and “keymap” when referring to either the complete set of layers or the “current (composite) keymap” produced by the set of active layers. I also think that others have been using “layer” and “keymap” in this way, fairly consistently. I could be wrong, and would welcome being corrected on this. @algernon, are you willing to comment?

obra:
I'd really like to decouple the renaming of the 'Layer' object from getting this landed. With the current name, it clearly doesn't make any sense to have the new feature hang off `Layer`, but that's true of .... most everything in here.

After talking terminology through with @algernon, we settled on what's now documented in https://kaleidoscope.readthedocs.io/en/latest/codebase/glossary.html - I -think- the new feature in this PR is reasonably described as the "Override Layer" - If that doesn't seem like it's an accurate description of what's in this PR, lmk.

gedankenexperimenter:
I can see why you would think the "Override Layer" term is appropriate, but I would like to convince you not to use it, as it is far more limited than what my goal is for the purpose of the structure in question. It is meant to be a full representation of the current keyboard state, not merely an extra layer that supersedes the keymap. I want to highlight a comment I made earlier:

> The structure that I'm asking for with this PR answers a different question: "What is this key doing _now_?" From the point of view of a plugin author, that's a very different question, and possibly not even related to the keymap.

With just this PR, It does effectively serve as an "override layer", but that's just a step on the way, broken up into smaller change sets to limit their scope and size. Because of the way the main keyscan loop works now, it's necessary to have this override behaviour, and looking at the code, that seems to be its only function at this point. My intention, however, is to alter that main loop so that event handlers only get called (and (in most cases) HID report(s) sent) in response to keyswitch toggle events. Most plugin interactions with this structure will not be via the setter method directly, but rather just by changing the `Key` value in an event handler.

In my event-driven code, the structure we're talking about gets altered when a keyswitch event is detected, and then when it's time to send a HID report, it gets iterated through, and keycodes added to the report based on its contents. This happens after a single pass through the plugin event handlers for each event, where plugins can refer to this structure for complete information about the current keyboard state, rather than the partial information that they currently have in mid-scan.

From the Kaleidoscope core point of view, the structure does override the keymap, especially at this stage, where all non-idle keys are processed every cycle. But from the plugin point of view, this structure is the reference point for the full state of the keyboard. It can be searched to see if a particular `Key` of any type is currently active, for example, and that should work as soon as this PR is merged, even though it becomes more obvious when Kaleidoscope becomes fully event-driven.

P.S. I still intend to rewrite the initial comment describing the changes, but I still have not had the time to do so.

obra:
For some reason I didn't get a notification from GitHub that you'd commented here, but was -just- coming here to say that with the following code, my name didn't seem to make sense.

```

  // Update the active keys cache if the key toggled on or off.
  if (key_addr.isValid()) {
    if (keyToggledOn(keyState)) {
      Layer.updateOverrideLayer(key_addr, mappedKey);
    } else if (keyToggledOff(keyState)) {
      Layer.updateOverrideLayer(key_addr, Key_Transparent);
    }
  }

```

No worries on timing.

Your description make it sound like the original name of 'LiveCompositeKeymap' isn't that far off the mark (although not a stellar name)


From reading your description, it sounds like your intent is that any key that's currently pressed would be set in this structure and any key that's not pressed would be transparent. is that right? I may still be having trouble understanding how this thing gets used.

Part of it is that I don't understand the question "What is this key doing now?" 

Keyswitches are pressed.

Entries in keymaps are handled by various types of handlers when keys are "toggled on", "pressed", "toggled off", or "unpressed"

Do you mean "what behavior this keyswitch should have if it were pressed right now"?

gedankenexperimenter:
> From reading your description, it sounds like your intent is that any key that's currently pressed would be set in this structure and any key that's not pressed would be transparent. is that right? I may still be having trouble understanding how this thing gets used.

This is correct, and has not changed since the conception of this design in Kaleidoglyph. The following is still true, and central to the design:

> This converts `Layer.live_composite_keymap_[]` from a simple cache to a representation of the keyboard's current state. Barring plugin activity, an idle key will have the value `Key_Transparent`, and a pressed key will have the value of whatever key it's currently mapped to in the keymap. A value of `Key_NoKey` will mask that key until it is released.

I'll try again to make this clear, and hopefully answer all of the following questions at once:

> Part of it is that I don't understand the question "What is this key doing now?"
> 
> Keyswitches are pressed.
> 
> Entries in keymaps are handled by various types of handlers when keys are "toggled on", "pressed", "toggled off", or "unpressed"
> 
> Do you mean "what behavior this keyswitch should have if it were pressed right now"?

To the last question: No. That's closer to the question answered by the existing `Layer.live_composite_keymap_[]` structure. When I say "What is this key doing now?" I mean the active state of the key (with the special value `Key_Transparent` used to represent a key that is idle (i.e. not pressed)). This may seem to be redundant, because that value could be determined by combining information about keyswitch state from the scanner(s) and from the (composite) keymap. That's not necessarily true when there are plugins active, however. Several critical plugins (OneShot and Qukeys, for example) will cause the **logical** (output) state of the keyboard to differ from the **physical** (input) state, with keys active when the keyswitch is not and/or _vice versa_.

Perhaps it would have been clearer if I had posed the question as, "What is this key reporting to the host right now?" (That's not perfectly accurate, though, since not all `Key` values are used in HID reports, and those reports can still be altered by plugins in the `beforeReportingState()` handler (e.g. `shift` keys and TopsyTurvy)).

Ultimately (but not in this PR), this structure can be used as a full replacement for one of the two static NKRO HID reports. Kaleidoglyph stores only one such report, to suppress redundant reports being sent to the host, with the new report only existing temporarily in memory, and then only after the `onKeyswitchEvent()` handlers are called, so only the equivalent of `beforeReportingState()` handlers can access it. That temporary report is then copied just before it gets sent, and is then discarded. Any plugins that need information about the current state of the keyboard need to query the "active keys" structure, which—unlike Kaleidoscope's current HID report—is always complete (except for the new event, which is also available to the event handler). Actually eliminating the extra permanent HID report required an extensive rewrite of the HID module, including changing its API, so I want to be clear that I'm not proposing making that change; I just want to illustrate that though the structure I'm introducing in this PR seems like a small modification to the existing "live composite keymap", the thing that it functionally replaces is not that keymap cache, but rather the information stored in the HID reports (from the point of view of plugins). After this change, plugins can and should stop looking for information about the keyboard's current state in those HID reports, and look instead in this "active keys array", or whatever you want to call it.

It contains all of the information that the HID report does, and additional information that the HID report is not capable of storing. And that's not limited to key position info. For example: two keys that set the same keycode, or whether a modifier comes from a flag or a modifiers-only key.

algernon:
Just a quick note: I had a pass at reviewing this again, and will reply in full a bit later. Just wanted to note I'm on it.

gedankenexperimenter:
I have done a rebase of a local copy of this branch to merge it with the recent restructuring in master and resolve the resulting conflicts. I'm inclined to force-push it here, but I will refrain from doing so if either @obra or @algernon would prefer me to leave it alone.

algernon:
I've been going back and forth on this PR, a lot. I really, really like most of the stuff: `EventHandlerResult::ABORT` is sweet, the tests amazing, and collapsing the two caches into one also felt right. However, I've had this feeling at the back of my mind that something isn't quite right, and it took a lot of time to figure out what, and why I have that feeling. Ultimately, it turned out to be `active_keys_`, and its expansion of scope. On one hand, I like that we no longer have two caches - that was a horrible bandaid, in hindsight. I understand why having a way to query what parts of the keyboard are doing _now_ is useful (checking modifier states, looking at what a key at a particular position is mapped to right now - all good things). I'm just not sure that we should conflate the cache with everything else.

I mean, the original purpose of the cache was to speed up lookups, and the way the cache was laid out, made sense for that purpose. Using the same(-ish) structure for more than speeding up lookups does not feel like the right way to approach things.

However, we really want to move forward, so I'd propose we cherry-pick ::ABORT, and everything else that does not depend on `active_keys_`. Then, for the time being, move `active_keys_` to the `Layer` object, and merge that, until we figure out where we want to go (I know where you want to go, but we'll need to find some common ground first), and how.

gedankenexperimenter:
@algernon — I'm afraid I don't understand your comments. What are the two old caches that have been collapsed, and what's the one new cache?

My point of view about this PR is that it has introduced a new structure which stores the state of the keyboard, and removed a cache. The `active_keys_` structure is not a cache. It does override the keymap, but thinking of it as a cache is very misleading. The `live_composite_keymap_` cache became almost pointless because the active key values override the keymap, since it was mainly there to speed up lookups that would happen every cycle for held keys, the cost for lookups on keyswitch toggle-on being negligible.

I'll reiterate what I said earlier that this structure does not belong to the mis-named `Layer` object. If it's unacceptable to have it as part of the `Runtime` object (despite it clearly representing a key aspect of the _runtime_ state of the keyboard), I'll create a new, separate class and object for it, with whatever names will make @obra happy.

obra:
I'm +1 to split out the abort stuff and lift that so it can get merged.

Once that's sorted, the other part of this might be clearer -- it really
does seem like both @algernon and I are having a really hard time wrapping
our heads around what your proposed new facility is and how it's going to
be used.

Your comment that it's intended to eventually be used to keep track of
prior state of the HID report definitely threw me, since the shape of a
given keyboard's layout seems like it wouldn't really be well suited to
representing the size and shape of a HID report. It also -feels- like this
would end up with us trying to use the data structure for two different
purposes, but that might again be down to me still not grasping the intent
behind the design.

Let's defer the decision about the final location of the datastructure and
its accessors until we're all on the same page about what it's supposed to
do.

gedankenexperimenter:
I don't think there's any point in merging the `ABORT` event handler result on its own. I'll go back to the drawing board. Maybe it would help if I close this PR entirely and start over from scratch. Before I do that, I'll respond to this:

> Your comment that it's intended to eventually be used to keep track of prior state of the HID report definitely threw me, since the shape of a given keyboard's layout seems like it wouldn't really be well suited to representing the size and shape of a HID report.

You've got it exactly backwards: the HID report is ill-suited to representing the state of the keyboard. It doesn't contain all of the information about the state of the keys and their current values, so we have no choice but to reconstruct that state every cycle, and every plugin that needs to know that missing information needs to separately track it.

In Kaleidoglyph, I have a completely rewritten HID module that does store one HID report, but its only purpose is to compare reports to prevent sending redundant ones. The new report only exists temporarily, and it isn't created until after the main event handler hooks are called (the equivalent of `onKeyswitchEvent()` — they don't touch the HID report, because it doesn't exist, and doesn't need to). That HID report goes out of scope after it gets sent. The same thing _could_ be done in Kaleidoscope, because all of the necessary information to construct that HID report (and more) exists in the active keys array.

The ultimate goal is for the HID report to not get reconstructed every cycle, but only in response to input events (and timeouts, but that's just quibbling).

>  It also -feels- like this would end up with us trying to use the data structure for two different purposes, but that might again be down to me still not grasping the intent behind the design.

I _think_ the second purpose that you're alluding to is the keymap cache. This would probably be clearer if I had explained the next phase of the transition. With this PR, a pressed key still gets its value looked up every cycle, so the active keys array looks like its function is basically the same as the old keymap cache. Ultimately, the design is to call event handlers only in response to events (keys toggling on and off), and then to call each plugin handler only once _per event_, rather than once _per active key per cycle_. This is not merely more efficient (0.5ms cycle times), but simplifies the logic of each plugin such that it's much easier to implement them and find and fix bugs.

The real purpose of the array is not fully realized in this PR. It's mean to be iterated after the keyswitch event handlers are called to populate the HID report(s), just before the `beforeReportingState()` hooks are called (which would only be called if an event is detected or generated, rather than running every cycle). It only acts like a cache in this transitional phase; ultimately, it is purely a structure for storing keyboard state, and it does not behave like a cache.

obra:
LiveKeys  naming doesn't seem to fit with https://kaleidoscope.readthedocs.io/en/latest/codebase/glossary.html, so either the glossary needs an entry or the name should get poked.

gedankenexperimenter:
> LiveKeys naming doesn't seem to fit with https://kaleidoscope.readthedocs.io/en/latest/codebase/glossary.html, so either the glossary needs an entry or the name should get poked.

The data that it stores is not (yet) described by any of the entries in the glossary, in my opinion. It does have the same _structure_ what is needed to implement a `Keymap` as defined in the glossary (it's a mapping from `KeyAddr` to `Key` values, with one entry per key on the keyboard), but what it represents is quite distinct. It does _look_ like a sort of "override layer" in the form it takes in this PR, but that's only because of the way the current event loop operates, where a lookup is done every cycle for active keys, rather than only when actual toggle-on or toggle-off events occur. Hopefully this will become more clear when I submit the follow-up PR that contains a working version of the bigger picture.

I'll add a glossary update to this PR to (hopefully) reflect what I have in mind. Note: I also added a `KeyMap` class, which deviates from the existing glossary slightly (capitalization matches common structures like `HashMap`). This was also intentional, since it is intended to be slightly more generic than something that represents the full set of keys on the keyboard (I started out with it as a template class over a number of keys, but that won't compile without `-std=c++17`).

gedankenexperimenter:
This PR is now out of date; see #1024 for an improved version of these commits.

PR #912, opened 2020-10-02
Rework our travis config. 

obra:
Example compilation is still mind-bogglingly slow, but I believe this is better than the status quo across the board.


algernon (approved):
Looks amazing, loving the split jobs. Will merge as soon as Travis finishes.

PR #916, opened 2020-10-06
Correct plugin order for Atreus example

gedankenexperimenter:
The biggest problem here is that Qukeys was listed last. It doesn't matter whether it comes before or after things like Focus, but it really needs to handle events before other keystroke-handling plugins like OneShot, TapDance, Macros, et cetera.

I also moved SpaceCadet up, since it does similar key value resolution, and MousKeys down, because it never changes the value of a key event.

I noticed this because of a comment inside [a user sketch posted on the forum](https://community.keyboard.io/t/atreus-german-qwerty-layout-firmware-with-umlaute/4159), reporting Qukeys causing key repetition.

### DIFF
--- a/examples/Devices/Keyboardio/Atreus/Atreus.ino
+++ b/examples/Devices/Keyboardio/Atreus/Atreus.ino
@@ -106,11 +106,11 @@ KALEIDOSCOPE_INIT_PLUGINS(
   Focus,
   FocusEEPROMCommand,
   FocusSettingsCommand,
-  OneShot,
+  Qukeys,
   SpaceCadet,
-  MouseKeys,
+  OneShot,
   Macros,
-  Qukeys
+  MouseKeys
 );
 
 const macro_t *macroAction(uint8_t macroIndex, uint8_t keyState) {


gedankenexperimenter:
I think this PR is somewhat urgent, as this is the first time Qukeys has been included in any of the hardware examples, and people are just now receiving their Atreus keyboards. Some of them are going to start with this example when customizing, so it's important that the plugin order isn't broken.

alfalfasprossen:
The same order is used in the experimental Model01 example, just in case that is intended to be the rolled-out version soon, it might make sense to fix this there as well? https://github.com/keyboardio/Chrysalis-Firmware-Bundle/blob/master/Keyboardio/Model01/experimental/Model01.ino#L439

Another side effect of Qukeys not being first is that OneShot doesn't work together with keys that are also Qukeys.

gedankenexperimenter:
Thanks, @alfalfasprossen! I was just looking for the source code for the Chrysalis "experimental" firmware to see if it contained the same error, but couldn't find it.

obra:
Thanks! We'll get a Kaleidoscope release built soon.

gedankenexperimenter:
I'm sorry for not catching this earlier. It was glaringly obvious, but only to me.

PR #917, opened 2020-10-06
Add minimum hold time for Qukeys

gedankenexperimenter:
This change makes Qukeys require a certain minimum amount of time for a key to be held before it is eligible to get its alternate (i.e. modifier) value. This should help faster typists avoid unintended modifiers in the output.

### DIFF
--- a/docs/plugins/Qukeys.md
+++ b/docs/plugins/Qukeys.md
@@ -97,6 +97,17 @@ likely to generate errors and out-of-order events.
 >
 > Defaults to `80`.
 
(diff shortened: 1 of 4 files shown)


gedankenexperimenter:
I'll test this and make sure it works this evening.

gedankenexperimenter:
I've tested it out, and it appears to work. I think this will help a bunch of people who've been getting unintended modifiers, and it was a very small change, so I'm recommending merging it soon.

PR #918, opened 2020-10-07
Our previous conditional building was really wrong. It never built if…

obra:
… any .hex existed.

This new version is still wrong, but less so. It at least checks the timestamp on the sketch


algernon (approved):
Shellcheck is happy now, will merge when Travis finishes.

PR #927, opened 2020-10-13
Add deprecation dates for old Qukeys functions

gedankenexperimenter:
These were deprecated on 2019-08-22. I set deprecation dates for 2020-12-31 in both the warnings and UPGRADING.md.


obra:
Doing deprecation removals as dated draft PRs sounds like a fantastic idea.

PR #929, opened 2020-10-14
Improve Qukeys protection against unintended modifiers

### LINKED ISSUE #925: Qukeys rollover improvement idea
I just had an idea that might improve the accuracy of Qukeys in guessing user intentions when there's rollover from a qukey to a subsequent key.

My hypothesis is that users are much less likely to intend the qukey press as a “hold” if the qukey's press follows the _previous_ key's press closely. Additionally, if there's rollover from a non-modifier (maybe whitespace, too?) key to the qukey, it's also less likely that a “hold” was intended.

Basically, we could add a requirement to have a minimum delay between key presses and/or a maximum overlap from the previous key (depending on key value). Iff the requirement is fulfilled, the qukey becomes eligible to take on its alternate value.

As a bonus, it could reduce delays in flushing certain events from the queue.

gedankenexperimenter:
Unintended modifiers are becoming an increasingly big problem among users of Qukeys. This change adds yet another configuration option to prevent this from happening while users are typing fast. It introduces a new requirement to make a qukey eligible to become a qukey; a minimum amount of time that must pass between the keypress event for a non-modifier key and the subsequent keypress event of the qukey.

Fixes #925 (at least in part)

### DIFF
--- a/docs/NEWS.md
+++ b/docs/NEWS.md
@@ -12,6 +12,17 @@ See [UPGRADING.md](UPGRADING.md) for more detailed instructions about upgrading 
 
 ## New features
 
+### Better protection against unintended modifiers from Qukeys
+
+Qukeys has two new configuration options for preventing unintended modifiers in
+the output, particularly when typing fast:
+
+- `Qukeys.setMinimumHoldTime(ms)` sets the minimum duration of a qukey press
+  required for it to be eligible to take on its alternate (modifier) value.
+- `Qukeys.setMinimumPriorInterval(ms)` sets the minimum interval between the
+  previous non-modifier key press and the press of the qukey required to make
+  the qukey eligible to take on its alternate (modifier) value.
+
 ### KALEIDOSCOPE_API_VERSION bump
 
 `KALEIDOSCOPE_API_VERSION` has been bumped to **2** due to the plugin API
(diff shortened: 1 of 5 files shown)


gedankenexperimenter:
I have not tested this at all, but there has been significant demand for improvements in this area, and I want something available for the adventurous users to be able to try out.

gedankenexperimenter:
I made a minor error in the code that protects against integer overflow problems, such that Qukeys completely failed. I fixed that, and now it appears to be working. I'd love to get some users to test it out, so I can come up with a reasonable default value.

gedankenexperimenter:
Rebased and updated. It now only protects against unintended alternate (modifier) values for qukeys that closely follow "printable" keys (letters, numbers, and punctuation, but not `space`, `backspace`, `enter`, _et al_). For fast typists, the vast majority of the errors this change protects against should come from those keys, and if they're using a home-row qukey for `shift`, it could get very frustrating to have to pause at the start of every sentence or proper name (I imagine; I'm nowhere near fast enough to have that problem myself).

An alternative could be to allow `shift` (as an alternate qukey value) to be exempt on the other end, but I believe that code change would be even more complex, so I'd rather try this and see if it works.

gedankenexperimenter:
I'm aware that I still haven't written any testcases for this yet. Hopefully soon, but I might wait until some of the new test infrastructure gets merged.

obra:
I'd still love testcases for this, but I'm going to merge now. Sorry it hung out so long.

gedankenexperimenter:
Testing revealed a minor problem with this. I will have a fix for it soon.

PR #937, opened 2020-10-19
Add `SimHarness.RunMillis(t)` function

gedankenexperimenter:
For now, this is just a wrapper around `RunCycles(n)`, but the two could diverge in the future. `RunMillis(t)` should always cause `t` milliseconds to elapse in the simulator, regardless of how many cycles that takes.

This could save us from needing to update a whole bunch of testcases based on timeouts in the future.

### DIFF
--- a/tests/simulator/timing/test/testcase.cpp
+++ b/tests/simulator/timing/test/testcase.cpp
@@ -1,0 +1,73 @@
+/* -*- mode: c++ -*-
+ * Copyright (C) 2020  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "testing/setup-googletest.h"
+
+SETUP_GOOGLETEST();
+
+namespace kaleidoscope {
+namespace testing {
+namespace {
+
+constexpr uint32_t t_0{0};
+constexpr uint32_t t_1{1};
+constexpr uint32_t t_100{100};
+constexpr uint32_t t_500{500};
+constexpr uint32_t t_500_000{500000};
+
+using ::testing::IsEmpty;
+
+class SimulatorTiming : public VirtualDeviceTest {
+ public:
+  void assertTimeElapses(uint32_t interval) {
+    // Record time at start
+    uint32_t start = Kaleidoscope.millisAtCycleStart();
+
+    // Run cycles until the given amount of time has elapsed
+    sim_.RunForMillis(interval);
+
+    // Record time at end
+    uint32_t end = Kaleidoscope.millisAtCycleStart();
+
+    ASSERT_EQ((end - start), interval);
+  }
+};
+
+TEST_F(SimulatorTiming, SimulatorStart) {
+  // Time doesn't advance in the simulator until after the first cycle runs
+  ASSERT_EQ(Kaleidoscope.millisAtCycleStart(), t_0);
+  RunCycle();
+  ASSERT_EQ(Kaleidoscope.millisAtCycleStart(), t_1);
+}
+
+TEST_F(SimulatorTiming, TimeElapses) {
+  assertTimeElapses(t_1);
+  assertTimeElapses(t_100);
+  assertTimeElapses(t_500);
+}
+
+TEST_F(SimulatorTiming, ZeroTimeElapses) {
+  assertTimeElapses(t_0);
+}
+
+TEST_F(SimulatorTiming, LongTimeElapses) {
+  // Bigger number than uint16_t can hold
+  assertTimeElapses(t_500_000);
+}
+
+}  // namespace
+}  // namespace testing
+}  // namespace kaleidoscope
--- a/testing/SimHarness.h
+++ b/testing/SimHarness.h
@@ -26,6 +26,9 @@ class SimHarness {
  public:
   void RunCycle();
   void RunCycles(size_t n);
+  void RunMillis(size_t t) {
+    RunCycles(t);
+  }
 
   void Press(uint8_t row, uint8_t col);
   void Release(uint8_t row, uint8_t col);
--- a/testing/SimHarness.h
+++ b/testing/SimHarness.h
@@ -26,8 +26,15 @@ class SimHarness {
  public:
   void RunCycle();
   void RunCycles(size_t n);
+  void RunForCycles(size_t n) {
+    RunCycles(n);
+  }
+
   void RunMillis(size_t t) {
     RunCycles(t);
+  }
+  void RunForMillis(size_t t) {
+    RunMillis(t);
   }
 
   void Press(uint8_t row, uint8_t col);
--- a/tests/simulator/timing/sketch.ino
+++ b/tests/simulator/timing/sketch.ino
@@ -1,0 +1,48 @@
+/* -*- mode: c++ -*-
+ * Copyright (C) 2020  Keyboard.io, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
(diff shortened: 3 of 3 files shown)


obra:
I feel like the name "RunForMillis" would be more readable. And... especially since this is intended to ensure that the testing system "does the right thing" for time-based tests, I'd really like this PR to include a test.

gedankenexperimenter:
I chose `RunMillis()` to match the pattern in `RunCycles()`. In English, one is "run for `n` cycles", and the other is "run for `t` milliseconds". Yes, it's more reasonable to drop the "for" in the first case, but I'm sure I'll get it wrong more often if there are two different patterns I have to remember. Perhaps it would be best to have all four:
```
RunCycles(n);
RunForCycles(n);
RunMillis(t);
RunForMillis(t);
```

I'll write a test to go with it as soon as I figure out how.

gedankenexperimenter:
I added variant function names, so there's both `RunMillis(t)` & `RunForMillis(t)`, and also `RunCycles(n)` & `RunForCycles(n)`. I'd rather have just one of each, but I'd much prefer them both to have the same form (either `RunFor(Cycles|Millis)` or `Run(Cycles|Millis)`) to reduce cognitive load.

I also added testcases, one of which currently fails because the simulator doesn't advance time on the first cycle. I have no idea why that happens.

gedankenexperimenter:
One other thought: I could future-proof it from the start (mostly) by having `RunForMillis(t)` do its own loop, calling `RunCycle()` and checking the time elapsed each time, but that seems unnecessary at this point, given that the proximity of the functions in the code.

gedankenexperimenter:
It would be great if some version of this would get merged soon. My impression is that it's just a matter of deciding on the names for these functions. I prefer `RunMillis(t)` & `RunCycles(n)`, with a second choice of `RunForMillis(t)` & `RunForCycles(n)`, but I'd much rather have mismatched names than a long wait until it gets finalized and merged. I expect to be making use of this function extensively in writing tests for both Qukeys and OneShot PRs, and I would like to delay that work as little as possible.

obra:
It's on my near-term list. Please feel free to write tests with the current
names and I'm happy to adapt them when evaluating those tests.

>

obra:
I'm seeing a test failure:

```
[ RUN      ] SimulatorTiming.SimulatorStart
test/testcase.cpp:53: Failure
Expected equality of these values:
  Kaleidoscope.millisAtCycleStart()
    Which is: 0
  t_1
    Which is: 1
[  FAILED  ] SimulatorTiming.SimulatorStart (0 ms)
```

Is that expected?

obra:
As I'm working through this one, I'm having trouble with naming.

I don't want to have multiple similar but different names (RunCycles and RunForCycles) or (RunMillis and RunForMillis)

RunCycles makes sense to me as it's an imperative and parallel to RunCycle.

However, RunMillis feels wrong, We're not running the millis, we're running for an amount of time, measured in millis.

Stylistically, I'm moving the implementations out of the header file and into the .cpp file.

obra:
I'm also taking the opportunity to add in a way to set the number of millis a cycle should take.

gedankenexperimenter:
In real hardware, not every cycle takes the same amount of time, and some cycles take less than a millisecond…

obra:
I don't think I've ever gotten one of our cycles under a millisecond when
testing on real hardware. But being able to set the number of ms per cycle
to a value of the test author's choice is a feature that I believe to be
useful. It will be possible to change the number in code at runtime, so if
you want to simulate longer or shorter cycles as part of a test suite,
that'll absolutely be doable.

gedankenexperimenter:
I've gotten Kaleidoscope to run under a millisecond on the Model01, and Kaleidoglyph to average under 500µs.

gedankenexperimenter:
One thing I recall from the Model01 LED/keyscanner experiments I did is that the cycle times can vary wildly, mainly based on when LED updates are done. Those take a very long time.

gedankenexperimenter:
I realize I'm not communicating my point clearly. I think it would be good to be able to set the number of milliseconds that a given cycle takes when calling `RunCycle()` (or whatever), because _sometimes_ the number of millis that passes is `0`, and sometimes it's more than `1`.

gedankenexperimenter:
(Not that I have any specific testcases for that yet, so I don't mind if it takes a while.)

obra:
I agree 100% that on hardware, especially hardware with LEDs, cycle times
are going to be variable. And what I'm planning here is not intended as
perfect or complete, but as "better than not having any configurability"

The API I'm planning gives you separate methods for setting and getting
millis per cycle. I don't want to complicate RunCycle with the timing
information, but if you find yourself doing it a lot, it should be
straightforward to write a wrapper.

Using a logic analyzer, I don't think I've seen cycles on the Model 01
running under a couple ms ever. Was the timing you were getting coming from
the Arduino side or measured with an analyzer or a scope?

>

gedankenexperimenter:
Timing I got was from serial output from the Model01, with it executing every `n` cycles of the main loop. I don't recall if I ever bothered to verify it on the host side, but it was convincing. It's been quite a while since I ran it, but I could verify it on the host if you need proof. Is there some reason to believe that the clock on the ATmega32u4 is inconsistent?

obra:
Nah. There is no need to run new tests. I was just surprised. It doesn’t match my recollection of what I saw, although my goal when measuring electrical signals was typically more about latency  than cycle time And it has been a while. It is, of course, possible to get a run loop under a millisecond on a device like an avr.  at the point at which it becomes useful to have such things in testing, it shouldn’t be a big deal to improve the simulator. 

did you see my question earlier about whether you see a failing test on the PR as submitted? i’m trying to decide if the weight fix for the bug is to have millis at cycle start as ++time instead of time++.

gedankenexperimenter:
Incidentally, Kaleidoglyph is faster primarily because it doesn't bother processing held keys. If I hold down a bunch of keys at a time, Kaleidoscope slows down substantially, but Kaleidoglyph barely notices, even in its worst case, where I'm hammering on keys all over the keyboard as fast as possible. I talked about this extensively with @algernon back when I first got Kaleidoglyph running, and asked him if ~500µs was a good cycle time. If I recall correctly, he first thought I meant 500ms, and the ensuing conversation led to the optimization where we skip the idle keys, which got Kaleidoscope into the same ballpark, speed-wise.

This is all related to the changes I want to make to Kaleidoscope, incidentally. If I do it right, cycle times should definitely drop, though probably not as dramatically as that first improvement.

gedankenexperimenter:
> did you see my question earlier about whether you see a failing test on the PR as submitted? i’m trying to decide if the weight fix for the bug is to have millis at cycle start as ++time instead of time++.

Yes. Didn't I report that somewhere already?

gedankenexperimenter:
Is it #938?

gedankenexperimenter:
Sorry about that. I'm sure I meant to link #938 to this PR. That testcase is expected to fail right now, but I feel like it shouldn't.

tests/simulator/timing/test/testcase.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    // Record time at end
+    uint32_t end = Kaleidoscope.millisAtCycleStart();
+
+    ASSERT_EQ((end - start), interval);
+  }
+};
+
+TEST_F(SimulatorTiming, SimulatorStart) {
+  // Time doesn't advance in the simulator until after the first cycle runs
```
gedankenexperimenter, line 50: I guess I should have gone into more detail with this comment. This test fails, because the timer reads `0` before the first cycle runs, and still `0` after. From then on, time advances. I don't know why.

gedankenexperimenter: See #938 for the related issue.

obra:
Got it. i have a fix

gedankenexperimenter:
Just making sure I understand correctly, @obra — You've taken over editing and implementing this one to your liking, so I shouldn't waste time working on it any further?

obra:
yep

gedankenexperimenter:
> As I'm working through this one, I'm having trouble with naming.
> 
> I don't want to have multiple similar but different names (RunCycles and RunForCycles) or (RunMillis and RunForMillis)

That is also my preference. I included both in an attempt to give us a better look at them.

> RunCycles makes sense to me as it's an imperative and parallel to RunCycle.
> 
> However, RunMillis feels wrong, We're not running the millis, we're running for an amount of time, measured in millis.

`RunMillis(t)` => "run for `t` milliseconds"
`RunCycles(n)` => "run for `n` cycles"

Sure, the names are brief, but can you really imagine interpreting `RunMillis()` to mean anything else?

obra:
I just don't like how RunMillis reads. 

#946 is an updated and reworked version of this PR with a slightly more extensive test suite.

Issue #942, opened 2020-10-20
make docker-simulator-tests doesn't detect changes to testing infrastructure code


obra:
changes to things inside testing/ don't get properly rebuilt if modified in between runs to "make docker-simulator-tests"

gedankenexperimenter:
I think sometimes changes to code in Kaleidoscope also don't trigger rebuilds of the appropriate source files. When I was working on it last night, I sometimes got linker errors if I didn't wipe out the whole `_build/` directory first.

obra:
I've found the cause and am testing a fix for this one.

PR #943, opened 2020-10-20
Attempt to add KeyAddr versions of Press() & Release()

gedankenexperimenter:
I don't understand why this doesn't work. I'm trying to add a versions of the `SimHarness::Press()` and `SimHarness::Release()` functions that use a `KeyAddr` parameter instead of row & column integers, but I get a linker error when trying to build the test.

If you've got some time to look at this, @algernon and/or @obra (or @paniag), I would appreciate the help. Using a single `KeyAddr` as the parameter makes it easier to define constants that can be used in testcases to make them less error-prone.

### DIFF
--- a/testing/SimHarness.cpp
+++ b/testing/SimHarness.cpp
@@ -14,7 +14,6 @@
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "Kaleidoscope.h"
 #include "testing/SimHarness.h"
 
 #include "testing/fix-macros.h"
@@ -31,16 +30,24 @@ void SimHarness::RunCycles(size_t n) {
   for (size_t i = 0; i < n; ++i) RunCycle();
 }
 
-void SimHarness::Press(uint8_t row, uint8_t col) {
+void SimHarness::Press(KeyAddr key_addr) {
   Kaleidoscope.device().keyScanner().setKeystate(
-    KeyAddr{row, col},
+    key_addr,
     kaleidoscope::Device::Props::KeyScanner::KeyState::Pressed);
 }
 
-void SimHarness::Release(uint8_t row, uint8_t col) {
+void SimHarness::Release(KeyAddr key_addr) {
   Kaleidoscope.device().keyScanner().setKeystate(
-    KeyAddr{row, col},
+    key_addr,
     kaleidoscope::Device::Props::KeyScanner::KeyState::NotPressed);
+}
+
+void SimHarness::Press(uint8_t row, uint8_t col) {
+  Press(KeyAddr{row, col});
+}
+
+void SimHarness::Release(uint8_t row, uint8_t col) {
+  Release(KeyAddr{row, col});
 }
 
 }  // namespace testing
--- a/testing/SimHarness.h
+++ b/testing/SimHarness.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include "Kaleidoscope.h"
 #include <cstddef>
 #include <cstdint>
 
@@ -27,6 +28,8 @@ class SimHarness {
   void RunCycle();
   void RunCycles(size_t n);
 
+  void Press(KeyAddr key_addr);
+  void Release(KeyAddr key_addr);
   void Press(uint8_t row, uint8_t col);
   void Release(uint8_t row, uint8_t col);
 };
--- a/tests/examples/basic-keypress/test/testcase.cpp
+++ b/tests/examples/basic-keypress/test/testcase.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "testing/setup-googletest.h"
+#include "Kaleidoscope.h"
 
 SETUP_GOOGLETEST();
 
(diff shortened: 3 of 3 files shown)


obra:
Thinking about this a bit,  depending on the internal KeyAddr implementation in the testing code may add additional complexity.  Is there a reason not to us trivial macros here?

For example, Instead of 

"constexpr KeyAddr key_addr_A{2, 1};"

using 

"#define KEY_A 2,1"

gedankenexperimenter:
Yes. I find macros like that to be awful. If I see that, I'm going to go looking for a function that takes three args, not four, and I'm going to waste tons of time trying to find it. If that becomes the norm, you're not going to get any tests from me.

I don't understand why you want row and column numbers at all on this level, much less _preferring_ them.

I still want to know why what I did here doesn't work.

gedankenexperimenter:
So, it occurred to me that the linker error might be caused by caching (see #942). When I clear out the cache, I don't get any linker errors…but I do get hundreds of lines of other errors. I guess there's some reason why Arduino header files (i.e. `Kaleidoscope.h`) can't be included in `testing` header files. Probably conflicting preprocessor macros?

I hate C++.

gedankenexperimenter:
I _can_ successfully replace the `KeyAddr` with a single `uint8_t`, then define my constants properly without misleading preprocessor macros in the testcase file, and use `KeyAddr::toInt()` to convert them when needed. It's still pretty damned ugly.

The compiler is perfectly happy with the `KeyAddr` type in the testcase source files, and even in `SimHarness.cpp` — why won't it allow it to exist in the header?

algernon:
The primary issue is that `Kaleidoscope.h` and the headers it includes are assuming an Arduino environment, while googletest (on which the testing infrastructure is built on) assumes standard C++. When you include `Kaleidoscope.h` anywhere where googletest may be included as well, you need to include `testing/fix-macros.h` as well, which fixes up most of the conflicts. It seems to be enough to get things compiling, even if the extra include is a tad ugly.

Perhaps we could add a small block of code to `Kaleidoscope.h` that does the same as `testing/fix-macros.h`, if we're compiling for the virtual arch. That would allow us to... not have `testing/fix-macros.h`. But then we get testing-related stuff in the core firmware parts, which is a bit iffy aswell.

Anyway, I pushed a small WIP commit to the [tmp/testing-keyaddr](https://github.com/keyboardio/Kaleidoscope/tree/tmp/testing-keyaddr) branch. I don't have the time to look through this PR and comment it up, but I wanted to get something out to you before I have the cycles to go through it in detail.

gedankenexperimenter:
I did try that (among other things), with no success. The only difference I see is that I put those includes before the standard library ones, not after, attempting to copy the order used elsewhere. Did you successfully compile and run it?

obra:
Ordering does matter for includes where you're redefining constants. (And indeed, @algernon is correct that Arduino and the C++ STL tend to get into conflict over some questionable decisions Arduino made about macros in their early days.)

On a meta-note, your reply in https://github.com/keyboardio/Kaleidoscope/pull/943#issuecomment-713191592 feels like you're fairly upset that I asked if a simpler solution might work.  Going straight to the ultimatum of "If you make this design decision, I will refuse to contribute" feels very adversarial and is very much not in keeping with the kind of community we're trying to foster.

algernon:
> Did you successfully compile and run it?

Yes, and so did [travis](https://travis-ci.org/github/keyboardio/Kaleidoscope/jobs/737586853).



> The only difference I see is that I put those includes before the standard library ones, not after

As @obra said, order matters here. The compiler will pre-process the files first, replacing all the pre-processor stuff with their final form, in the order they happen. So if we include `Kaleidoscope.h`, which ends up including a bunch of Arduino things, that happen to define `min` and `max` and the other symbols `testing/fix-macros.h` undefs, well... without undefining them later, we'll have a conflict.

Every place where we include `testing/fix-macros.h`, it happens _after_ `Kaleidoscope.h` (or other Arduino-esque headers) gets included, directly or indirectly.

gedankenexperimenter:
> Every place where we include `testing/fix-macros.h`, it happens _after_ `Kaleidoscope.h` (or other Arduino-esque headers) gets included, directly or indirectly.

Yes, I understood that, and added the `#include`s in that order, but it still didn't compile for me. Maybe on that iteration I had forgotten to wipe out the build cache, because it's working for me now. It's quite plausible that I skipped that step one time; I tried to build it with many different variations at least 20 times that I can count (times that I wiped out the build cache). Another possibility is that I neglected to save the file that one time. Anyway, it is working for me now; I just wish the compiler was able to give me a coherent error message when something like this happens.

obra:
I've got it cleaned up and should have something pushed shortly.

obra:
Merged a fixed-up version as 39f0bbe15959cb4889c1fcbc3b818dfcbb3244ec

gedankenexperimenter:
Heh. I think we made identical fixed-up versions independently. =)

gedankenexperimenter:
> On a meta-note, your reply in [#943 (comment)](https://github.com/keyboardio/Kaleidoscope/pull/943#issuecomment-713191592) feels like you're fairly upset that I asked if a simpler solution might work. Going straight to the ultimatum of "If you make this design decision, I will refuse to contribute" feels very adversarial and is very much not in keeping with the kind of community we're trying to foster.

Well said. Please forgive my hasty and adversarial response. I do want to go on record that I'm nearing a "straw that broke the camel's back" situation with what is _to me_ an enormous quantity of code hidden behind undiscoverable (to _me_) preprocessor macros, both in the testing infrastructure and in Kaleidoscope itself. I do not mean to make any threats in order to get my way; I'm just at the limit of my tolerance for games of hide and seek with the code that I'm trying to debug, and I'd like you to know that. It's not that I will refuse to contribute; it's that I will be unable to contribute (it's not a matter of skill, but of the stress that it induces).

Issue #941, opened 2020-10-20
Test simulator: Virtual device produces multiple reports if multiple keys toggle on in one cycle


gedankenexperimenter:
I may be doing something wrong, but when I write a testcase that sends two "simultaneous" keypresses (`sim_.Press(x, y)`) before calling `RunCycle()`, only one of them ends up in the ensuing report.

PR with failing testcase coming shortly.

gedankenexperimenter:
Now that I've played around with this testcase, I suspect that it's just that I don't understand the test simulator. In Kaleidoscope, if two ordinary keyboard keys toggle on in the same cycle, I expect that to generate just one keyboard HID report. In the simulator, however, I get two elements in `RunCycle()->HIDReports()->Keyboard()`. The first contains only one keycode (the one processed first in scan order), and the second contains both. What is the purpose of this first (bogus) report? This looks to me like we're not seeing actual HID reports that are sent to the host, but in-progress HID report states built during the scan cycle. I suppose this is useful for debugging scan-order issues, but I'm primarily interested in the output to the host, not the intermediate states, which could change without affecting the output. Do we have a method for accessing only reports that are actually sent?

gedankenexperimenter:
I'm now mostly convinced that my second analysis was wrong, and that it's the virtual hardware device that's producing multiple reports. I can't find the code that gets run for the virtual device, though, so I'm stuck.

obra:
https://github.com/keyboardio/Kaleidoscope/blob/master/src/kaleidoscope/driver/hid/base/Keyboard.h#L179 is the code path that's causing a second HID report for two keypresses in the same cycle.

It was originally https://github.com/keyboardio/Kaleidoscope-HIDAdaptor-KeyboardioHID/commit/7b91aec54900d63390e4226a997213a19195355a  which itself reverted https://github.com/keyboardio/Kaleidoscope-HIDAdaptor-KeyboardioHID/commit/ff727ee5efbed8e052cfcde12df1d0888840a803

I believe the original code was added in: a4368f13e7a1b58ef7212147f4f0e1ced131ff75

obra:
Reworked the sample sketch for the test and merged to master.

gedankenexperimenter:
[...]

PR #945, opened 2020-10-20
Testcase for simultaneous keypresses

gedankenexperimenter:
This test runs two keypresses in the same cycle, and checks to see that only one report is generated, and that it contains both active keycodes (and no others).

This test currently fails; see #941.

### DIFF
--- a/tests/issues/941/sketch.ino
+++ b/tests/issues/941/sketch.ino
@@ -1,0 +1,537 @@
+// -*- mode: c++ -*-
+// Copyright 2016 Keyboardio, inc. <jesse@keyboard.io>
+// See "LICENSE" for license details
+
+#ifndef BUILD_INFORMATION
+#define BUILD_INFORMATION "locally built"
+#endif
+
+
+/**
+ * These #include directives pull in the Kaleidoscope firmware core,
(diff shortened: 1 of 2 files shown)


obra:
https://github.com/keyboardio/Kaleidoscope/commit/dbfefd463052c205e58fa4d186b1c0e843ec68e6 is the equivalent test I'd worked up a few hours ago.

gedankenexperimenter:
That test is almost equivalent; it doesn't verify that there are no unexpected keycodes in the report(s). I wanted to do it this way to put an example on record.

gedankenexperimenter:
I reworked this testcase, now that @obra has solved the mystery of the multiple reports, and corrected my understanding of what's going on with the simulator. I added a third key in the same cycle, and as expected, only two reports were generated, not three. This is consistent with what the HID code does to deal with the problem, and not, in my opinion, a bug.

I changed the testcase so that it simply verifies that at least one report was sent, and checks that the contents of the last one are what was expected (the three keycodes for the pressed keys, and no others). It now also passes the test, rather than failing.

Closes #941.

Furthermore, we should add some testcases for the rollover cases involving modifier-flagged keys, but that's for another PR. Thanks to @obra for finding the problem when I was unable to navigate my way through the code.

PR #946, opened 2020-10-22
RunForMillis and friends

obra:
This updated branch replaces #937 and adds an additional feature to let you set the number of millis per cycle.


obra:
I think the right thing is probably to convert RunCycles into RunForCycles

obra:
For now, I'm going with RunCycles and RunForMillis. I know it's inconsistent.

gedankenexperimenter:
Not my preference, but I'm sure my feelings on the matter were exaggerated. It'll be fine. Thanks for taking over this one, and getting it merged so quickly.

PR #948, opened 2020-10-24
Add modifier information to KeyboardReport class

gedankenexperimenter:
This change adds the keycodes for keyboard modifier keys to the vector returned by KeyboardReport::ActiveKeycodes(). Also, a simple method for getting the modifiers byte from a keyboard hid report.

Previously, modifiers were left out of the active keycodes returned by the testing infrastructure.

### DIFF
--- a/testing/KeyboardReport.cpp
+++ b/testing/KeyboardReport.cpp
@@ -28,13 +28,38 @@ KeyboardReport::KeyboardReport(const void* data) {
 }
 
 std::vector<uint8_t> KeyboardReport::ActiveKeycodes() const {
-  std::vector<uint8_t> active_keys;
+  auto keycodes = ActiveNonModifierKeycodes();
+  auto mods = ActiveModifierKeycodes();
+  keycodes.insert(keycodes.end(), mods.begin(), mods.end());
+  return keycodes;
+}
+
+std::vector<uint8_t> KeyboardReport::ActiveModifierKeycodes() const {
+  constexpr uint8_t modifier_keycode_offset{HID_KEYBOARD_FIRST_MODIFIER};
+
+  std::vector<uint8_t> active_modifiers;
+
+  uint8_t modifiers{report_data_.modifiers}, mask{1};
+
+  for (uint8_t i{0}; modifiers != 0; ++i, modifiers >>= 1) {
+    if (modifiers & mask) {
+      active_modifiers.push_back(i + modifier_keycode_offset);
+    }
+  }
+
+  return active_modifiers;
+}
+
+std::vector<uint8_t> KeyboardReport::ActiveNonModifierKeycodes() const {
+  std::vector<uint8_t> active_keycodes;
+
   for (uint8_t i = 0; i < HID_LAST_KEY; ++i) {
     uint8_t bit = 1 << (uint8_t(i) % 8);
-    uint8_t key_code = report_data_.keys[i / 8] & bit;
-    if (key_code) active_keys.push_back(i);
+    uint8_t keycode = report_data_.keys[i / 8] & bit;
+    if (keycode) active_keycodes.push_back(i);
   }
-  return active_keys;
+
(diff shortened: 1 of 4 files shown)


gedankenexperimenter:
The lack of modifiers had me scratching my head about an empty report I was seeing in the rudimentary testcases I'm writing for Qukeys today.

testing/KeyboardReport.cpp
```diff
@@ -34,8 +34,19 @@ std::vector<uint8_t> KeyboardReport::ActiveKeycodes() const {
     uint8_t key_code = report_data_.keys[i / 8] & bit;
     if (key_code) active_keys.push_back(i);
   }
+  for (uint8_t i{0}; i < 8; ++i) {
```
obra, line 37: Instead if the 8, could you calculate it with HID_KEYBOARD_LAST_MODIFIER?

obra:
Just one code change I'd like to see on this one. 

A test suite for the Modifiers would be really good, but I won't insist on it.

testing/KeyboardReport.cpp
gedankenexperimenter: I thought about that, but really the right thing is "the number of bits in `report_data_.modifiers`". I'll change it to leave out the literal later tonight.

gedankenexperimenter:
> A test suite for the Modifiers would be really good, but I won't insist on it.

I'll add that, too. I've already done a more complex version of it for the basic Qukeys test I wrote this afternoon; a testcase for this will be trivial.

testing/KeyboardReport.cpp
```diff
@@ -34,8 +34,19 @@ std::vector<uint8_t> KeyboardReport::ActiveKeycodes() const {
...
+    uint8_t bit = 1 << i;
```
obra, line 38: Very minor: we've got a gcc platform level  _BV(i) macro for this instead of the bitshift.

gedankenexperimenter: To me, that's just yet another mysterious macro that I need to go searching for, with no real clues about its provenance. I'd really rather not use it.

gedankenexperimenter:
I think I should add an `ActiveModifierKeycodes()` method to this class, as well. Later today, perhaps.

gedankenexperimenter:
In addition, I think the `Modifiers()` name is poor. How about `ModifiersByte()` or `ModifierBits()`?

gedankenexperimenter:
I've updated the function names and now I think it's more consistent. I went with `ModifierBits()`, but I'm not especially attached to that name; I just wanted something less vague than `Modifiers()`, to make it clear that it's not going to return a vector of keycodes.

I do think the separate `ActiveModifierKeycodes()` function will get used, given that those keycodes are given special treatment in many places, but they should also be included in `ActiveKeycodes()`, which should reflect the whole report. Maybe there's a use for an `ActiveNonModifierKeycodes()`? It would be trivial to add, and would make it clear that `ActiveKeycodes()` does include modifiers…

obra:
Is there a reason that returning the raw modifier bits rather than the modifier keycodes is useful, or was it just the first step in adding a feature? If we can get what we need to test with "just" the keycodes, it'll present a simpler, more consistent API.

gedankenexperimenter:
I'm fine with leaving it out entirely, if you prefer. For some reason, I thought it would be handy to compare a known value of modifier bits in a single byte. This probably is just a habit from working mainly with 8-bit MCUs for so long.

gedankenexperimenter:
I recall now; I added that before it occurred to me to make `ActiveModifierKeycodes()`, which ought to produce much more readable test code. If you agree that `ModifierBits()` should be removed, I'll go ahead and do that.

obra:
Yeah. Let's kill it for now. If it turns out necessary later, we can always
add it back.

obra:
Thanks! merged.

PR #949, opened 2020-10-24
Add some basic Qukeys testcases

gedankenexperimenter:
Just a couple for now; more coming. Requires #948.


tests/plugins/Qukeys/basic/sketch.h
```diff
@@ -0,0 +1,7 @@
+// -*- mode: c++ -*-
+
```
gedankenexperimenter, line 2: There's no copyright header in the `sketch.*` files; @obra, do you care about that?

gedankenexperimenter:
I want to add at least two more quick testcases before this gets merged; one basic full-overlap, and one basic rollover. There are a whole bunch of other ones to add, but I want to put the more systematic coverage off while I think about what I want to add to the testing infrastructure to make it easier to write better testcases.

obra:
It's better to have the copyright, but the sketch is barely code

FWIW, I'm more than happy to merge incremental updates that are "add tests". You might even reasonably say that I'd be overjoyed to do so.

I often don't figure out what infrastructure I want to have until I've written a bunch of stuff that could have benefitted from that infrastructure, but we all have different design processes.

gedankenexperimenter:
> I often don't figure out what infrastructure I want to have until I've written a bunch of stuff that could have benefitted from that infrastructure, but we all have different design processes.

To be clear, I'm engaged in that process right now; not all of it shows up here.

tests/plugins/Qukeys/basic/test/testcase.cpp
```diff
@@ -0,0 +1,164 @@
...
+constexpr KeyAddr key_addr_X{3, 2};
+
+using ::testing::IsEmpty;
+
+class QukeysBasic : public VirtualDeviceTest {};
+
+TEST_F(QukeysBasic, TapQukeyAlone) {
+
+  std::unique_ptr<State> state{nullptr};
+  std::set<uint8_t> expected_keycodes{};
```
obra, line 44: in terms of making the tests feel cleaner and clearer, making expected_keycodes something that can marshal Keys itself, would go a long way.

tests/plugins/Qukeys/basic/test/testcase.cpp
```diff
@@ -0,0 +1,164 @@
...
+  ASSERT_EQ(state->HIDReports()->Keyboard().size(), 0);
+
+  // Release `A`
+  sim_.Release(key_addr_A);
+  expected_keycodes.insert(Key_A.getKeyCode());
+
+  state = VirtualDeviceTest::RunCycle();
+
+  ASSERT_EQ(state->HIDReports()->Keyboard().size(), 1);
+  EXPECT_THAT(state->HIDReports()->Keyboard(0).ActiveKeycodes(),
```
obra, line 60: This expectation definitely feels like something that could either be a function or a macro. Probably at first in this test suite and possibly later lifted up to shared testing infrastructure.

tests/plugins/Qukeys/basic/test/testcase.cpp
```diff
@@ -0,0 +1,164 @@
...
+              ::testing::ElementsAreArray(expected_keycodes));
+
+  expected_keycodes.erase(Key_A.getKeyCode());
+
+  for (int i{0}; i < 10; ++i) {
```
obra, line 65: Comments around what the tests are testing would be very useful here. One way you can do that is to add them as strings after the assertions.

For example:

EXPECT_THAT(foo, ::testing::Le(2)) << "We still generated key reports for every cycle after the second one";

As an aside, I totally don't understand the logic of what's going on in this loop.

obra, line 65: int i{0} is a very uncommon way to initialize ints. I see that you do it in QuKeys, but it's not a convention we use anywhere else in the codebase that I'm aware of. 

When I saw it on your constexprs, I had to stop and try to figure out what's going on. 

 I'd love to hear more about why you've used this style.

tests/plugins/Qukeys/basic/test/testcase.cpp
gedankenexperimenter: Again, exactly what I was thinking. That's my plan, but maybe not until I get back to working on the event-driven branch for a day or so, so I can get Qukeys working properly there. I don't want to take long enough that I completely lose my place.

tests/plugins/Qukeys/basic/test/testcase.cpp
gedankenexperimenter: Thanks for the advice. As it turns out, I had already stumbled across that, and have it on my list of improvements to make.

Regarding what's going on in the loop: In the current version of Qukeys, a tap results in two reports after the key is released, and because of the mid-cycle report problem, I made sure to prevent it from sending more than one report per cycle. That mid-cycle report problem will be solved by the event-driven code that I'm working on in another branch, and I don't want the test to fail because it sends two reports in one cycle.

Now, as it turns out, when I wrote this little loop, I was still unaware of how the test simulator accumulates reports (and purges them), so I could simplify it quite a bit. It's unnecessarily complex because I thought calling `sim_.RunCycle()` twice in a row would throw away any information about a report generated by the first cycle. What the loop is doing is checking up to ten times for a report. It's totally unnecessary.

gedankenexperimenter: `int i{0}` is C++ list initialization. It's something I picked up via conversations with Florian, and I've been using it ever since. It avoids at least one problem where the compiler parses something as a function declaration instead of a variable initialization via a constructor. There are a number of articles on "[the most vexing parse](https://www.fluentcpp.com/2018/01/30/most-vexing-parse/)". With braces, that problem is avoided, and it's possible to consistently use that style to initialize all variables in C++, even primitives.

That said, I'm not going to argue about it if you want to use `=` in cases like this. I just find it useful as a marker that a) I'm looking at/writing C++, not C, and b) this is a variable initialization, not merely assignment. Of course, this breaks down when using `auto`…

gedankenexperimenter:
I'm definitely going to rewrite all of these, and soon, with added functionality in the testcase class as a pilot for adding to the `VirtualDeviceTest` parent class for use in other test suites. I've already got a design building in my head, with queue-like structures for hid reports with timestamps and for expected keycodes that will match those reports, possibly populated via the same method call.

However, as I want to switch gears for a bit and get back to getting Qukeys to work on the event-driven Kaleidoscope branch for a few days, it would be quite helpful to have this merged into master despite its current poor quality in order to save some trouble in rebasing. I completely understand if you don't want to merge it as-is, though, and I can deal with the rebase issues for at least a few days.

Both the event-driven changes and the test infrastructure improvements are equally important to me, so I don't want to spend too much time focused solely on one while neglecting the other. Fortunately, it will make a lot of sense to go back and forth, because if the former is going to proceed, it should be extensively verified by the latter.

tests/plugins/Qukeys/basic/test/testcase.cpp
gedankenexperimenter: I'm glad to hear that you're thinking the same thing as me. I think `expected_keycodes` would be a member variable structure, and would get populated by `Key` values via methods.

obra:
If I were to change the single-item list initializations -in this pr- to more conventional assignments, would that mess with your ability to make progress?  (I'm happy to delay those changes if it will)

gedankenexperimenter:
> If I were to change the single-item list initializations -in this pr- to more conventional assignments, would that mess with your ability to make progress? (I'm happy to delay those changes if it will)

Not at all; go ahead if you prefer it that way.

gedankenexperimenter:
I just discovered that I need to make a few adjustments for this to work with the event-driven branch, so I'll add those, and probably make other minor improvements shortly.

gedankenexperimenter:
It's still only "good enough for now", but I've improved it in several ways, mostly making things more consistent.

obra:
"The perfect is the enemy of the good" 

I'd much rather iterate on something than have it sit fallow. (Our large PR backlog is a testament to us not yet being good enough at this.)

gedankenexperimenter:
By all means, go ahead and merge this now if you're satisfied with it being like this for a couple of weeks. I expect it to be much less time before I make further improvements, regardless of when the merge happens, but I can't say for sure.

obra:
Thanks!

PR #950, opened 2020-10-26
Add rudimentary tests for Redial plugin

gedankenexperimenter:
Again, very basic, rough testcases, but ones that work. (I needed something as simple as possible to test compare master and event-driven, and Redial was as simple as it gets.)


gedankenexperimenter:
TravisCI seems to be running very slow today…

PR #955, opened 2020-10-28
Add timestamps to simulator HID report classes

gedankenexperimenter:
This will be very useful in testcases, especially anything that involves timeouts.

### DIFF
--- a/tests/simulator/timestamps/test/testcase.cpp
+++ b/tests/simulator/timestamps/test/testcase.cpp
@@ -1,0 +1,59 @@
+/* -*- mode: c++ -*-
+ * Copyright (C) 2020  Eric Paniagua (epaniagua@google.com)
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "testing/setup-googletest.h"
+#include "Kaleidoscope.h"
+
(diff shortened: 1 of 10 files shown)


gedankenexperimenter:
I only included a test for `KeyboardReport`, but the code is identical in all HID report types.

tests/simulator/timestamps/test/testcase.cpp
```diff
@@ -0,0 +1,59 @@
...
+  sim_.RunForMillis(delay);
+  current_time += delay;
+
+  sim_.Release(key_addr_A);
+  state = RunCycle();
+  current_time += sim_.CycleTime();
+
+  ASSERT_EQ(state->HIDReports()->Keyboard().size(), 1);
+  EXPECT_THAT(state->HIDReports()->Keyboard(0).Timestamp(), current_time)
+      << "The report should have a correct timestamp";
```
obra, line 54: The description "The report should have a correct timestamp"; doesn't tell the person working with the tests something very useful -- rather than "correct timestamp", it would be better to describe what the correct timestamp is.

If you can amend the commit to fix the two explanatory messages, I'd be very happy to merge this.

gedankenexperimenter: If the test fails because the two arguments to `EXPECT_THAT()` are unequal, it displays those values, showing what they are. The comment is just there to explain the context; if it shows up, the correct value for the timestamp will appear just above it, along with the observed value that doesn't match. Including the expected value in the comment is redundant, but I suppose it's harmless, if you really want it that way.

obra: In this case is "a correct timestamp"   "the most recent key report should be stamped with the current time"?

obra: I think I phrased this poorly - I'd like to change the messages to explain what's being tested to make it easier to understand.   Is it: "Time should not change between when the last report was generated and now"?

I'm happy to actually make the change, but want to make sure I understand what's being tested before I do.

gedankenexperimenter: It's just to verify that the timestamp on the report matches the time in the simulator when the report was sent. Please feel free to change the wording to whatever you like.

obra:
Merged by hand, with a tweak based on my understanding of what the test is testing.

PR #956, opened 2020-10-28
Improvements to Qukeys testcases

gedankenexperimenter:
I'm experimenting here with adding helper functions to the Qukeys basic test suite as a pilot for `VirtualDeviceTest`. This hardly finished, and I want to simplify it more, but it feels much nicer to write the tests this way. (built on top of #955)

### DIFF
--- a/tests/plugins/Qukeys/basic/test/testcase.cpp
+++ b/tests/plugins/Qukeys/basic/test/testcase.cpp
@@ -32,182 +32,190 @@ constexpr KeyAddr key_addr_X{3, 2};
 
 using ::testing::IsEmpty;
 
+class ExpectedKeyboardReport {
+ public:
+  ExpectedKeyboardReport(uint32_t timestamp,
+                         const std::set<uint8_t> &keycodes) {
+    timestamp_ = timestamp;
+    keycodes_ = std::set<uint8_t>(keycodes);
+  }
+
+  const std::set<uint8_t> & Keycodes() const {
+    return keycodes_;
+  }
+
+ private:
+  uint32_t timestamp_;
+  std::set<uint8_t> keycodes_;
+};
+
 class QukeysBasic : public VirtualDeviceTest {
  protected:
+  std::set<uint8_t> current_keyboard_keycodes_ = {};
+  std::vector<ExpectedKeyboardReport> expected_reports_ = {};
+
   std::set<uint8_t> expected_keycodes_ = {};
   std::unique_ptr<State> state_ = nullptr;
+  uint32_t last_event_timestamp_ = 0;
+
+  uint32_t pressKey(uint32_t delay, KeyAddr addr) {
+    sim_.RunForMillis(delay);
+    sim_.Press(addr);
+    last_event_timestamp_ = Runtime.millisAtCycleStart();
+    return last_event_timestamp_;
+  }
+  uint32_t releaseKey(uint32_t delay, KeyAddr addr) {
+    sim_.RunForMillis(delay);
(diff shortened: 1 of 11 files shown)


tests/plugins/Qukeys/basic/test/testcase.cpp
```diff
@@ -32,182 +32,248 @@ constexpr KeyAddr key_addr_X{3, 2};
...
-      << "The second report should be empty";
-
-  state_ = RunCycle();
-  ASSERT_EQ(state_->HIDReports()->Keyboard().size(), 0);
+  pressKey(10, key_addr_A);
+  releaseKey(50, key_addr_A);
+  expectReportAfterCyclesWith(
+      1, Key_A, "Report should contain only `A`");
+  expectReportAfterCyclesWithout(
+      1, Key_A, "Report should be empty");
```
gedankenexperimenter, line 166: This is now pretty close to what I'm aiming for testcases to look like (at least, the ones primarily concerned with keyswitch input & USB output).  This testcase is very simple; it simulates a single key press, followed by a release, followed by two keyboard HID reports, in that order. In each case, the first parameter is the number of milliseconds or cycles to run the simulator before the event. The `pressKey()` & `releaseKey()` methods specify the keyswitch affected. The `expectReport*()` methods specify changes to the report (in terms of `Key` values for readability), and a message to display on failure. All of the test verification is done afterwards, to make it a simple as possible to follow what the test is doing.

tests/plugins/Qukeys/basic/test/testcase.cpp
```diff
@@ -32,182 +32,248 @@ constexpr KeyAddr key_addr_X{3, 2};
...
+
+  constexpr int expected_report_count = 2;
+  ASSERT_EQ(state->HIDReports()->Keyboard().size(), expected_report_count)
+      << "There should be two HID reports after a qukey is tapped";
+
+  for (int i = 0; i < expected_report_count; ++i) {
+    EXPECT_THAT(state->HIDReports()->Keyboard(i).ActiveKeycodes(),
+                ::testing::ElementsAreArray(expected_reports_[i].Keycodes()))
+        << expected_reports_[i].Message();
+  }
```
gedankenexperimenter, line 179: When I was writing a series of `EXPECT_THAT()` statements, I did a lot of copying and pasting, and several times failed to update an index number. This is the reason for the inclusion of the failure message in the `expected_reports_[]` entries. I'm not sure how well this will work until I write a bunch of new testcases for Qukeys and OneShot, but I think that this will be a useful model for doing them clearly and with very few errors.

gedankenexperimenter:
@obra, @algernon, and @paniag — I've got to the point where I've built enough infrastructure into this pilot test suite that I can conveniently write tests with a relatively simple sequence of single-statement input and (expected) output events, followed by verification of the output's correctness (both values and timing of reports). When you get a chance, I'd appreciate it if you could take a look at what I've done here so that we can find ways to improve it and settle on some sort of standard before we write a whole bunch of testcases that will later need to be forced to conform. The main place to start, I think, is the first testcase in `Qukeys/basic/testcase.cpp`.

My plan is to take what works well and submit it as a change to `VirtualDeviceTest` to avoid code duplication (and the inevitable divergence that follows). When you have a little time, please let me know what you think.

obra:
I’m mostly unavailable until next wednesday due to an outside commitment. sorry for my upcoming absence

gedankenexperimenter:
Over the weekend, I came up with some substantial improvements. I plan to submit that as a new PR tomorrow, so I'm closing this one.

PR #958, opened 2020-10-31
Set BOARD_HARDWARE_PATH correctly for run-docker on macOS

gedankenexperimenter:
Because `ARDUINO_LOCAL_LIB_PATH` was being set to a default value for other operating systems before checking the OS version, it would never get set correctly for macOS (unless it was already set externally). This change fixes that problem so that it's no longer necessary to set `BOARD_HARDWARE_PATH` externally when executing `run-docker`.


obra:
Thank you!

PR #959, opened 2020-10-31
Allow for run-docker to be executed by scripts without a TTY

gedankenexperimenter:
This lets a git hook script call run-docker without failing, by setting `DOCKER_RUN_NON_INTERACTIVE`. It's ugly, but I couldn't find any way to detect it automatically that would work for both the command-line and from a git hook.

### DIFF
--- a/bin/run-docker
+++ b/bin/run-docker
@@ -26,9 +26,13 @@ tar  -cf "${XFER_DIR}/kaleidoscope.tar"  \
 (cd "${BOARD_HARDWARE_PATH}/keyboardio" && tar  -cf "${XFER_DIR}/bundle.tar" \
 	--exclude .git --exclude avr/libraries/Kaleidoscope .)
 
+if [ -z "${DOCKER_RUN_NON_INTERACTIVE}" ]; then
+    DOCKER_RUN_INTERACTIVE_OPTS="-it"
+fi
+
 echo "Building the docker image..."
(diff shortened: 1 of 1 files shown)


bin/run-docker
```diff
@@ -26,9 +26,13 @@ tar  -cf "${XFER_DIR}/kaleidoscope.tar"  \
 (cd "${BOARD_HARDWARE_PATH}/keyboardio" && tar  -cf "${XFER_DIR}/bundle.tar" \
 	--exclude .git --exclude avr/libraries/Kaleidoscope .)
 
+if [ -z "${DOCKER_RUN_NON_INTERACTIVE}" ]; then
+    DOCKER_RUN_INTERACTIVE_OPTS="-it"
+fi
+
```
obra, line 32: Does this work for you?

```
if [[ -t 0 ]]; then
    : "${DOCKER_RUN_INTERACTIVE_OPTS:=--tty --interactive}"
fi

```

gedankenexperimenter: I could have sworn I tried that, but apparently not; your version works perfectly for me.

obra:
Awesome. I'll commit it from here.  (FWIW, I had first found a -t 1 version
of the technique and that -doesn't- work.)

gedankenexperimenter:
That's probably what I did, then. It's been long enough that I don't remember the details clearly.

PR #961, opened 2020-11-02
Add event queues for expected HID reports and keyswitch events to testing simulator

gedankenexperimenter:
This change adds a set of functions to the `VirtualDeviceTest` class to make it possible to write simpler testcases involving timed keyswitch press and release events along with corresponding keyboard HID reports.

Other outputs (Consumer & System Control HID reports, LEDs, et cetera) are not (yet) included.

Built on top of #955

### DIFF
--- a/testing/AbsoluteMouseReport.cpp
+++ b/testing/AbsoluteMouseReport.cpp
@@ -16,6 +16,9 @@
 
 #include "testing/AbsoluteMouseReport.h"
 
+#include "Kaleidoscope.h"
+#include "testing/fix-macros.h"
+
 #include <cstring>
 
 #include "MouseButtons.h"
@@ -27,6 +30,11 @@ AbsoluteMouseReport::AbsoluteMouseReport(const void* data) {
   const ReportData& report_data =
     *static_cast<const ReportData*>(data);
   memcpy(&report_data_, &report_data, sizeof(report_data_));
+  timestamp_ = Runtime.millisAtCycleStart();
+}
+
+uint32_t AbsoluteMouseReport::Timestamp() const {
+  return timestamp_;
 }
 
 std::vector<uint8_t> AbsoluteMouseReport::Buttons() const {
--- a/testing/AbsoluteMouseReport.h
+++ b/testing/AbsoluteMouseReport.h
@@ -33,12 +33,14 @@ class AbsoluteMouseReport {
 
   AbsoluteMouseReport(const void* data);
 
+  uint32_t Timestamp() const;
   std::vector<uint8_t> Buttons() const;
   uint16_t XAxis() const;
   uint16_t YAxis() const;
   int8_t Wheel() const;
 
  private:
+  uint32_t timestamp_;
   ReportData report_data_;
 };
 
--- a/testing/ConsumerControlReport.cpp
+++ b/testing/ConsumerControlReport.cpp
@@ -16,6 +16,9 @@
 
 #include "testing/ConsumerControlReport.h"
 
+#include "Kaleidoscope.h"
+#include "testing/fix-macros.h"
+
 #include <cstring>
 
 namespace kaleidoscope {
@@ -25,6 +28,11 @@ ConsumerControlReport::ConsumerControlReport(const void *data) {
   const ReportData& report_data =
     *static_cast<const ReportData*>(data);
   memcpy(&report_data_, &report_data, sizeof(report_data_));
+  timestamp_ = Runtime.millisAtCycleStart();
+}
+
+uint32_t ConsumerControlReport::Timestamp() const {
+  return timestamp_;
 }
 
 std::vector<uint16_t> ConsumerControlReport::ActiveKeycodes() const {
--- a/testing/ConsumerControlReport.h
+++ b/testing/ConsumerControlReport.h
@@ -33,9 +33,11 @@ class ConsumerControlReport {
 
   ConsumerControlReport(const void *data);
 
+  uint32_t Timestamp() const;
   std::vector<uint16_t> ActiveKeycodes() const;
 
  private:
+  uint32_t timestamp_;
   ReportData report_data_;
 };
 
--- a/testing/ExpectedKeyboardReport.cpp
+++ b/testing/ExpectedKeyboardReport.cpp
@@ -1,0 +1,43 @@
+/* -*- mode: c++ -*-
+ * Copyright (C) 2020  Keyboard.io, Inc
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "testing/ExpectedKeyboardReport.h"
+
+namespace kaleidoscope {
+namespace testing {
+
+ExpectedKeyboardReport::ExpectedKeyboardReport(uint32_t timestamp,
+    const std::set<uint8_t> &keycodes,
+    std::string message) {
+  timestamp_ = timestamp;
+  keycodes_ = std::set<uint8_t>(keycodes);
+  failure_message_ = message;
+}
+
+const std::set<uint8_t> & ExpectedKeyboardReport::Keycodes() const {
+  return keycodes_;
+}
+
+uint32_t ExpectedKeyboardReport::Timestamp() const {
+  return timestamp_;
+}
+
+const std::string & ExpectedKeyboardReport::Message() const {
+  return failure_message_;
+}
+
+}  // namespace testing
+}  // namespace kaleidoscope
--- a/testing/ExpectedKeyboardReport.h
+++ b/testing/ExpectedKeyboardReport.h
@@ -1,0 +1,46 @@
+/* -*- mode: c++ -*-
+ * Copyright (C) 2020  Keyboard.io, Inc
+ *
+ * This program is free software: you can redistribute it and/or modify it under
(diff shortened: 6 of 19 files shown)


gedankenexperimenter:
This is the third or fourth iteration in my attempts to make testcase writing more clear and straightforward. I'm finally starting to get satisfied with how it works.

I can now write many testcases as a sequence of statements specifying input events and expected output events, with optional delays (where the simulator runs) specified in either milliseconds or cycles, and a relatively uncluttered way to specify which keycodes should be added to or removed from the current report. For example:
```c++
// let the simulator run for 10ms, then toggle on KeyAddr `key_addr_A`:
PressKey(Millis{10}, key_addr_A);
// run the simulator for 1 cycle, then expect a HID report with the keycode for `Key_A` added to it:
ExpectReport(Cycles{1}, AddKeycodes{Key_A}, "add key A to report");
```
Then, after the whole sequence of input and output has been describe, in chronological order of events, the testcase calls the `LoadState()` function to gather all the accumulated observed HID reports from the simulator. Once that's done, I can call `ASSERT_EQ()` to verify that it recorded the expected number of HID reports, and use a `for` loop and `EXPECT_THAT()` calls to verify the whole sequence of HID reports.

This has the advantage that all the input events will be sent before any `ASSERT*()` calls, which can abort the current testcase, and interfere with the results of subsequent testcases in the same suite if they come before other input events.

gedankenexperimenter:
Regarding the wrapper classes (`Millis`, `Cycles`, `AddKeycodes`, `RemoveKeycodes`): these help with the polymorphic functions, especially `ExpectReport()`. Previously I had functions with terrible names like `ExpectReportAfterMillisWithout(uin32_t, Key, string)`. Now I can make it clear in the arguments themselves what they're specifying, and it's much clearer, especially when making more than one change at a time:
```c++
ExpectReport(AddKeycodes{Key_A, Key_B}, RemoveKeycodes{Key_LeftShift}, "report should drop shift, add A & B");
```
This is much better than using `std::set<Key>` directly, particularly because it tells you what the argument is being used for, but also because it's less cluttered with symbols.

gedankenexperimenter:
A note about the list initializations: both types of wrapper classes need to use the curly braces:
```c++
Millis{15};
AddKeycodes{Key_X};
AddKeycodes{Key_A, Key_B};
```
In particular, the latter one is using an initializer list as a single argument to the constructor, not a variadic constructor. Parens won't work in that case, so I think it's best to always use braces for consistency.

obra:
I'm going to give it a bit longer, but this feels like it's doing too good a job at obscuring what's going on behind the scenes, replacing simple imperative programming with a DSL. 

`PressKey(Millis{10}, key_addr_A);` reads to me that you're pressing the key at addr A for 10ms, not that you're waiting 10ms before pressing key_addr_A.

Can you talk a bit more about why you feel this is preferable to something like the statements below?

runForMillis(10);
pressKey(key_addr_A);

gedankenexperimenter:
I've done some further work on this, and I'm liking how it works more and more. I prefer it because I can write most test scripts as a single statement for each event (input or output), and the first argument is (almost) always the delay in either milliseconds or cycles before that event is transmitted or expected.  For example:
```c++
sim_.RunForMillis(10);
PressKey(key_addr_X);
sim_.RunCycle();
ExpectReport(...);
sim_.RunForMillis(50);
ReleaseKey(key_addr_X);
sim_.RunCycle();
ExpectReport(...);
```
…vs:
```c++
PressKey(Millis{10}, key_addr_X);
ExpectReport(Cycles{1}, ...);
ReleaseKey(Millis{50}, key_addr_X);
ExpectReport(Cycles{1}, ...);
```
That's not just half as many lines of code. Each line begins by telling me what should be happening next. The second thing that line tells me is when it should happen (relative to the previous event). That's followed by other information about how the state should change, all tightly bound to that event. If I'm looking at a testcase, the first question I have is "what is the sequence of key presses and releases?", and that's right there. I might also be looking for where in that sequence reports should be sent. The next thing I need to know, frequently, is the timing.

Yes, it is necessary to know that `PressKey()` only means that a key toggles on, and it's necessary to know that the first argument to all these functions is a delay before the event, but at least that also happens chronologically, from left to right, then top to bottom, just like reading English. I've now written testcases both ways, and my version is orders of magnitude easier to work with, both reading and writing. I will happily rattle off dozens of testcases in a day with this infrastructure; doing so without it is going to take much, much longer, and will involve making many more errors. I'm not saying I won't do it, but it will be painful, and I will definitely be far less productive.

obra:
Ok. Understanding your intent here helps a lot. It may well be that if we workshop the naming and syntax we can get something that's still easy for you to work with and still has the clarity I want.

Looking at your simple example case.

To me: 

```
sim_.RunForMillis(10);
PressKey(key_addr_X);
sim_.RunCycle();
ExpectReport(...);
```

is much easier to understand than:

```
PressKey(Millis{10}, key_addr_X);
ExpectReport(Cycles{1}, ...);
```

In the latter, I have to know that PressKey really means "wait the first arg amount of time, before doing the thing in the second arg" and also that PressKey implicitly executes a keyboard cycle. 
I'm not a -huge- fan of the fact that we're mixing local and sim_ methods in the former. Cleaning that up would be great.

I'm a little lukewarm on ExpectReport not working like the rest of gtest, but that's a separate issue.

```
runForMillis(10);
pressKey(key_addrX);
runCycle();
expectReport(...);
```
has more newlines in it than your current proposal and does more things explicitly than implicitly. Do you find it a lot harder to understand? 

Trying to think through ways to get you something that reads like you want/

Most things we test aren't as timing specific as the stuff you focus on.

I could see a helper like 

`waitThenPress(10, keyAddrX);`

Assuming we could convince astyle that it was kosher, I could see something like:

```
runForMillis(10); pressKey(key_addr_x);
runForMillis(20); pressKey(key_addr_y); 
runForMillis(10); releaseKey(key_addr_x);
runCycle(); // or runForMillis() - just something to get at least one cycle to run so we get the report
expectReports(....)
```

As an aside, "key" is a term we overload a lot. In this case, I wonder if we want to call this 'pressSwitch' - as it's pointing to a place in the matrix. 

[more on all this later]

gedankenexperimenter:
You're missing the most important thing, in my opinion. The thing I want to see all the way on the left of the line is "what event is happening?". I read this:
```
runForMillis(10); pressKey(key_addr_x);
runForMillis(20); pressKey(key_addr_y); 
runForMillis(10); releaseKey(key_addr_x);
runCycle(); // or runForMillis() - just something to get at least one cycle to run so we get the report
expectReports(....)
```
…and I have to hunt for the events that are happening. What I'm trying to do is promote the events to the most visible, prominent place. I don't want to have to search for them, obscured by all the other noise. For just a single keypress, it's not a big deal, but I'm going to have to write testcases involving rollover, often with more than two keys, and long sequences of presses and releases. Without condensing them as much as possible, it's going to be unreadable noise.

obra:
[For those following along, @gedankenexperimenter and I have been having a side conversation about this branch, so discussion here may be a bit disjoint.]

One thing I'm finding as I write some tests with this API is that ExpectReport() with a description of the -changes- to the key report, rather than the explicit report we expect is being a little harder for me to follow along with.  When you were designing the current version of the API, did you try a version where you just spelled out the exact expected keycodes, rather than the changes on each report?

If you tried that and it didn't feel good, I'd love to know what about it seemed off. Was it readability or writability? Or something else?

gedankenexperimenter:
> If you tried that and it didn't feel good, I'd love to know what about it seemed off. Was it readability or writability? Or something else?

I didn't try it out, but I did consider it, and decided that I preferred the “changes only” mode for most cases, because most of the time the _changes_ to the report correspond to the most recent input event(s) moreso than the totality of that report. This way, the contents of the `ExpectReport()` statement would indicate the changes that resulted from an input event, making it easier to follow.

That said, I did have concerns in the back of my mind about the tracking of the current expected report status in this way, and I'm certain that I would eventually want both options, depending on the testcase. While I don't think it would be terrible to have only the option of specifying all keycodes in a given report, I think it would be nice to have the option depending on the testcase.

I do think we should think more about other HID reports in making this design decision. NKRO is a special case because it allows all 232 (I think) keycodes in a single report. 6KRO, Consumer, and System Control reports are more limited, and are probably a good argument in favour of the default choice being to specify the whole list of active kecodes in the `ExpectReport()` or equivalent statement. I certainly didn't get to the point where I considered it a finished product, after all.

obra:
Understood. I'm coming at this from the "generating from a data file" side
of things, so should be able to either output a view or easily-scannable
comments that show off the expected changes to the report, while making the
actual test code a fully enumerated list of the expected values.

I -think- that'll end up with easier easier-to-debug tests, while still
making it easier for a reader of the tests to understand what's going on.

I agree that having the option of doing it either way may make sense,
depending on the test case.

I might want to split it apart so that ExpectedReport just takes some kind
of keycode set object and that object has the add and remove methods. If
you want to modify and pass in updated versions of the keycode set, rather
than constructing a new one, that should be easy to do.

I don't know quite what the code would look like, though.

gedankenexperimenter:
> I might want to split it apart so that ExpectedReport just takes some kind of keycode set object and that object has the add and remove methods. If you want to modify and pass in updated versions of the keycode set, rather than constructing a new one, that should be easy to do.

Not having tried it yet, this sounds like a better idea than what I've done so far. One of the arguments to `ExpectReport()` is a `Keycodes` object (or reference), like one of these:
```c++
ExpectReport(Keycodes{Key_X, Key_Y, Key_Z}, "description");
// or...
ExpectReport(current_keycodes_.add({Key_X, Key_Y}));
```
…where the `Keycodes` class in defined as a child of `std::set<Key>`, with two methods each for `add()` & `remove()`:
```c++
void add(Key);
void add(std::initializer_list<Key>);
```
I'm pretty sure that's the simplest way to do it, though I may have a detail wrong here. I'll check when I can, unless you beat me to it.

obra:
Here's my first hacky attempt -
https://github.com/keyboardio/Kaleidoscope/commit/a22a1e0d7cc5ee7dd454ae1033f3160bbfd3e16c

Critique is appreciated.

obra:
The core Event Queues are now on master. The Millis{}/Cycles{} classes and the polymorphic methods that use them aren't.

I think it makes sense to close this PR and to move those to a new PR. Thoughts, @gedankenexperimenter?

PR #962, opened 2020-11-04
Simplify testcase sketches with default keymap support

gedankenexperimenter:
I'm finding that a lot of testcases don't require sketches with more than a few mapped keys, and because the mapping of the key positions doesn't correspond in a very straightforward way to `KeyAddr{r,c}` values, that was an extra mental burden. With this system, it's possible to define just the keys of interest to the test suite in `keymap-overrides.h`, which gets picked up by the shared `common.h` header so both the sketch and its test suite have access to the same values.

In addition, there's a default set of layers in the `default-layers.h` file, so any testcase sketch can define its keymap by adding the key definitions of interest to `keymap-overrides.h`, and include a line like this:
```
KEYMAPS(LAYER_0, LAYER_1, LAYER_2);
```
Since we're not actually typing on the virtual keyboard, it doesn't really matter where the keys are in relation to each other, except in the relatively rare case of scan-order dependence (which doesn't correspond in an obvious way to the physical layout, either).

Built on top of #961.


obra:
As a general comment, chaining a bunch of not-really-related PRs on top of each other, makes it increasingly hard to review or consider them the deeper you get.  

In specific, since I've requested a very simple change to one of the PRs you're building on and am uncomfortable about another in the chain, this one is basically stuck in purgatory. https://github.com/keyboardio/Kaleidoscope/pull/962/files has all the changes from those other PRs mixed in with this one.

That said, I don't feel like change this makes tests more comprehensible or maintainable. It adds a layer of indirection to key layouts just for tests, so the sketches are unlike what we use for real keyboards.

If your concern is that the Model 01's layout is more complicated than you need for a bunch of tests, it might make sense to use a simpler keyboard layout like the Atreus, Planck, or even the gHeavy ButterStick, which has a -very- simple key layout:

         R0C0, R0C1, R0C2, R0C3, R0C4, R0C5, R0C6, R0C7, R0C8, R0C9, 
         R1C0, R1C1, R1C2, R1C3, R1C4, R1C5, R1C6, R1C7, R1C8, R1C9  

Would that address the problem you're working to solve with this PR?

gedankenexperimenter:
> As a general comment, chaining a bunch of not-really-related PRs on top of each other, makes it increasingly hard to review or consider them the deeper you get.

I was busy working on making the testing infrastructure more usable while other people were occupied elsewhere. I was rather enthusiastic about the work that I was doing, and at each stage dreaded going back to the way things were working before, so I chose to continue. If you don't want me to stack PRs like this, I'll just hold back on submitting them in the first place. I was not expecting you to review them without merging the earlier ones.

> In specific, since I've requested a very simple change to one of the PRs you're building on and am uncomfortable about another in the chain, this one is basically stuck in purgatory.

That's a demoralizing comment. Thanks.

> That said, I don't feel like change this makes tests more comprehensible or maintainable. It adds a layer of indirection to key layouts just for tests, so the sketches are unlike what we use for real keyboards.

I have now written a bunch of testcases, and found that there are two rather time-consuming steps. Writing the test itself takes a while (with the test infrastructure on master; much less so with the other work I've done on it), and putting the sketch together also takes a while, as it's necessary to plan out and edit the keymap.

This change cuts down drastically on the amount of time it takes me to put together a sketch, and once it's done, it becomes very easy to see what's important, because most of the extraneous stuff is left to defaults. I find it to be a _huge_ improvement in how easy it is to set up a test sketch, how much time it takes, how likely I am to make a mistake and need to debug it, and how stress-free that whole process is.

The other advantage is that this makes the same keymap values available to both the sketch and the testcase (if the testcase author wants to use them).

> Would that address the problem you're working to solve with this PR?

Maybe about 10%, but it would miss out on the vast majority of the benefit this provides me.

obra:
Adding some notes on this PR:

@gedankenexperimenter and I have been discussing this one on-and-off on Discord. I've also discussed it a tiny, tiny bit with @algernon.  

All three of us are on the same page about this PR being a potential solution for a very real problem with the testing infrastructure. 

As of right now, I still feel like there's got to be a better way to get the benefits of this PR while still leaving the test sketches as close to "normal" user sketches as possible...but I don't yet know what that better way is.

gedankenexperimenter:
> As of right now, I still feel like there's got to be a better way to get the benefits of this PR while still leaving the test sketches as close to "normal" user sketches as possible...but I don't yet know what that better way is.

Though I disagree somewhat regarding the value of keeping testcase sketches as close to "normal" ones as possible, I think the ideal situation would be to generate the `sketch.ino` file from a common configuration file that's also used to generate the testcase code (I would tend towards having two separate files—one for the config parameters, and another for the testcase procedures). Then the generated sketch files could look more like real sketches, while still having one source that contains just the relevant info.

Perhaps we could find a way to rework what I've done here such that the sketches that I've set up here are just precursors to the actual compiles sketches. Instead of using the C++ preprocessor and `#include` directives, we could use a relatively simple Perl script to construct the sketch from a template file, using values in a YAML (or whatever) config file that can also be used when generating the testcase.

gedankenexperimenter:
It's been six months; I'm giving up on this one.

PR #966, opened 2020-11-09
Standardize Macros key definitions

gedankenexperimenter:
This change brings the Macros plugin into line with other plugins, in defining its keys within the `kaleidoscope::ranges` enum rather than dedicating a flags bit for it to use.

I've included some rudimentary testcases to prove that `M(n)` definitions in the keymap still trigger the correct macro, and that it doesn't interfere with normal keyboard keys.


gedankenexperimenter:
Merging this may make some future changes that I'm planning to submit work better. If we can check the `RESERVED` bit in the `Key` flags to determine whether or not a given key value is plugin-associated, it may be possible to skip calling event handlers in some cases.

obra:
Thanks! This looks like a reasonable modernization.

One thing I'd like to try is to run the tests against master without the refactoring, just to make sure they don't change.

I can do that with a cherry-pick and push to a branch, so I don't think I'm looking for anything from you on it.  If you think of it when pushing other branches that add tests and refactoring (and don't -require- the refactoring  for the tests to work), if you can order the pushed commits so the tests come first, that'd be awesome.

@algernon anything about this PR you wanna weigh in on before merge?

gedankenexperimenter:
Excellent point re: order of commits. In fact, I'm a bit embarrassed that I didn't do it that way; I recall suggesting a couple weeks ago that bugfix PRs should start with a failing test, and this is logically the same situation. Fortunately, it's trivial to reorder the commits in this case.

obra:
Manually tested the new tests at step 1 and at HEAD. Both looked good. Merging. Thanks.

PR #971, opened 2020-11-11
Fix Qukeys minimum prior interval error

### LINKED ISSUE #970: Qukeys rare errors with minimum prior interval
Because of where the code for guarding against integer overflow is placed in the `beforeReportingState()` hook, it doesn't always get called. The result is that there's a very small chance (less than 0.1 second per minute with default settings, I think) that a qukey will immediately return its primary value if the keyboard has just started up or if there hasn't been a keypress for 65 seconds (65536 ms).

gedankenexperimenter:
The code for guarding against integer overflow on the prior interval timestamp was in the wrong place, and wouldn't get executed on cycles when the keyboard was idle, leading to a very slim chance of getting the wrong qukey value if all keys were idle long enough (65 seconds).

Also fixed the same problem in the first quarter-second after the keyboard power on. Not likely to ever be observed, but costs nothing extra to fix.

Fixes #970


gedankenexperimenter:
Note: The existing basic test suite for Qukeys only verifies the on-startup part of the fix. I'll try to add another test for the main part using ktest.

gedankenexperimenter:
Sorry about all the force-pushes; I did things out of order, then made a few mistakes. This was a good exercise in preparing a PR with testcases.

obra:
Force pushes on a PR to improve code quality are almost always the right thing, especially if nobody else is building off them. :)

obra:
Thanks. Merged.

gedankenexperimenter:
I just want to let the record show that this is a case where a bug was discovered by the automated testing simulator, 100%. It would have been very unlikely to occur in the wild, and on the occasions when it would have happened, the consequences wouldn't have been bad, but it had the potential to be a nagging Heisenbug that would have been possible, but not likely, to discover by code inspection (and nearly impossible to reproduce intentionally). If this had been reported by a user, I might have spent hours trying and failing to reproduce it, and staring at the code hoping for an idea. As it is, I did take an hour or so working it out, and another half hour preparing a nice, neat PR to fix it, but with hardly any stress because the simulator could reliably reproduce the issue.

My point is that this exemplifies the benefit we get from the test simulator. It is a huge relief to have this tool, even at this stage where it's still in its infancy. I am very grateful to everyone who contributed to getting it up and running, particularly @noseglasses, @paniag, @algernon, @cdisselkoen, and @obra.

PR #972, opened 2020-11-12
Improve report verification in testing

gedankenexperimenter:
This replaces the `CHECK_EXPECTED_REPORTS()` macro with a new `CheckReports()` method. The new method verifies both the content and timestamps of expected keyboard reports, and provides more output on failures (including details of any unexpected reports).


gedankenexperimenter:
Oops. That was not at all what I meant to do.

gedankenexperimenter:
TravisCI's build test failed, but that was a bogus result; all testcases pass when I run it locally.

PR #977, opened 2020-11-15
Update TapDance documentation

PedroDiogo:
The code example was using an outdated header for `tapDanceAction` that used `row` and `col`.
Updated to use `key_addr` instead.

Solution found in this forum reply: https://community.keyboard.io/t/status-of-using-tapdance-with-chrysalis/4143/5


obra:
Thank you!

PR #979, opened 2020-11-16
Suppress idle keyswitch events from virtual hardware device

### LINKED ISSUE #978: Virtual hardware device generates events for idle keyswitches
Unlike all the real hardware devices, which don't call `handleKeyswitchEvent()` for idle keyswitches, the virtual device does it unconditionally, every cycle, for every `KeyAddr`.

gedankenexperimenter:
The virtual hardware device, unlike others, was calling `handleKeyswitchEvent()` for every keyswitch, every cycle. It should suppress calls corresponding to idle keyswitches, just like the other devices.

Fixes #978.

### DIFF
--- a/src/kaleidoscope/device/virtual/Virtual.cpp
+++ b/src/kaleidoscope/device/virtual/Virtual.cpp
@@ -260,7 +260,8 @@ void VirtualKeyScanner::actOnMatrixScan() {
       break;
     }
 
-    handleKeyswitchEvent(Key_NoKey, key_addr, key_state);
+    if (key_state != 0)
+      handleKeyswitchEvent(Key_NoKey, key_addr, key_state);
     keystates_prev_[key_addr.toInt()] = keystates_[key_addr.toInt()];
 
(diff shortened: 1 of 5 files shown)


gedankenexperimenter:
I'm 95% certain of this fix, but I haven't actually tested it yet, and I didn't want to have to go searching for the code again in the morning.
I'll promote it from draft to regular PR once I've written a testcase. I'll do that by writing a sketch with a simple custom plugin that causes a HID report for an idle keyswitch event.

gedankenexperimenter:
I made a change to the ktest generator in order to get a test that would verify this PR, because ktest wasn't checking HID reports if there were none expected (the testcase I wrote should pass if there are no reports, and fail if there are any). I could create a separate PR just for that change, but I wasn't sure that was warranted.

gedankenexperimenter:
I altered the testcase to make it send only one HID report, instead of two per cycle if it fails. This one should be all ready to go.

obra:
Thanks! I'd thought that I'd previously fixed the "always check the expected reports" thing, but pretty clearly hadn't.

PR #981, opened 2020-11-18
Make ktest match for "keyswitch" token non-greedy

gedankenexperimenter:
The greedy match was including whitespace if more than one space character separated the "keyswitch" token and the first coordinate.


obra:
Thanks!

PR #982, opened 2020-11-18
Add common utility methods for `Key` & `KeyAddr` classes

gedankenexperimenter:
### `Key` changes

In working with several plugins, I've realized that several of them need to perform the same tests, especially checks for Keyboard modifier keys. Qukeys, OneShot, TopsyTurvy, and Shapeshifter all need this same test, and it makes most sense to me to implement it as an instance method of the `Key` class, like so:
```c++
key.isKeyboardModifier();
```
It could also be implemented as an external function (`isKeyboardModifier(key)`), but I think it reads better as a method.

In addition to that, I'm including the related methods `isKeyboardKey()` & `isShiftKey()`. The former includes a check that should used in many places where it's easy to assume that a `Key` value is a Keyboard key, but might not be. (e.g. checking for modifier flags without first verifying that those bits are actually modifier flags). The latter is useful in at least a few plugins, where the status of `shift` is of particular interest. In both cases, the test I've written includes not just "plain" modifiers, but also "combo" modifiers that have other mod flags applied, such as `Key_Meh`. 

In order to write these as methods, I needed access to the `RESERVED` and `SYNTHETIC` definitions, so I moved all of the macros defining bits in the flags byte to a separate file, and included it before the definition of the `Key` class. I'd prefer it if those were defined as `constexpr` instead, but that's beyond the scope of what I'm trying to accomplish here.

### `KeyAddr` changes

I've also found myself wanting to initialize or set `KeyAddr` variables to an invalid address quite often, and having difficulty remembering how. Right now, the most "proper" way to do it is one of these:
```c++
KeyAddr my_addr1 = KeyAddr(KeyAddr::invalid_state);
KeyAddr my_addr2{KeyAddr::invalid_state};
KeyAddr my_addr3;
```
When assigning to a variable:
```c++
my_addr1 = KeyAddr(KeyAddr::invalid_state);
my_addr3 = KeyAddr();
```
The `my_addr3` versions are nice an brief, but it's not obvious when looking at the code what value they'll get. The other examples are quite verbose, and difficult to remember. So, I added one constant, and one method to the `KeyAddr` (`MatrixAddr`, really) class to help make the client code easier to read and write:

- `KeyAddr::invalid`: a `constexpr KeyAddr`, usable in place of `KeyAddr(KeyAddr::invalid_state)`
- `key_addr.clear()`: an instance method that sets the stored value to `KeyAddr::invalid`

The latter is, I think, a reasonable counterpart to the `key_addr.isValid()` method, and the former is nice for explicit initialization when a `KeyAddr` should begin with an invalid value.

### DIFF
--- a/src/kaleidoscope/key_defs.h
+++ b/src/kaleidoscope/key_defs.h
@@ -26,6 +26,52 @@
 
 #include "kaleidoscope/key_defs_aliases.h"
 
+// -----------------------------------------------------------------------------
+// Constant keycode values
+#define HID_FIRST_KEY HID_KEYBOARD_NO_EVENT
+#define HID_LAST_KEY HID_KEYPAD_HEXADECIMAL
+#define HID_KEYBOARD_FIRST_MODIFIER HID_KEYBOARD_LEFT_CONTROL
+#define HID_KEYBOARD_LAST_MODIFIER HID_KEYBOARD_RIGHT_GUI
+
+// -----------------------------------------------------------------------------
+// Constant flags values
+#define KEY_FLAGS         B00000000
+#define CTRL_HELD         B00000001
+#define LALT_HELD         B00000010
+#define RALT_HELD         B00000100
+#define SHIFT_HELD        B00001000
+#define GUI_HELD          B00010000
+
+#define SYNTHETIC         B01000000
+#define RESERVED          B10000000
+
+// we assert that synthetic keys can never have keys held, so we reuse the _HELD bits
+#define IS_SYSCTL                  B00000001
+#define IS_INTERNAL                B00000010
+#define SWITCH_TO_KEYMAP           B00000100
+#define IS_CONSUMER                B00001000
+
+// HID Usage Types: Because these constants, like the ones above, are
+// used in the flags byte of the Key class, they can't overlap any of
+// the above bits. Nor can we use `SYNTHETIC` and `RESERVED` to encode
+// the HID usage type of a keycode, which leaves us with only two
+// bits. Since we don't currently do anything different based on HID
+// usage type, these are currently all set to zeroes.
+#define HID_TYPE_CA    B00000000
+#define HID_TYPE_CL    B00000000
+#define HID_TYPE_LC    B00000000
+#define HID_TYPE_MC    B00000000
+#define HID_TYPE_NARY  B00000000
+#define HID_TYPE_OOC   B00000000
+#define HID_TYPE_OSC   B00000000
+#define HID_TYPE_RTC   B00000000
+#define HID_TYPE_SEL   B00000000
+#define HID_TYPE_SV    B00000000
+// Mask defining the allowed usage type flag bits:
+#define HID_TYPE_MASK  B00110000
+
+
+// =============================================================================
 namespace kaleidoscope {
 
 class Key {
@@ -121,9 +167,83 @@ class Key {
                pgm_read_byte(&(this->getFlags()))};
   }
 
+  // ---------------------------------------------------------------------------
+  // Builtin Key variant test functions
+  //
+  // The following functions return `true` if the `key` parameter is of the
+  // corresponding HID type (Keyboard, Consumer Control, or System Control).
+  constexpr bool isKeyboardKey() {
+    return ((flags_ & (RESERVED | SYNTHETIC)) == 0);
+  }
+  constexpr bool isSystemControlKey() {
+    return ((flags_ & system_control_mask_) == (SYNTHETIC | IS_SYSCTL));
+  }
+  constexpr bool isConsumerControlKey() {
+    return ((flags_ & consumer_control_mask_) == (SYNTHETIC | IS_CONSUMER));
+  }
+  // Not a HID type, but also a builtin `Key` variant.
+  constexpr bool isLayerKey() {
+    return (flags_ == (SYNTHETIC | SWITCH_TO_KEYMAP));
+  }
+
+  // ---------------------------------------------------------------------------
+  // Additional utility functions for builtin `Key` variants
+  //
+  // An "intentional" Keyboard modifier is any HID Keyboard key that has a
+  // keycode byte corresponding to a modifier. This includes combination
+  // modifier keys like `LSHIFT(Key_RightAlt)` and `Key_Meh`. It will not match
+  // a key with only modifier flags (e.g. `LCTRL(RALT(Key_NoKey))`); this is an
+  // intentional feature so that plugins can distinguish between the two.
+  constexpr bool isKeyboardModifier() {
+    return (isKeyboardKey() &&
+            (keyCode_ >= HID_KEYBOARD_FIRST_MODIFIER &&
+             keyCode_ <= HID_KEYBOARD_LAST_MODIFIER));
+  }
+  // It's important to distinguish between "intentional" modifiers and
+  // "incidental" modifiers (non-modifier keys with a modifier flag). The
+  // following test matches any Keyboard modifier key that includes a `shift`
+  // keycode either in its keycode byte or in its flags. In this way, we treat
+  // all of the modifiers in a combination modifier key the same. There should
+  // be no difference between `LSHIFT(Key_RightAlt)` and `RALT(Key_LeftShift)`.
+  //
+  // The `shift` modifiers are special even among the Keyboard modifier keys
+  // (and several plugins single them out for this reason), because they are
+  // used in a conceptually different way: to get a different symbol in the
+  // output. We don't normally think "type `shift`+`1`"; we think "type `!`".
+  constexpr bool isKeyboardShift() {
+    return (isKeyboardModifier() &&
+            ((keyCode_ == HID_KEYBOARD_LEFT_SHIFT ||
+              keyCode_ == HID_KEYBOARD_RIGHT_SHIFT) ||
+             ((flags_ & SHIFT_HELD) != 0)));
+  }
+  // Layer shift keys are conceptually similar to Keyboard modifier keys in that
+  // they are used chorded to change the result of typing those other
+  // keys. They're even more similar to `shift` keys. For both reasons, it's
+  // worth singling them out.
+  constexpr bool isLayerShift() {
+    return (isLayerKey() &&
+            (keyCode_ >= LAYER_SHIFT_OFFSET));
+  }
+
  private:
   uint8_t keyCode_;
   uint8_t flags_;
+
+  // -----------------------------------------------------------------------------
+  // Useful constants for the `Key` object's `flags` byte.
+
+  // There are two bits available for Consumer Control & System Control keys to
+  // use for storing information about the HID Usage Type. These bits are ones
+  // in the following mask.
+  static constexpr uint8_t hid_type_mask_ = HID_TYPE_MASK;
+
+  // The bits that must match exactly when checking if a `Key` value corresponds
+  // to a System Control keycode.
+  static constexpr uint8_t system_control_mask_ = ~hid_type_mask_;
+
+  // The bits that must match exactly when checking if a `Key` value corresponds
+  // to a Consumer Control keycode.
+  static constexpr uint8_t consumer_control_mask_ = (RESERVED | SYNTHETIC | IS_CONSUMER);
 
 };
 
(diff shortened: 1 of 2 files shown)


src/kaleidoscope/key_defs_flags.h
```diff
@@ -0,0 +1,58 @@
...
+#define HID_TYPE_LC    B00000000
+#define HID_TYPE_MC    B00000000
+#define HID_TYPE_NARY  B00000000
+#define HID_TYPE_OOC   B00000000
+#define HID_TYPE_OSC   B00000000
+#define HID_TYPE_RTC   B00000000
+#define HID_TYPE_SEL   B00000000
+#define HID_TYPE_SV    B00000000
+// Mask defining the allowed usage type flag bits:
+#define HID_TYPE_MASK  B00110000
```
gedankenexperimenter, line 58: This section hasn't been altered; I just cut and pasted it out of `key_defs.h`.

gedankenexperimenter:
Interesting. I didn't check carefully enough, and it failed on building hardware devices other than the Model01 and virtual. I'm not sure how KeyboardioHID's `HIDAliases.h` was getting included for some devices, but I added a copy of it to Kaleidoscope, and that seems to fix the problem.

src/kaleidoscope/key_defs.h
```diff
@@ -77,6 +79,29 @@ class Key {
            );
   }
 
+  constexpr bool isKeyboardKey() const {
```
obra, line 82: the "isKeyboard" prefix feels a little funny to me. 

isModifier makes tons of sense to me.

isKeyboardShift feels like a kind of strange thing to be breaking out. Can you tell me a bit more about it?

gedankenexperimenter: Sure. The `isKeyboard*` prefix is meant to convey the HID type of the key (as opposed to Consumer & System keys, and maybe Mouse keys). It also helps to distinguish from other `Key` types—in particular Layer keys.

`isModifier()` is, to me, not specific enough, because there are other things that "modify" key behaviour, some of which get treated the same way as HID Keyboard modifiers by some plugins. Specifically, layer shift keys, which fall into the same category for at least a few plugins, notably Qukeys and OneShot. `isKeyboardModifier()` implies (and includes) `isKeyboardKey()`, and is, I think, less ambiguous.

Regarding `isKeyboardShift()` (as opposed to `isLayerShift()`, thus the name I chose) — Shift is conceptually special, even among modifiers, because it's the one that changes the visible output character. As such, there are plugins (TopsyTurvy, ShapeShifter are examples) that have a particular interest in whether or not the user is holding an intentional `shift` key. I don't know of any plugins that care specifically about any other modifiers, because those don't get used to affect the keymap (for instance, wanting to map a key that's got both `,` and `;` or `+` and `&`).

One of the reasons I'm pushing for the whole package of changes I've been working on is because it's very difficult with Kaleidoscope as it exists today to write a plugin that will allow me to rearrange non-alphabetic symbols at will to create the keymap that I want. That plugin also needs the same test as TopsyTurvy and ShapeShifter (which, even combined can't give me that freedom to rearrange symbols).

I would use `isKeyboardShift()` in several places, though probably not quite as much as `isKeyboardModifier()` or `isLayerShift()`, and it would be nice to just have it available, and not need to rewrite it each time, possibly making errors.

gedankenexperimenter: I started to add `isConsumerKey()` & `isSystemKey()` as well, but the only place that comes to mind that would use them already has a more efficient version than could be constructed this way. Also, `isConsumerKey()` would require furthe rearranging of the code the put the consumer key bitmask in the right place.

obra: Got it.

I like the concept and the rationales make sense. The names don't feel right to me yet, but I don't have a better idea right now.

I want to look at all of this when I'm not overtired.

gedankenexperimenter: In case it helps, here's an example of what `isKeyboardShift()` would look like "in the wild":
https://github.com/keyboardio/Kaleidoscope/blob/c3c8eae89cf3d4c2af7a3b43650db9f702ebc974/src/kaleidoscope/plugin/TopsyTurvy.cpp#L55

gedankenexperimenter: Also, Qukeys and OneShot would have several places with a test that looks like this:
```c++
if (key.isKeyboardModifier() || key.isLayerShift()) {...}
```

gedankenexperimenter:
Regarding function names, I have two more points that I think are worth consideration:

1. We might make a whole series of functions for different `Key` types and sub-types at some point, and they could all share one consistent naming scheme (`is` + `<HID or other type>` + `<subtype>`). For example:
- `isKeyboardModifier()`
- `isConsumerOsc()`
- `isLayerShift()`
I think this is a strong argument for the `isKeyboard*` prefix, because without it, the naming scheme is less consistent, which means a larger mental burden when trying to recall the correct function name.

2. While I prefer these functions as methods (`if (key.isLayerShift())` _vs_ `if (isLayerShift(key))`), largely because I think it reads more naturally ("if [the] *key* is [a] layer shift" _vs "if [this] is [a] layer shift *key*"), the whole thing is more extensible to plugin-specific keys if they're external functions. I don't think it makes sense to put `isOneShotModifier()` or `isTapDanceKey()` in as a method in the `Key` class.

gedankenexperimenter:
I've now talked myself into favouring the function versions rather than methods because of consistency. I've written the code, but I don't want to bother pushing it here unless @obra agrees. It would mean that I could do away with the `key_defs_flags.h` change, too, which always felt like a half-measure to me, anyway.

gedankenexperimenter:
I'm not completely convinced that, because of all the custom `Key` variants that are defined in plugin code, it's better to have even the core variants use the function version, not the method version, for consistency. I'll force-push an update shortly.

obra:
I've been pondering the state of this as of two days ago and haven't yet dug into the new version.  

In terms of naming, I think the reason the "Keyboard" prefix feels wrong to me is that in Kaleidoscope, "Keyboard" is the default. That is, Kaleidoscope is primarily a keyboard firmware and it make sense, I think, to bake that into the naming. But now I'll take a look at the current state of the PR.

src/kaleidoscope/MatrixAddr.h
```diff
@@ -37,6 +37,8 @@ class MatrixAddr {
 
   static constexpr uint8_t invalid_state = 255;
 
+  static constexpr ThisType invalid{invalid_state};
```
obra, line 40: I'd prefer that we default to using = assignment where possible

src/kaleidoscope/MatrixAddr.h
```diff
@@ -90,6 +92,10 @@ class MatrixAddr {
     offset_ = this->row() * cols + c;
   }
 
+  void clear() {
```
obra, line 95: This is absolutely not criticism and is not something that I've been asking for to date, but I'd like us to get better about adding some level of inline doc for functions and methods as we add them.  What I'd be looking for is one or two lines of inline comment above the method explaining why you might call it (and in cases where the calling convention isn't blindingly obvious, how.)

If it's not part of your process, I don't think it makes sense to put a ton of effort into this until we're agreed that a PR's code looks ready to go or the code is harder to understand.

(This particular method is pretty obvious. ;)

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+
+constexpr bool isLayerKey(Key key) {
+  return (key.getFlags() == (SYNTHETIC | SWITCH_TO_KEYMAP));
+}
+constexpr bool isLayerShift(Key key) {
+  return (isLayerKey(key) &&
+          (key.getKeyCode() >= LAYER_SHIFT_OFFSET));
+}
+
+constexpr uint8_t hid_type_flags_mask = HID_TYPE_MASK;
```
obra, line 230: Are these intended as private? If so, they should get named_ to make that clear.

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+constexpr bool isKeyboardKey(Key key) {
+  return ((key.getFlags() & (RESERVED | SYNTHETIC)) == 0);
+}
+constexpr bool isKeyboardModifier(Key key) {
+  return (isKeyboardKey(key) &&
+          (key.getKeyCode() >= HID_KEYBOARD_FIRST_MODIFIER &&
+           key.getKeyCode() <= HID_KEYBOARD_LAST_MODIFIER));
+}
+constexpr bool isKeyboardShift(Key key) {
+  return (isKeyboardModifier(key) &&
```
obra, line 216: If I'm reading this right, this method will only trigger if the key is:

* a keyboard key
* a modifier key
and 
* ( is a shift key OR is another modifer key with the shift bit set)

That last bit was momentarily surprising, but feels like the correct choice for the question we want to ask.   This is a place where doc explaining what's expected probably makes sense.

"isKeyboardShift" still feels like the wrong name. But I think we should talk the names through not as inline comments.

src/kaleidoscope/HIDAliases.h
```diff
@@ -0,0 +1,33 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#pragma once
+
+#include "HIDTables.h"
+
+#define HID_FIRST_KEY HID_KEYBOARD_NO_EVENT
```
obra, line 30: I think these four defines should get lifted into key_defs.h. I don't think they really merit a whole file on their own in Kaleidoscope.

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
 #define ___ Key_Transparent
 #define XXX Key_NoKey
 
+namespace kaleidoscope {
+constexpr bool isKeyboardKey(Key key) {
+  return ((key.getFlags() & (RESERVED | SYNTHETIC)) == 0);
+}
+constexpr bool isKeyboardModifier(Key key) {
```
obra, line 210: I think this makes most sense as "isModifierKey" - as part of that, I think we should affirm that "Modifier"  always means  one of the keys between HID_KEYBOARD_FIRST_MODIFIER and HID_KEYBOARD_LAST_MODIFIER, inclusive.

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+  return (isKeyboardKey(key) &&
+          (key.getKeyCode() >= HID_KEYBOARD_FIRST_MODIFIER &&
+           key.getKeyCode() <= HID_KEYBOARD_LAST_MODIFIER));
+}
+constexpr bool isKeyboardShift(Key key) {
```
obra, line 215: Like "isModifierKey", i think this should be "isShiftKey"

other kinds of shifts that are not referring to keyboard keys should get other names, perhaps ones that don't end in "Key".

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+  return (isKeyboardModifier(key) &&
+          ((key.getKeyCode() == HID_KEYBOARD_LEFT_SHIFT ||
+            key.getKeyCode() == HID_KEYBOARD_RIGHT_SHIFT) ||
+           ((key.getFlags() & SHIFT_HELD) != 0)));
+}
+
+constexpr bool isLayerKey(Key key) {
```
obra, line 222: I don't have a great naming proposal for "isLayerKey" - I think I'd like it to not end in the word "Key"

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+constexpr uint8_t hid_type_flags_mask = HID_TYPE_MASK;
+constexpr uint8_t system_flags_mask = ~hid_type_flags_mask;
+
+constexpr bool isSystemKey(Key key) {
+  return ((key.getFlags() & system_flags_mask) == (SYNTHETIC | IS_SYSCTL));
+}
+
+constexpr uint8_t consumer_flags_mask = (RESERVED | SYNTHETIC | IS_CONSUMER);
+
+constexpr bool isConsumerKey(Key key) {
```
obra, line 239: isConsumerControl

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+}
+
+constexpr uint8_t hid_type_flags_mask = HID_TYPE_MASK;
+constexpr uint8_t system_flags_mask = ~hid_type_flags_mask;
+
+constexpr bool isSystemKey(Key key) {
+  return ((key.getFlags() & system_flags_mask) == (SYNTHETIC | IS_SYSCTL));
+}
+
+constexpr uint8_t consumer_flags_mask = (RESERVED | SYNTHETIC | IS_CONSUMER);
```
obra, line 237: is this private? it should get named_ that way

src/kaleidoscope/key_defs.h
```diff
@@ -202,6 +203,45 @@ typedef kaleidoscope::Key Key_;
...
+}
+constexpr bool isLayerShift(Key key) {
+  return (isLayerKey(key) &&
+          (key.getKeyCode() >= LAYER_SHIFT_OFFSET));
+}
+
+constexpr uint8_t hid_type_flags_mask = HID_TYPE_MASK;
+constexpr uint8_t system_flags_mask = ~hid_type_flags_mask;
+
+constexpr bool isSystemKey(Key key) {
```
obra, line 233: isSystemControl

obra:
I'd love it if "isModifierKey" in src/kaleidoscope/driver/hid/base/Keyboard.h got replaced with a call into the new isModifierKey. Since it was private, no deprecation stuff should be needed.

src/kaleidoscope/MatrixAddr.h
gedankenexperimenter: <pedantic>This isn't technically _assignment_; it's just the "normal" C++ way to call a constructor for initialization, and because of the "most vexing parse" issue, the curly braces are better to use than parens.</pedantic>

That said, we can use "=" here, and I'm okay with changing it if that's what you prefer, but it does look more confusing to me:
```c++
static constexpr ThisType invalid = ThisType(invalid_state);
```
The way I wrote it is much more idiomatic C++, and I personally prefer it, but this certainly isn't a hill I'll fight and die on.

src/kaleidoscope/MatrixAddr.h
gedankenexperimenter: I totally agree. I think failing to do so this time was a bit uncharacteristic. I will happily add those comments once you make a decision on the function names.

src/kaleidoscope/key_defs.h
gedankenexperimenter: I didn't give any thought to privacy here. I did consider using a namespace (e.g. `kaleidoscope::key_flags::hid_type_mask`), which I think would be a good idea, but I don't see any reason to make these constants private (nor do I object to it, for that matter).

src/kaleidoscope/key_defs.h
gedankenexperimenter: Yes, exactly (I definitely should have written explanatory comments here). Basically, the assumption is that `LCTRL(Key_LeftShift)` and `LSHIFT(Key_LeftControl)` are two different expressions that indicate the same intention on the part of the user: a key that's both `control` and `shift`. So it shouldn't matter which one is in the keycode byte and which one is in the mod flags; for the purpose of this function both are modifier keys that include `shift`.

I've been thinking about the `shift` that's part of a combo modifier like this as an "intentional shift", and one that's a mod flag on a non-modifier (e.g. `LSHIFT(Key_1)`) as an "incidental shift". There are at least a few plugins that should make a distinction between the two, and this is meant to help with that.

Also, it's worth noting that `LSHIFT(Key_NoKey)` will not return `true` from this function, and that's intentional, as well. It will allow, for instance, OneShot with the forthcoming "auto one-shot modifiers" option to not identify that key as an auto-one-shot key.

src/kaleidoscope/HIDAliases.h
gedankenexperimenter: No problem. I was waffling on that.

src/kaleidoscope/key_defs.h
gedankenexperimenter: Okay, but I desperately want a term that encompasses both Keyboard HID modifiers and layer shift keys, since they are logically equivalent from the user's point of view. There's an existing function (whose name should probably change regardless) called `isModifierKey()` in Qukeys that does exactly that, grouping those two together.

src/kaleidoscope/key_defs.h
gedankenexperimenter: I'm using the `Key` postfix for the functions that are determining which variant of the `Key` class the object is. Plugins can then get in on this game: `isOneShotKey()`, `isTapDanceKey()`, `isDualUseKey`, _et cetera_.

gedankenexperimenter:
Re: naming these functions:

The most important thing to me about the names that I came up with is that it's a naming _scheme_, not just a collection of unrelated names. `is<VariantName>Key()` is the template, and I want it to be extensible to any variant used by any plugin, so that it's trivial to choose a new one, and easy to remember the existing ones, without any exceptions to the rule. The `Key` postfix is, to me, a useful device for identifying the function as one of the family of `Key` variant testers. I think it reads better than the same name without it, but it's not critical.

"Keyboard" is by far the most vexing term (except perhaps "Key" itself) when it comes to ambiguity. It can refer to the hardware device, or the HID Keyboard protocol, but if there's going to be a naming _scheme_, not just a collection of loosely-associated names that don't all follow a pattern, there needs to be _some_ label that identifies the `Key` variant that corresponds to the values that result in HID Keyboard device output. I'm fine with it being something else, but it's got to be _something_. "Standard" or "normal" or something to that effect is fine with me if you really don't like "Keyboard", but I think it would be best for it to match the terms used in the HID module, so that `is<Variant>Key(key)` means that `key` is ultimately handled by the `<Variant>` HID report, through an object of the same name. And that would mean keeping the term `Keyboard`.

I don't mind if you prefer some other naming scheme, but I really do want to convince you that it's better to have a scheme than no scheme, and it will be better not to make special exceptions from it, even for the most common variant.

gedankenexperimenter:
I've discovered a minor problem with `KeyAddr::invalid`. Curiously, if I use it as a default parameter value in a function definition in a header file, I get a linker error. I can use it in other places without any problems, but it doesn't surprise me too much that it breaks down somewhere, and with an inscrutable error. Therefore, I'm going to replace it with a class method: `KeyAddr::invalid()`, which doesn't suffer from this problem. I'll also suggest an alternative name: `KeyAddr::none()`. @obra, do you have a preference?

obra:
"invalid" is definitely not a stellar name for this. 

"none" is better.

Is nonexistent better? Is noAddress?

gedankenexperimenter:
My thoughts on the KeyAddr name(s):

- `KeyAddr::invalid()` matches `key_addr.isValid()` pretty well.
- `KeyAddr::none()` makes sense, but might cause some confusion because it returns just a single value, whereas `KeyAddr::all()` returns an iterator. It's also nice and short, and is probably my favourite.
- `KeyAddr::nonexistent()` avoids the comparisons to `KeyAddr::all()`, but it's multisyllabic, and I'm pretty sure it takes me longer to read, because I hear words in my head when I read. Also, it's a name that I might choose as a test function that returns a boolean (in a different project that doesn't have an established `isFoo(bar)` pattern). `KeyAddr::invalid()` also falls into this category.
- `KeyAddr::noAddress()` is my least favourite of these. The repetition of "address" doesn't help me at all, and it strikes me as having the same downsides as `KeyAddr::nonexistent()`

If you think it's acceptable, I'll change it to `KeyAddr::none()`. If, at any point, we think of a better option, it will be easy to add, update, and deprecate.

obra:
none is fine with me

gedankenexperimenter:
I updated the PR to include the changes discussed on Discord, making the test functions instance methods of the `Key` class. This required some reorganization of the code, because some preprocessor `#define` statements needed to be moved up to the top of the file. I also changed the helper constants, moving them from the `kaleidoscope::key_flags` namespace to private class (const) variables.

PR #1000, opened 2020-12-21
Move plugins back out into their own directories

obra:
One of Kaleidoscope's marquee features is that most functionality is implemented in plugins.  

By moving the plugins back into their own directories, we make the core easier to comprehend, dramatically cut build times, and do a better job of spotlighting the plugins in Kaleidoscope.

In followup PRs, we should extract the LEDControl plugin and move the examples back to their plugins. (But we should be moving them back into Kaleidoscope as part of the build process.)


obra:
I'm making the executive decision to merge this, since it dramatically improves build times, reduces the chances of naming conflicts and better highlights one of the things that makes Kaleidoscope unique.

PR #1002, opened 2021-01-02
Remove stuff that was deprecated as of december 31

obra:


algernon (approved):
I usually move the removed APIs in docs/UPGRADING.md to the "Removed APIs" section, but it makes sense to leave them in place a little after the removal.

So this has my stamp of approval as-is. Thank you!

PR #1006, opened 2021-01-12
This small plugin adds affordances for print-statement debugging.

obra:


plugins/Kaleidoscope-Devel-ArduinoTrace/src/kaleidoscope/plugin/ArduinoTrace.cpp
```diff
@@ -0,0 +1,28 @@
...
+//namespace kaleidoscope {
+//namespace plugin {
+//
+//EventHandlerResult ArduinoTrace::onSetup() {
+//  return EventHandlerResult::OK;
+//}
+//}
+//}
+//
+//kaleidoscope::plugin::ArduinoTrace ArduinoTrace;
```
algernon, line 28: If the plugin doesn't provide anything else other than pulling ArduinoTrace in, do we even need thtis file?

plugins/Kaleidoscope-Devel-ArduinoTrace/src/kaleidoscope/plugin/ArduinoTrace.h
```diff
@@ -0,0 +1,45 @@
...
+/*
+#pragma once
+
+#include "kaleidoscope/Runtime.h"
+
+#ifndef ARDUINOTRACE_SERIAL
+#define ARDUINOTRACE_SERIAL DebugStderr
+#endif
+
+#include "ArduinoTrace.h"
```
algernon, line 26: Is it intended that the entire contents are commented out? If so, I'm not sure I understand how this plugin works...

algernon (changes requested):
The plugin header and cpp both seem to be completely commented out - I'm not sure that's intended.

obra:
Good catch. Those files should have been git rm-ed. I've now force-pushed

PR #1012, opened 2021-01-28
Clarify the Macros plugin `Key` value range

gedankenexperimenter:
This replaces the magic number `24576` with the old definition of the start of the Macros plugin's `Key` range, to make it clear that it will match bitwise tests for "not a plugin key" (the two high bits are both zeros) and the `SYNTHETIC` flags bit.

I also elaborated on the reasons for keeping the existing `Key` values stable and added a comment addressing the more likely case where new values are being added, rather than fighting the last war and focusing only on the one legacy plugin range that exists outside the canonical plugin range, whose values have already been incorporated here.

### DIFF
--- a/plugins/Kaleidoscope-Ranges/src/Kaleidoscope-Ranges.h
+++ b/plugins/Kaleidoscope-Ranges/src/Kaleidoscope-Ranges.h
@@ -17,6 +17,9 @@
 
 #pragma once
 
+// Included for definition of legacy Macros plugin key range:
+#include "kaleidoscope/key_defs.h"
+
 namespace kaleidoscope {
 namespace ranges {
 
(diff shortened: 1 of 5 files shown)


gedankenexperimenter:
I believe that what we really need is a way for Chrysalis to query the version number of Kaleidoscope's firmware, so that it can update any keymap entries to new values as necessary when flashing a new firmware version. That is, no doubt, a longer term solution, and it should still be possible for Chrysalis to upgrade keymaps by the user telling it which version of Kaleidoscope the values had been set for, and which new version has been or will be flashed.

Keeping these legacy `Key` values in the core range indefinitely, without a plan to transition them to the plugin range to modernize the code is very troubling.

gedankenexperimenter:
If I have time, I'll write a testcase that will verify that existing `Key` values in Kaleidoscope-Ranges haven't changed, and add it to this PR.

PR #1021, opened 2021-03-26
Use new --build-property option if arduino-cli version is > 0.14

gedankenexperimenter:
The option `--build-properties` was deprecated in v0.14.0 of arduino-cli. This change uses the new option `--build-property` instead if the version of arduino-cli being called is newer than that, thus avoiding deprecation warning messages and inoculating the build system against the removal of the deprecated version of the option.


obra:
Thanks!

gedankenexperimenter:
Note: I only changed one makefile that might benefit from this. The other one doesn't get called when running tests, so I was less certain of making edits to it. Also, Emacs complains about what appears to be spurious trailing whitespace on one line in that file, making me even more hesitant to change it.

I do not enjoy dealing with makefiles...

PR #1029, opened 2021-04-15
Remove (mostly) useless `MouseWrapper.diagonalize()` function

### LINKED ISSUE #633: Mouse movement is not "diagonalized"
I've been looking at the MouseKeys code, and noticed that it appears to be intended to keep the mouse cursor movement speed consistent whether it's moving in just one direction or on a diagonal. However, this isn't what happens. [`diagonalize()`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys/MouseWrapper.cpp#L137-L142) doesn't really get used.

It gets called from [`MouseWrapper_::move()`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys/MouseWrapper.cpp#L156-L157), which in turn is called from [`MouseKeys_::beforeReportingState()`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys.cpp#L104), but there, the value for each of the parameters `moveX` and `moveY` is either [`0`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys.cpp#L85), [`speed`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys.cpp#L102), or [`-speed`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys.cpp#L100). In turn, [`MouseKeys_::speed`](https://github.com/keyboardio/Kaleidoscope/blob/5acb97940b7be559122f39d0d2bac80b53149cb4/src/kaleidoscope/plugin/MouseKeys.cpp#L27) has a default value of `1`, for which `diagonalize(1)` will always return `1`. This means that a combination of mouse movement keys will result in faster motion on the diagonal than rectilinear motion from a single mouse movement key because the x & y components don't actually get scaled back.

It would be possible to do the scaling if it were done on sub-pixels per millisecond, rather than pixels per second, but I'm not convinced that it's really worth the trouble. It would probably be better to just accept that diagonal mouse motion is faster, and remove the `diagonalize()` function entirely.

gedankenexperimenter:
Because the mouse speed values are so low, `diagonalize()` was never actually used (integer division caused the result to be zero, which triggered a fallback to the original input value (`1`), which meant that the curbing speed on diagonal motion never actually happened. Since it's not clear that slowing down diagonal motion is better, and it saves 108 bytes of PROGMEM to remove the function, let's just do without it.

Closes #633

### DIFF
--- a/plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys/MouseWrapper.cpp
+++ b/plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys/MouseWrapper.cpp
@@ -110,30 +110,12 @@ uint8_t MouseWrapper_::acceleration(uint8_t cycles) {
   }
 }
 
-// Get the diagonalized version of a value, i.e. value * sqrt(2) / 2. If the
-// value ends up being zero, return the original value instead.
-static int16_t diagonalize(int16_t value) {
-  // 99 / 140 closely approximates sqrt(2) / 2. Since integer division
-  // truncates towards zero we do not need to worry about truncation errors.
-  int16_t diagonalValue = value * 99 / 140;
-  return (diagonalValue == 0 ? value : diagonalValue);
-}
-
 void MouseWrapper_::move(int8_t x, int8_t y) {
   int16_t moveX = 0;
   int16_t moveY = 0;
   static int8_t remainderX = 0;
   static int8_t remainderY = 0;
   int16_t effectiveSpeedLimit = speedLimit;
-
-  if (x != 0 && y != 0) {
-    // For diagonal movements, we apply a diagonalized speed limit. The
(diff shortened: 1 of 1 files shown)


algernon (approved):

obra:
Thanks!

PR #1030, opened 2021-04-15
Fix timeout checker

gedankenexperimenter:
`Runtime.hasTimeExpired()` had a minor flaw. Because it was comparing two values using `>` instead of `>=`, it meant that a timeout set at 20ms wouldn't actually time out until 21ms elapsed.

I will add updates to all the affected testcases later; I'd like to wait until #1024 is settled first, so I'm leaving this in draft status for now.


gedankenexperimenter:
I've added updates to existing testcases that used timeouts to remove the extra 1 ms padding.

gedankenexperimenter:
Note: This PR is not 100% compatible with #1036 as-is. If both PRs get merged, the second one will need to make a minor adjustment to the AutoShift timeout testcase to remove the extra 1 ms padding.

PR #1037, opened 2021-05-29
Add SpaceCadet "no-delay" mode

gedankenexperimenter:
This changes SpaceCadet's binary "enabled/disabled" status into a set of modes: "on", "off", and "no-delay". The first two modes are unchanged, but the new "no-delay" mode will result in SpaceCadet immediately sending the primary (modifier) value to the host when the key is pressed, then replacing that value with the "tap" value if it's released before timing out.

This new mode can be turned on by calling `SpaceCadet.enableWithoutDelay()`. I chose that name over `enableNoDelay()` to reduce ambiguity; the function turns on "no-delay" mode, which is not independent of SpaceCadet being "on" or "off".

Closes #506

### DIFF
--- a/docs/NEWS.md
+++ b/docs/NEWS.md
@@ -12,6 +12,13 @@ See [UPGRADING.md](UPGRADING.md) for more detailed instructions about upgrading 
 
 ## New features
 
+### SpaceCadet "no-delay" mode
+
+SpaceCadet can now be enabled in "no-delay" mode, wherein the primary (modifier)
+value of the key will be sent to the host immediately when the key is pressed.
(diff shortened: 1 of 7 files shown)


gedankenexperimenter:
Note: if #1030 is merged before this, some tests will fail because of timeout padding, and will need adjustment.

gedankenexperimenter:
I've discovered a problem with this one, where a SpaceCadet key doesn't get released until the timeout expires.

ryansch:
This is great!!

gedankenexperimenter:
@ryansch, if you notice any problems with it, please let me know! I didn't do any extensive testing on real hardware.

PR #1045, opened 2021-06-05
Remove expired deprecated implementation inside OneShot

obra:
The deprecations on this stuff were expired. @gedankenexperimenter - does this look right to you?

### DIFF
--- a/docs/UPGRADING.md
+++ b/docs/UPGRADING.md
@@ -628,10 +628,9 @@ Older versions of the plugin were based on `Key` values; OneShot is now based on
 `KeyAddr` coordinates instead, in order to improve reliability and
 functionality.
 
-The following deprecated functions and variables will be removed after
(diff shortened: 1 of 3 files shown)


gedankenexperimenter:
> The deprecations on this stuff were expired. @gedankenexperimenter - does this look right to you?

Not quite. Those deprecation dates were speculative, from several months ago, and I thought I had removed them, but I guess I missed some; sorry about that.  We should do a full pass and give lots of stuff removal dates, in fact, and correct some of these.  How far in the future should deprecation dates be for stuff that was deprecated last week?

obra:
Given the current timelines around releases, I'd be ok with something as
short as 30 days.

gedankenexperimenter:
Sorry about not taking notice of this PR when putting together #1115.  These changes were replicated there, so I'm closing this one.

PR #1047, opened 2021-06-06
Saving 36 bytes on the Model 01 example firmware

obra:


src/kaleidoscope/key_defs.h
```diff
@@ -212,7 +212,7 @@ class Key {
   // used in a conceptually different way: to get a different symbol in the
   // output. We don't normally think "type `shift`+`1`"; we think "type `!`".
   constexpr bool isKeyboardShift() const {
-    return (isKeyboardModifier() &&
+    return (isKeyboardKey() &&
```
gedankenexperimenter, line 215: Unfortunately, though this looks right, it's not.  The trouble is that there are plugins that need to distinguish between a "shift that is part of a modifier key" and a "shift flag that is part of a symbol".  For purposes of this test, `LSHIFT(Key_RightAlt)` and `RALT(Key_LeftShift)` both need to return `true`.  `LSHIFT(Key_1)`, on the other hand, returns `false`.

If there isn't a OneShot test that fails because of this change, that's an oversight on my part; there should be.

gedankenexperimenter (changes requested):
It surprises me that inlining saves bytes in progmem, but if so, I'm certainly not against it. The other change, unfortunately, will break some functionality.

src/kaleidoscope/key_defs.h
obra: mm. This wasn't super-obvious from the code.
I wonder if we want a function to encapsulate "             ((flags_ & SHIFT_HELD) != 0)))"

obra:
> It surprises me that inlining saves bytes in progmem, but if so, I'm certainly not against it. The other change, unfortunately, will break some functionality.

Well, there's an overhead to function calls, especially function calls that are non-void. If the compiler makes the wrong optimization choice, we need to give it a better hint.

PR #1048, opened 2021-06-07
Correct broken links and typos in `device-apis.md`

gedankenexperimenter:
There were a bunch of broken links in this document, mainly because of directory restructuring, but some file names and class names were off, too ("ATMega32U4" instead of "ATmega32U4", "Catalina" instead of "Caterina"). I've checked all the links, and they seem to be correct. I don't have the skills to check the embedded example code, however.


obra:
<3 Thank you

PR #1049, opened 2021-06-07
Add removal dates for various functions & variables

gedankenexperimenter:
There were many deprecated functions that had uncertain dates, so they were omitted.  I'm proposing August 1 as the earliest date for removal.  I separated the plugin deprecations from the core ones to make it easier to update them independently.


obra:
Merged. In an ideal world, we'd also update the docs, but this is far better than not doing it. 

Thanks!

gedankenexperimenter:
Sorry about overlooking the docs; I certainly have a tendency to forget those.

PR #1051, opened 2021-06-08
Add ModLayer plugin

gedankenexperimenter:
This is a plugin that allows a single key to act as a combination layer shift and modifier simultaneously.  There are a couple of reasons why it might be better to not have it as a plugin, but instead to just make it part of Kaleidoscope's core: it's extremely simple, so souldn't cost much, and as a plugin it doesn't correctly pass certain tests like `key.isKeyboardModifier()` & `key.isLayerShift()`.

I'll experiment with comparing the core vs plugin implementations, and keep this PR updated with my findings.


gedankenexperimenter:
I've marked this as "ready", but I think we should still discuss the possibility of making `ModLayer` keys simply part of the core, and not a plugin.  The main reason to do so is that `Key::isLayerShift()` and `Key::isKeyboardModifier()` could return `true` for ModLayer keys.  The main disadvantage of that is if those functions are inlined, it would take more PROGMEM space.

PR #1055, opened 2021-06-08
Fix trailing comments on ktest `PRESS` & `RELEASE` lines

gedankenexperimenter:
`ktest` wasn't inserting `//` before the line-end comment for a `PRESS` or `RELEASE` directive.


obra:
Thanks!

Issue #1061, opened 2021-06-13
Custom modifiers for screen readers


Jookia:
**Is your feature request related to a problem? Please describe.**
Currently I can one-shot press modifiers, or sticky hold them, but I can't do this with regular keys. Some applications have their own fake modifiers that I'd like to use.

**Describe the solution you'd like**
I'd like to be able to add custom modifiers, like insert and caps lock.

**Describe alternatives you've considered**
Holding keys down kinda sucks.

**Additional context**
See here for some common screen reader keybinds:
https://help.gnome.org/users/orca/stable/preferences_key_bindings.html.en

gedankenexperimenter:
OneShot has two very new features that might be good enough, accessible by putting one of two special OneShot keys in your keymap:
- `OneShot_MetaStickyKey` — tap this, then the next key you tap will become "sticky" until the next time it gets pressed.
- `OneShot_ActiveStickyKey` — hold any key(s), then tap this key, and all the held ("active") keys will become "sticky".

Jookia:
Those do look helpful, but aren't exactly what I'd like. Is there a
function call I could make to trigger stickiness for a certain key?

gedankenexperimenter:
It would also be possible to modify OneShot or (I'm pretty sure, but I'll have to check) add a simple custom plugin to a sketch that will turn any key you define into a OneShot key; that might be the best solution.

We could consider defining the auto-OneShot detection function weakly, so it can be overridden in user sketches, as well.

gedankenexperimenter:
I've just had a look, and OneShot doesn't currently expose a way to make keys sticky, but I think it definitely should. I'm hopeful that will even let us move the handling of the aforementioned special keys into a separate plugin.

algernon:
The major reason OneShot is restricted to layer and modifier keys is quite simple: space. OneShot stores the active OneShot keys in a 16-bit bitmap, 8 modifiers + 8 layer keys. If we increase that to 24 or 32 bits, the code size and memory usage will grow prohibitively much.

What we _could_ do, is allow one to set what keys are considered modifiers. That would still limit us to 8 modifiers total (or 16 at max, if we make it possible to have 0 OSL keys), but perhaps that's enough for the time being.

alfalfasprossen:
I might have a plugin that does what you need https://github.com/alfalfasprossen/Kaleidoscope-Repeater
[...]

PR #1062, opened 2021-06-14
Move handling of OneShot "meta" keys to a new `OneShotMetaKeys` plugin

gedankenexperimenter:
This PR removes the handling of the two special OneShot keys `OneShot_MetaStickyKey` and `OneShot_ActiveStickyKey` from the OneShot plugin and puts it in a (new) separate plugin named OneShotMetaKeys.  To do so, I've added some public functions to OneShot that allow other plugins to manipulate the OneShot state of key addresses.

In addition, I made `OneShot.isStickable(key)` a weakly-defined function, so that it can be overridden by user code.

This decreases the size of OneShot by almost 500 bytes in progmem!

### DIFF
--- a/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
+++ b/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
@@ -133,7 +133,7 @@ bool OneShot::isStickable(Key key) {
     return true;
 #endif
   }
-  return false;
+  return true;
 }
 
 bool OneShot::isTemporary(KeyAddr key_addr) {
--- a/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
+++ b/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
@@ -145,12 +145,41 @@ bool OneShot::isTemporary(KeyAddr key_addr) {
   return temp_addrs_.read(key_addr);
 }
 
+bool OneShot::isPending(KeyAddr key_addr) {
+  return (glue_addrs_.read(key_addr) && temp_addrs_.read(key_addr));
+}
+
 bool OneShot::isSticky(KeyAddr key_addr) {
   return (glue_addrs_.read(key_addr) && !temp_addrs_.read(key_addr));
 }
 
 bool OneShot::isActive(KeyAddr key_addr) {
   return (isTemporary(key_addr) || glue_addrs_.read(key_addr));
+}
+
+// ----------------------------------------------------------------------------
+// Public state-setting functions
+
+void OneShot::setPending(KeyAddr key_addr) {
+  temp_addrs_.set(key_addr);
+  glue_addrs_.clear(key_addr);
+  start_time_ = Runtime.millisAtCycleStart();
+}
+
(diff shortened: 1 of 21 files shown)


gedankenexperimenter:
The changes in here should also make it possible to address #1061

plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
```diff
@@ -128,18 +129,18 @@ bool OneShot::isStickable(Key key) {
     if (n < oneshot_key_count) {
       return bitRead(stickable_keys_, n);
     }
-#ifndef ONESHOT_WITHOUT_METASTICKY
-  } else if (key == OneShot_MetaStickyKey) {
-    return true;
-#endif
   }
-  return false;
+  return true;
```
algernon, line 133: I'm not sure I understand why this is changed from returning `false` to `true`. Calling `OneShot.isStickableDefault(Key_A)` should return `false` under normal circumstances.

algernon (changes requested):
Looks good, apart from the one change I left an in-line comment about. Once that's resolved, this should be ready for merge.

plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
gedankenexperimenter: I disagree.  There is much more utility in a _sticky_ `A` than a _one-shot_ `A`.

obra:
(This sounds like it wants explanatory comments in the code discussing the rationale. they need not be extensive)

plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
gedankenexperimenter: By default, there's no setting in OneShot that would allow any non-modifier, non-layer shift key to become sticky, because normally it would need to enter the one-shot state first, so this doesn't change anything for someone using OneShot on its own, with no customizations.

But if the default here is `false`, and the user wants to use the `OneShot_MetaStickyKey`, which itself acts like a OneShot key, that key will be either blocked from becoming sticky, or the user will need to override the `OneShot.isStickable()` function.  We can't just have the OneShotMetaKeys plugin override `OneShot.isStickable()`, because that would prevent the user from overriding it, and/or result in multiple `weak` definitions of the same function, which probably results in undefined behaviour or link-order dependence.

It also means that if someone wants to write a very simple plugin like the `OneShotInsert` plugin in the testcase for 1061, they would also need to override `OneShot.isStickable()` in order to make the chosen key(s) behave like a "normal" OneShot key.

The default should be `true`, which is the same as the default behaviour for `OSM()` & `OSL()` keys—those ones need extra code added to turn off stickiness, too.

algernon: With this explanation, the change makes sense, thank you! Can we have a little explanation above the return, so whoever reads it in the future, will not be as confused as I was?

gedankenexperimenter: Done!

algernon:
Splendid, merging!

PR #1070, opened 2021-06-18
[WIP] Add (builtin) ModLayer combination keys

gedankenexperimenter:
This is a replacement for #1051, but with support for ModLayer keys in the Kaleidoscope core.  This way, the feature takes substantially less space in PROGMEM (~250 bytes), and provides more functionality (plugins can treat ModLayer keys the same way as other "momentary" keys.

There are a number of open questions about how to this should be implemented (function naming, keymap markup, helper function(s)), but the basic functionality is there.


obra:
Just on general principles, let's get docs in the same PR.

PR #1080, opened 2021-09-03
Atreus: Add build date/time to version info

The-Compiler:
Example output:

    Keyboardio Atreus - Kaleidoscope locally built on Sep  3 2021 at 09:40:27

Let me note that I've only tested this on my Model01 (https://github.com/keyboardio/Model01-Firmware/pull/110) and the example output is based on that. I can test on a real Atreus later, but I don't see why anything would be different to the Model01 in this regard.

### DIFF
--- a/examples/Devices/Keyboardio/Atreus/Atreus.ino
+++ b/examples/Devices/Keyboardio/Atreus/Atreus.ino
@@ -18,7 +18,7 @@
  */
 
 #ifndef BUILD_INFORMATION
-#define BUILD_INFORMATION "locally built"
+#define BUILD_INFORMATION "locally built on " __DATE__ " at " __TIME__
 #endif
 
 #include "Kaleidoscope.h"


obra:
Thanks. Merged. We *may* want to consider moving this into Kaleidoscope proper, since it does seem to be a useful thing to want in a sketch.

The-Compiler:
FWIW I do think it's a good idea to have the build date and time of the sketch (rather than the underlying Kaleidoscope, which is probably not rebuilt when the sketch changes)?

When doing some kind of adjustments in the sketch and them not working out as intended, I think it's good to have a way to verify that it was indeed flashed properly as a debugging aid. Or when having multiple keyboards (in my case: Model 01, Atreus, soon Model 100) and doing changes to the configuration of all three, it's great to have a way to double-check all of them are running the newest config (especially if one is in the office and can't be flashed right away).

obra:
> FWIW I do think it's a good idea to have the build date and time of the sketch (rather than the underlying Kaleidoscope, which is probably not rebuilt when the sketch changes?).

This is a valid concern.  (The core is...unreliably rebuilt.)

> When doing some kind of adjustments in the sketch and them not working out as intended, I think it's good to have a way to verify that it was indeed flashed properly as a debugging aid. Or when having multiple keyboards (in my case: Model 01, Atreus, soon Model 100) and doing changes to the configuration of all three, it's great to have a way to double-check all of them are running the newest config (especially if one is in the office and can't be flashed right away).

I was more thinking of defining a "standard" build information helper macro. We do have some hooks inside the sketch to have code that gets shoved into it at compile time.  I don't know that this is a priority or wants too much magic, but it wouldn't be a bad thing to just provide to the user.

PR #1097, opened 2021-12-23
Example ShiftBlocker plugin

gedankenexperimenter:
This is an example custom plugin called "ShiftBlocker".  It uses the `onAddToReport()` hook to suppress any `shift` key that's being held from being added to the HID report before it's sent.  This is useful along with Macros for typing capital letter accented characters using dead keys (where the dead key must not be typed with the `shift` active, but the accented character should have a held `shift` applied).

### DIFF
--- a/examples/Features/ShiftBlocker/ShiftBlocker.ino
+++ b/examples/Features/ShiftBlocker/ShiftBlocker.ino
@@ -1,0 +1,106 @@
+/* -*- mode: c++ -*-
+ * ShiftBlocker -- A Kaleidoscope Example
+ * Copyright (C) 2016-2022  Keyboard.io, Inc.
+ *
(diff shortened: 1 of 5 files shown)


obra:
Sorry I somehow missed this before now.

If it's "just" intended as an example, I think it might make sense to inline the code in the sketch file, rather than in a separate header file.  It's not all that common for examples to be multi-file and "build a custom plugin inline in your sketch" is another useful pattern for users to see.

gedankenexperimenter:
I'm not sure why I used a separate header file for this one.  I think it's fine to also have examples where we're including header files that are written alongside the sketch, but it's certainly more common to write the custom plugin in the sketch file itself.  I'll reconfigure it.

PR #1101, opened 2022-02-02
DynamicMacros: Allow large macro maps

lugray:
The previous implementation required that the entire macro map be provided all at once on one serial line of input. For large macro maps, this could exceed the input buffer size.

This provides new serial commands `macros.reset` and `macros.appendMap` to allow breaking a large macro map across multiple serial lines. This necessitated an alternate tracking of the end of the macro map, saving it explicitly to be able to pick back up where writing last left off, which allowed some other quality of life improvements:
  - No longer requires the extra `MACRO_ACTION_END`
  - A macros.map read can easily send only the active macro map instead of the entire memory space.

Old serial scripts will continue to work.

Breaking changes:
  - We use two bytes that could previously have been part of a macro map internally. Maps which just fit in the reserved space will need more space reserved.
  - Previously stored maps will need to be recreated, since the use of the storage space has changed.

### DIFF
--- a/plugins/Kaleidoscope-DynamicMacros/README.md
+++ b/plugins/Kaleidoscope-DynamicMacros/README.md
@@ -58,7 +58,8 @@ The plugin provides a `DynamicMacros` object, with the following methods and pro
 
 > Reserves `size` bytes of storage for dynamic macros. This must be called from
 > the `setup()` method of your sketch, otherwise dynamic macros will not
-> function.
+> function. Two bytes of this are used for internal tracking, the remainder is used
+> for your macro map.
 
 ### `.play(macro_id)`
 
@@ -73,21 +74,28 @@ please refer to the documentation therein.
 
 ## Focus commands
 
-The plugin provides two Focus commands: `macros.map` and `macros.trigger`.
+The plugin provides the following Focus commands:
 
 ### `macros.map [macros...]`
 
 > Without arguments, displays all the stored macros. Each macro is terminated by
-> an end marker (`MACRO_ACTION_END`), and the last macro is followed by an
-> additional marker. The plugin will send back the entire dynamic macro storage
-> space, even the data after the final marker.
+> an end marker (`MACRO_ACTION_END`). The plugin will send back only the used
+> dynamic macro storage space.
 
 > With arguments, it replaces the current set of dynamic macros with the newly
-> given ones. Macros are terminated by an end marker, and the last macro must be
-> terminated by an additional one.
+> given ones. Macros are terminated by an end marker.
 
 > In both cases, the data sent or expected is a sequence of 8-bit values, a
 > memory dump.
+
+### `macros.appendMap [macros...]`
+
+> Append to the current macro map. This is provided to allow for macro maps that
+> exceed the serial input buffer for a single line.
+
+### `macros.reset`
+
+> Clear all previously defined macros.
 
 ### `macros.trigger macro_id`
 
--- a/plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
+++ b/plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
@@ -24,6 +24,8 @@ namespace plugin {
 
 uint16_t DynamicMacros::storage_base_;
 uint16_t DynamicMacros::storage_size_;
+uint16_t DynamicMacros::used_size_base_;
+uint16_t DynamicMacros::used_size_;
 uint16_t DynamicMacros::map_[];
 Key DynamicMacros::active_macro_keys_[];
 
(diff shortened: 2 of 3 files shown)


lugray:
cc @algernon since you're the original author.

obra:
Hi Lisa,

Thanks so much for the contribution. I know @algernon has been reviewing it
and has some notes about how he'd like to see this evolve, but it sounds
like he's running a day or two behind on getting those posted. I'm really
sorry for the delay.

best,
jesse

algernon:
Hi!

First of all, apologies for the late response, I had most of my thoughts written a day or two ago, but didn't get around to actually post them. Sorry!

I love that you not only found an issue where the design of the protocol and (at least some of) its commands might run into unexpected problems, but you're even proposing a way to fix them. With that said, I'd love to know if you've seen the problem in practice, or is it theoretical? I'm asking this because we use a similar method for updating the keymaps, and at ~1.5-2k of data, I have not seen it happen yet. Okay, that's not _quite_ true, because on macOS, if we send too fast from host to device, data will be lost, but that happens at much smaller chunks (about 64-128 bytes, I believe?) too.

This would be important to know, because if this is a reproducible problem, then we need to fix much more than just DynamicMacros, because the same problem will eventually surface for keymaps and DynamicTapDance too, and Colormap aswell, to name a few off the top of my head.

The firmware itself should have no trouble with large datasets, because DynamicMacros - like pretty much all similar plugins - reads value by value, and does not read entire lines at a time. The only fixed-size buffer in the Focus implementation is the one holding the command name, which is limited to 31 bytes at the moment.

Assuming the root problem is reproducible, then rather than applying a workaround to DynamicMacros, I'd rather see the protocol itself updated for a proper fix. Something like splitting lines every once in a while, with a `\` signalling continuation, and perhaps even a bit of delay between lines, to allow the other end to read. Or perhaps we don't even need to split lines, just add a delay here and there? Or commands that have a larger set of data to deal with could be changed to return them in chunks instead, chunks the size the host sets: for example, we could have `macros.map` still deal with the full data, but `macros.cmap` (for chunked map) would take an extra set of arguments: the starting offset, and the number of bytes to read/write. `macros.cmap 0 10` would return 10 bytes of the map, starting at offset 0. `macros.cmap 0 -10 1 2 3 4 5 6 7 8 9 10` would write 10 bytes (the minus sign signalling input) to the map, starting at offset 0. We would need a `macros.size` command in this case though, to be able to figure out the size of the whole thing, and adjust our requests appropriately.

With that said, there's a lot of things I like about this PR, on the idea level: `appendMap` and `reset` are interesting. However, Focus and Focus commands are intentionally kept few and simple, with most logic deferred to the host side of things, for compiled code size reasons mostly (not such a big deal on ARM MCUs, but on AVR, every new command is a noticable bump in code size when we have very limited space). So rather than having explicit `reset` commands, most plugins just expect the host to issue a full update with null-values (eg, full 255s), the same size as returned by the same command. For example, to clear all dynamic macros, one could just do a `focus macros.map $(focus macros.map | sed -e 's/[0-9]+/255/g')`. This highlights another important property of the protocol design: we always return the full set of data. That lets the host figure out how big the keymap / dynamic macro area / etc are, without having to issue a separate command. With this proposed PR, Chrysalis - or any other tool - would be unable to determine the space reserved for dynamic macros, except via trial and error. Even if we added an error case when the data written back either via `appendMap` or `map` would overflow, we'd still defer that discovery to write-time, while currently we could warn the user before even sending the data to the keyboard. Deferring it to write-time has another downside: even in the failure case, we will overwrite previous contents, so the host would need to keep a backup, and restore in that case. Mind you, that's still the case currently, but the host knows in advance whether what they're about to send will fit or not.

In short, the idea behind Focus is that commands are simple and dumb, and most logic is pushed to the host side. If something can be done by building on core commands, then we generally try not to have overlapping focus commands. There are exceptions, of course, when the cost-benefit ratio tips the scales the other way (such as updating a single LED outside of colormap: updating all would cost more than it does to have an extra command that updates a single led). I'd like to keep this design. But if the size of our data, whether input or output, causes issues, we'll need to fix that, on a protocol level, rather than on a plugin-by-plugin basis.

lugray:
> Okay, that's not quite true, because on macOS, if we send too fast from host to device, data will be lost, but that happens at much smaller chunks (about 64-128 bytes, I believe?) too.

This was the key. I'm working from macOS, and assumed that setting the baud rate would be sufficient to send the data appropriately. I guess not. I was seeing the transmission cut off at a reproducible byte, so made assumptions. Splitting my `macros.map` across several writes _without_ newlines was entirely sufficient.

Thanks for the detailed write up and explanation of the thought behind the protocol!

algernon:
Out of curiosity, what was the cutoff you saw? And did the problem only occur when sending data to the keyboard, and not the other way around?

This is something that should be documented somewhere visible, so having an exact cut-off size would be very useful to know. I never got around to figuring out the exact size.

PR #1102, opened 2022-02-07
Improvements to Macros upgrading documentation

gedankenexperimenter:
We've been getting a few reports about problems people are having with old Macros code when upgrading to newer versions of Kaleidoscope.  This adds two new sections to the Macros documentation: one describes what to do with deprecated `MACRODOWN()` code, and another explains why existing Macros that use `U(key)` don't work properly, and provides an example solution to the problem.

In addition, the Macros/UPGRADING.md document was not appearing anywhere on kaleidoscope.readthedocs.io, so I moved its contents to the bottom of the README.md file instead, to make that documentation more discoverable.


obra:
Thank you so much! Merging.

obra:
sure!

PR #1106, opened 2022-02-10
Add ktest command to emulate delays without running cycles

gedankenexperimenter:
This adds a function `DelayForMillis()` to emulate a delay without calling `Kaleidoscope.loop()`.  This can happen for a number of reasons, such as a plugin doing some communication over the serial port.

Also added the corresponding `DELAY` command for ktest scripts.

### DIFF
--- a/testing/bin/ktest-to-cxx
+++ b/testing/bin/ktest-to-cxx
@@ -164,6 +164,25 @@ sub load_from_text {
                 }
 
             },
+            delay => sub {
+                my $content = shift;
+                if ( $content =~ /^(\d+)\s*(\w*?)$/ ) {
+                    my $count = $1;
+                    my $unit  = $2;
+                    if ( $unit =~ /milli|ms/ ) {
+                        return { millis => $count };
+                    }
+                    else {
+                        die
+"Line $line_num: failed to parse a 'delay' clause: $content";
+                    }
+                }
+                else {
+                    die
+"Line $line_num: failed to parse a 'delay' clause: $content";
+                }
+
+            },
 
         };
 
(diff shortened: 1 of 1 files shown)


testing/SimHarness.cpp
```diff
@@ -45,6 +45,12 @@ void SimHarness::RunForMillis(size_t t) {
   }
 }
 
+void SimHarness::DelayForMillis(size_t t) {
```
obra, line 48: Is there a reason to have this wrapper rather than just calling the regular arduino "delay()" function, which defaults to ms? It looks like it's already there in the sim.
image


..and it looks like using micros may...not be all that useful right now:

image

gedankenexperimenter: You're absolutely correct.  For some reason, I was thinking that the parameter for `delay()` was in seconds, not milliseconds.  I'll fix it and force-push.

gedankenexperimenter: Also, I couldn't find the code for the simulator's `millis()` function when I looked for it.

gedankenexperimenter:
Updated without superfluous wrapper.  It's just a change to `ktest-to-cxx` now, and it works just as well.

testing/SimHarness.cpp
obra: Fair :)

(It's in Arduino.cpp in the virtual/ arduino core)

obra:
Looks great. thanks!

gedankenexperimenter:
Incidentally, I just found a bug in `delay()`; should I attach the fix for that to this PR?

obra:
sure, but as a stand-alone commit.

obra:
(I'm holding on merging this until you push that commit. If you'd rather this merge immediately, please click the merge button :)

gedankenexperimenter:
I still can't find the file where that code you referenced came from.  It doesn't seem to be in the Kaleidoscope repo.

obra:
It's not. The Arduino virtual isn't part of kaleidoscope proper. It's over
here:
https://github.com/keyboardio/Kaleidoscope-Bundle-Keyboardio/blob/master/virtual/cores/arduino/Arduino.c

gedankenexperimenter:
Ah, it's in the bundle repo.  I submit a PR to fix that separately

PR #1107, opened 2022-02-11
Fix Qukeys memory bounds violation

gedankenexperimenter:
This adds a check to `KeyAddrEventQueue::remove()` to ensure that it doesn't walk all over memory if it's called when the queue is empty.  Without this fix, it was possible for Qukeys to call it, causing the queue length to be set to 255, which would then result in a significant area of memory out of bounds for its `KeyAddrEventQueue` to get shifted one or two bytes, probably causing the keyboard to crash.


gedankenexperimenter:
I considered simply changing `KeyAddrEventQueue::length_` to a signed integer, but that resulted in a much bigger increase in PROGMEM size, while not being as robust a solution (the included testcase would still fail, it would just fail later).  The change to Qukeys isn't necessary to fix the bug, but it's still much better to not check an empty queue's timeout.

I'll see if I can trigger the bug with a testcase that actually uses Qukeys, but it's a bit trickier to do than the plugin I wrote for the testcase.

PR #1111, opened 2022-02-19
Distinguish between TapDance "tap" and "hold" timeouts

gedankenexperimenter:
We got a request via Discord for TapDance behaviour that would allow different `Key` values depending on whether the TapDance key was tapped or pressed and held (at any tap count).  This is a minimal change that enables such a configuration, by using `ActionType::Hold` if the TapDance key is still being held when it times out, and `ActionType::Timeout` if it has already been released.  The user `tapDanceAction()` function can then check which action type was sent, and call one of two different `tapDanceActionKeys()` functions to get the different key values.


gedankenexperimenter:
Note: The automated test failure was not because any testcases failed; they all passed when I ran the whole test suite.

obra:
Checks also pass when I reran them.

algernon (approved):

PR #1114, opened 2022-02-21
Restructure MouseKeys to be aware of Macros virtual keys

### LINKED ISSUE #1113: MouseKeys not working with Macros
**Describe the bug**
MouseKeys (mouseScrollUp, mouseScrollDn) do not work with Macros. While the sketch compiles just fine, nothing happens when a key triggering a macro using mouse keys is pressed. 

**To Reproduce**
Example implementation of macroAction:

``` c++
const macro_t* macroAction(uint8_t macroIndex, KeyEvent &event) {
  if (keyToggledOn(event.state)) {
    switch (macroIndex) {
      case MACRO_ZOOM_OUT:
        return MACRO(D(LeftControl), T(mouseScrollUp), U(LeftControl));
    }
  }
  return MACRO_NONE;
}
```
**Expected behavior**
The macro `MACRO_ZOOM_OUT` in the example code should send CTRL+MouseScrollUp to the OS, just as if I had pressed CTRL on my keyboard and scrolled up using my actual mouse wheel at the same time.

 - OS: Ubuntu Linux and Windows 10
 - Kaleidoscope Version 2022.2.18175535
 - Device: Keyboardio Model 01

gedankenexperimenter:
This change is designed to fix #1113 by restructuring how MouseKeys processes key events for mouse cursor and wheel movement.  Instead of only sending those reports in `afterEachCycle()` directly, I've shifted that code out to two new helper functions: `sendMouseMoveReport()` & `sendMouseWheelReport()`.  This allows us to both send periodic reports when a key is held, but also in response to key events directly, even when a mouse movement (or wheel) key toggles on and off within a single cycle.

However, because MouseKeys used to poll the keys of the keyboard to find out which direction to move the cursor (or wheel), it still wouldn't respond properly to mouse wheel keys (for example) that appear in Macros' virtual keys array.  So, rather than simply iterating through `live_keys[]` looking for mouse movement keys, we store the movement and wheel directions in persistent variables, and update those from a new `onAddToReport()` handler.  Last, because Macros virtual keys only call `Hooks::onAddToReport()` from the `beforeReportingState()`, we need to wait until after that to send the mouse move and wheel reports, so we can't send the reports from the `onKeyEvent()` handler, and doing it from `beforeReportingState()` would be plugin initialization order dependent, so I also added an `afterReportingState()` handler to MouseKeys to call the new move and wheel report functions.

I haven't tested this yet, and I don't even know if our automated testing infrastructure is set up to test mouse reports, so this is still a work in progress.

### DIFF
--- a/plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys.cpp
+++ b/plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys.cpp
@@ -37,6 +37,11 @@ uint16_t MouseKeys_::move_start_time_;
 uint16_t MouseKeys_::accel_start_time_;
 uint16_t MouseKeys_::wheel_start_time_;
 
+uint8_t MouseKeys_::move_direction_ = 0;
+uint8_t MouseKeys_::wheel_direction_ = 0;
+uint8_t MouseKeys_::pending_move_direction_ = 0;
+uint8_t MouseKeys_::pending_wheel_direction_ = 0;
+
 // =============================================================================
 // Configuration functions
 
@@ -105,75 +110,12 @@ EventHandlerResult MouseKeys_::afterEachCycle() {
   }
 
   // Check timeout for position update interval.
-  bool update_position = Runtime.hasTimeExpired(move_start_time_, speedDelay);
-  if (update_position) {
-    move_start_time_ = Runtime.millisAtCycleStart();
-    // Determine which mouse movement directions are active by searching through
-    // all the currently active keys for mouse movement keys, and adding them to
-    // a bitfield (`directions`).
-    uint8_t directions = 0;
-    int8_t vx = 0;
-    int8_t vy = 0;
-    for (Key key : live_keys.all()) {
-      if (isMouseKey(key) && isMouseMoveKey(key)) {
-        directions |= key.getKeyCode();
-      }
-    }
-
-    if (directions == 0) {
-      // If there are no mouse movement keys held, reset speed to zero.
-      MouseWrapper.accelStep = 0;
-    } else {
-      // For each active direction, add the mouse movement speed.
-      if (directions & KEY_MOUSE_LEFT)
-        vx -= speed;
-      if (directions & KEY_MOUSE_RIGHT)
-        vx += speed;
-      if (directions & KEY_MOUSE_UP)
-        vy -= speed;
-      if (directions & KEY_MOUSE_DOWN)
-        vy += speed;
-
-      // Prepare the mouse report.
-      MouseWrapper.move(vx, vy);
-      // Send the report.
-      Runtime.hid().mouse().sendReport();
-    }
-  }
+  if (Runtime.hasTimeExpired(move_start_time_, speedDelay))
+    sendMouseMoveReport();
 
   // Check timeout for scroll report interval.
-  bool update_wheel = Runtime.hasTimeExpired(wheel_start_time_, wheelDelay);
-  if (update_wheel) {
-    wheel_start_time_ = Runtime.millisAtCycleStart();
-    // Determine which scroll wheel keys are active, and add their directions to
-    // a bitfield (`directions`).
-    uint8_t directions = 0;
-    int8_t vx = 0;
-    int8_t vy = 0;
-    for (Key key : live_keys.all()) {
-      if (isMouseKey(key) && isMouseWheelKey(key)) {
-        directions |= key.getKeyCode();
-      }
-    }
-
-    if (directions != 0) {
-      // Horizontal scroll wheel:
-      if (directions & KEY_MOUSE_LEFT)
-        vx -= wheelSpeed;
-      if (directions & KEY_MOUSE_RIGHT)
-        vx += wheelSpeed;
-      // Vertical scroll wheel (note coordinates are opposite movement):
-      if (directions & KEY_MOUSE_UP)
-        vy += wheelSpeed;
-      if (directions & KEY_MOUSE_DOWN)
-        vy -= wheelSpeed;
-
-      // Add scroll wheel changes to HID report.
-      Runtime.hid().mouse().move(0, 0, vy, vx);
-      // Send the report.
-      Runtime.hid().mouse().sendReport();
-    }
-  }
+  if (Runtime.hasTimeExpired(wheel_start_time_, wheelDelay))
+    sendMouseWheelReport();
 
   return EventHandlerResult::OK;
 }
@@ -183,6 +125,9 @@ EventHandlerResult MouseKeys_::onKeyEvent(KeyEvent &event) {
   if (!isMouseKey(event.key))
     return EventHandlerResult::OK;
 
+  pending_move_direction_ = 0;
+  pending_wheel_direction_ = 0;
+
   if (isMouseButtonKey(event.key)) {
     sendMouseButtonReport(event);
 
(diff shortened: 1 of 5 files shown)


gedankenexperimenter:
I also wrote the code for an alternative solution, which required more extensive changes, adding two entirely new handlers so that plugins could search for active virtual keys (i.e. Macros), but I haven't submitted that yet.  This solution is slightly less robust, but I think it should be good enough to be reliable, without requiring those more extensive changes.  It should also be more efficient than either the existing MouseKeys or the version in that alternate solution.

gedankenexperimenter:
I have now tested this, and MouseKeys appears to work both on its own, and from within a Macro.

I included a second commit that makes a small optimization at the cost of a small (in my opinion) amount of code readability.  It only saves two bytes of RAM and four bytes of PROGMEM, so I'm not sure that it's worth it.

gedankenexperimenter:
While working on a more comprehensive MouseKeys rewrite (based on this code), I discovered that while this PR fixes mouse movement and wheel events in Macros code, it doesn't handle mouse buttons as well.  The solution for mouse buttons is the same, and I could lift it from that rewrite, or we could review the whole thing together, or…

gedankenexperimenter:
This PR should also get a testcase, now that I've worked out how to set up mouse report verification in the simulator.

gedankenexperimenter:
I have a testcase to add to this, once #1120 is merged.

obra:
#1120 is now merged

plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys.h
```diff
@@ -41,20 +41,34 @@ class MouseKeys_ : public kaleidoscope::Plugin {
...
   static uint16_t move_start_time_;
   static uint16_t accel_start_time_;
   static uint16_t wheel_start_time_;
 
+  // Mouse cursor and wheel movement directions are stored in a single bitfield
+  // to save space.  The low four bits are for cursor movement, and the high
+  // four are for wheel movement.
+  static constexpr uint8_t wheel_offset_ = 4;
+  static constexpr uint8_t wheel_mask_ = static_cast<uint8_t>(uint8_t(~0) << wheel_offset_);
+  static constexpr uint8_t move_mask_ = static_cast<uint8_t>(uint8_t(~0) ^ wheel_mask_);
```
obra, line 57: In theory, could  this be done more cleanly using a struct?

obra:
Other than the one question about the bitfield, this looks good to me.

plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys.h
gedankenexperimenter: Yeah, I'm not thrilled about all of the ridiculous gymnastics to get these bitmasks defined.  I could just write:
```c++
static constexpr uint8_t wheel_mask_ {0b11110000};
static constexpr uint8_t move_mask_  {0b00001111};
```
…and that would probably be clearer, even though it doesn't use the 4-bit offset defined above to construct them.

I don't see how a struct would help, though.  Did you have something specific in mind?

gedankenexperimenter:
Testcase now included.  It seems I wasn't quite right about the restructuring being necessary for the button keys, but it still reduces PROGMEM usage, and it will be necessary if my other planned MouseKeys changes get adopted.

plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys.h
gedankenexperimenter: I do not have a stable preference for either making it more unambiguous to humans or making it more rigidly defined for the compiler, so feel free to make the call on this.

obra: I tend to land pretty solidly on "making it clear to humans" over rigid definition for the compiler. There are typically only one or two compilers that need to get it right, but plenty of humans who need to deal with the code over its lifecycle.

gedankenexperimenter: If there was a clearer way of enforcing the relationship via code, with just the one constant (the 4-bit offset), and the other constants derived from that, I would do it.  This is one of those cases where I really wish I could write this in Rust:
```rust
const OFFSET: u8 = 4;
const MOVE_MASK: u8 = u8::MAX >> OFFSET;
const WHEEL_MASK: u8 = !MOVE_MASK;
```
…but C++ makes it much less clear.  I will update momentarily.

gedankenexperimenter: I forgot to mention that I fixed the bitmask constants to make them more human-readable.

plugins/Kaleidoscope-MouseKeys/src/kaleidoscope/plugin/MouseKeys.cpp
```diff
@@ -183,50 +124,85 @@ EventHandlerResult MouseKeys_::onKeyEvent(KeyEvent &event) {
...
+
+  // Reports for mouse cursor and wheel movement keys are sent from the
+  // `afterReportingState()` handler (when first toggled on) and
+  // `afterEachCycle()` handler (when held).  We need to return `OK` here so
+  // that processing of events for these keys will complete.
+  return EventHandlerResult::OK;
+}
+
+// -----------------------------------------------------------------------------
+EventHandlerResult MouseKeys_::afterReportingState(const KeyEvent &event) {
```
obra, line 148: In an evented world, "afterReportingState" does seem a more right time to act than "onKeyEvent", but it makes me feel like we should be doing the setup in "beforeReportingState" and then having the core actually send Mouse reports like we do for Keyboard, SystemControl and ConsumerControl.

To be clear: I don't think we want to make that heavy a change a precondition of getting this merged. 

I'd love to hear your thinking about why this is afterReportingState, rather than beforeReportingState or if it's essentially a no-op and you just picked one arbitrarily.

gedankenexperimenter: I'm pretty sure it could just as easily work in `beforeReportingState()`, but there is one advantage (or perhaps disadvantage) to doing it in `afterReportingState()` instead: the former is abortable (if a plugin returns `ABORT`, the others won't get called).

I think there's a reason we don't want the mouse reports sent in the core: if you don't have MouseKeys in your sketch, it might be able to optimize all the mouse stuff out, making the binary much smaller for people who have no use for mouse keys.  I haven't checked to be sure that actually happens, though.

I'm okay with changing it to `beforeReportingState()`, if you prefer it that way, as long as it works (I haven't tested it to confirm).

obra: I don't have a strong feeling there. I wanted to understand the rationale, though. - It's probably worth sticking that in a comment along with the code?

Indeed, I agree that we ought to be optimizing mouse stuff out if it's not used.

gedankenexperimenter: One more noteworthy thing: mouse movement and scroll reports get sent at times other than in response to a key toggling on or off.

obra:
@algernon - I'd love to have you glance at this quickly before it gets merged down

algernon (approved):

PR #1115, opened 2022-02-28
Remove deprecated code dated 2021-08-01

gedankenexperimenter:
This removes deprecated code with a removal date of August 2021.  Also removed: LEDControl code without an explicit deprecation date, but which was deprecated in May 2021.


gedankenexperimenter:
Note: `key_events.h` was removed.  That file was `#include`d by `Kaleidoscope.h`, so I added a couple of lines to that main header file to add `key_defs.h` and `keyswitch_state.h`, which were previously included via the removed file.  I didn't include all of the headers that were in `key_events.h`, because some of them were internal, and I don't think we need them in the main header file for Kaleidoscope.

gedankenexperimenter:
I found some more deprecated code that can be removed.  I'll add it to this PR shortly.

obra:
The big thing I'd like to see before merging this one is a doc change recording what was removed along with the notes about how to fix your code that was previously included in the deprecation notices.

gedankenexperimenter:
Right.  I'll go over the docs, and also separate changes into commits by plugin, to make it easy to revert specific parts quickly, should the need arise.

gedankenexperimenter:
Here's the big question that I don't feel I can answer, @obra: Where would you like the documentation about removed API functions and variables to go?  It's not at all obvious to me which things you want in NEWS.md, which things in UPGRADING.md, and where to put them in either document.  Or if you'd prefer (some of) this info in other places, such as plugins' README.md files.

obra:
That all makes sense. Do you have thoughts on the least painful way to get
us there?

gedankenexperimenter:
So, it turns out there was a pretty clear place to put API removals in UPGRADING.md, so I followed the existing pattern, and updated each commit appropriately.  I also included a deprecation of some code for Leader, and fixing a couple of typos in that section of UPGRADING.

I still want to reorganize the NEWS and UPGRADING docs as discussed, but I think it's best to save that for later.

obra:
>I still want to reorganize the NEWS and UPGRADING docs as discussed, but I think it's best to save that for later.

works for me.

PR #1118, opened 2022-03-02
Multiple fixes for library search paths.

### LINKED ISSUE #1116: Kaleidoscope's build system sticks our parent dir in the arduino search path. It shouldn't.
```First up, the reason that "things outside the Kaleidoscope dir are getting included"
Arduino needs to be able to find the Kaleidoscope library. To do that, we need to tell it that Kaleidoscope is in a directory that is a search path.
The way we have done so is to stick its parent dir in the list. This is clearly bad and wrong.
I suppose one way to fix it is to construct a path that has only a single symlink inside it and stick that in the search path.
I'm not in love with that idea, but it'll get us the isolation we need.
Let me make a ticket for that now before I look at the next step```

Reported by @gedankenexperimenter in the thread including https://canary.discord.com/channels/492408953041321984/504378238647402507/948651561842802709

### LINKED ISSUE #1117: Instructions for development on KeyboardioHID (or similar dependencies)
We currently don't have a well-defined procedure for how to set up a build environment for testing changes to KeyboardioHID (or any other external libraries that a Kaleidoscope build depends on).  Once that procedure exists, we should document it somewhere in the Kaleidoscope repo.

[@obra — feel free to rename this issue if you think of something more suitable.]

### PR DESCRIPTION
**Pull request by obra**
Switch to including Kaleidoscope directly, rather than its parent directory.
Add a new environment variable to add 'override' libraries.

Fixes #1116 
Fixes #1117

### DIFF
--- a/etc/makefiles/sketch.mk
+++ b/etc/makefiles/sketch.mk
@@ -47,12 +47,6 @@ HEX_FILE_WITH_BOOTLOADER_PATH 	:= $(OUTPUT_PATH)/$(OUTPUT_FILE_PREFIX)-with-boot
 ELF_FILE_PATH 			:= $(OUTPUT_PATH)/$(OUTPUT_FILE_PREFIX).elf
 LIB_FILE_PATH 			:= $(OUTPUT_PATH)/$(OUTPUT_FILE_PREFIX).a
 
-
-KALEIDOSCOPE_PLATFORM_LIB_DIR := $(abspath $(KALEIDOSCOPE_DIR)/..)
-
-
-
-
 ifeq ($(FQBN),)
 possible_fqbns = $(shell $(ARDUINO_CLI) board list --format=json |grep FQBN| grep -v "keyboardio:virtual"|cut -d: -f 2-)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: etc/makefiles/sketch.mk [hunk 1]
```diff
@@ -167,9 +167,20 @@ else
 local_cflags_property =
 endif
 
-compile:
+# If you set KALEIDOSCOPE_LOCAL_LIB_DIR to the name of a directory, 
+# all of the Arduino libraries inside that directory should be used 
+# in preference to any library with the same name further dow the search path
+
+ifdef ($(KALEIDOSCOPE_LOCAL_LIB_DIR))
```
gedankenexperimenter, line 174: Two minor errors here. It should be:
```
ifneq ($(KALEIDOSCOPE_LOCAL_LIB_DIR),)
_arduino_local_libraries_prop =  --libraries "${KALEIDOSCOPE_LOCAL_LIB_DIR}"
```
(`ifneq`, not `ifdef`, and `--libraries`, not `-libraries`)

gedankenexperimenter: Wit those minor adjustments, it all works well for me.  Thanks!

obra:
Thank you! Fixed and merging.

PR #1119, opened 2022-03-03
When working from the command line, use git cores exclusively. 

obra:
Instead of installing both prebuilt and git cores when working from the commandline, install the devtools using the package manager and then get the arduino cores exclusively with git.


obra:
Based on @gedankenexperimenter's testing, I'm going to merge this down.

PR #1121, opened 2022-03-04
Remove `static` declaration from OneShot variables and functions

gedankenexperimenter:
In the past, we declared most plugin variables and functions `static` because it resulted in a smaller binary, taking up less PROGMEM space by a significant margin.  After some experimentation, I think I've determined that a lot of that savings can be had by marking functions `const` where possible, indicating that those functions do not alter the object's internal state.

I feel that removing all the `static` declarations and adding `const` where appropriate makes the code clearer, and look more like "standard" C++ code.  In this case, on the Atreus example sketch, it actually results in a smaller PROGMEM footprint (by two bytes) and also saves one byte of RAM.  In other plugins, there do seem to be larger savings to keeping certain variables `static`, but those should probably be treated as the exception, not the rule.

This PR also includes a fix to `KeyAddrBitField`, which was not quite `const` correct.  That was necessary to get the new version of OneShot to compile.

I would like to give other plugins the same treatment, eventually (and possibly do a general code style correction on them), but I think it's useful to treat this PR as a kind of pilot project for that.

### DIFF
--- a/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.h
+++ b/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.h
@@ -191,12 +191,12 @@ class OneShot : public kaleidoscope::Plugin {
   uint16_t hold_timeout_{250};
   int16_t double_tap_timeout_{-1};
 
-  // --------------------------------------------------------------------------
-  // State variables
   uint16_t stickable_keys_{uint16_t(-1)};
   bool auto_modifiers_{false};
   bool auto_layers_{false};
 
+  // --------------------------------------------------------------------------
+  // State variables
   KeyAddrBitfield temp_addrs_;
   KeyAddrBitfield glue_addrs_;
 
--- a/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.h
+++ b/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.h
@@ -187,21 +187,21 @@ class OneShot : public kaleidoscope::Plugin {
 
   // --------------------------------------------------------------------------
   // Configuration variables
-  uint16_t timeout_{2500};
-  uint16_t hold_timeout_{250};
-  int16_t double_tap_timeout_{-1};
+  uint16_t timeout_ = 2500;
+  uint16_t hold_timeout_ = 250;
+  int16_t double_tap_timeout_ = -1;
 
-  uint16_t stickable_keys_{uint16_t(-1)};
-  bool auto_modifiers_{false};
-  bool auto_layers_{false};
+  uint16_t stickable_keys_ = -1;
+  bool auto_modifiers_ = false;
+  bool auto_layers_ = false;
 
   // --------------------------------------------------------------------------
   // State variables
   KeyAddrBitfield temp_addrs_;
   KeyAddrBitfield glue_addrs_;
 
-  uint16_t start_time_{0};
-  KeyAddr prev_key_addr_{invalid_key_addr};
+  uint16_t start_time_ = 0;
+  KeyAddr prev_key_addr_ = invalid_key_addr;
 
   // --------------------------------------------------------------------------
   // Internal utility functions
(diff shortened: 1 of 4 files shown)


gedankenexperimenter:
I forgot to mention: One of the big reasons I prefer variables being non-`static` is that default/initial values are visible in the header files where they're declared, rather than being "hidden" in the cpp files.  The other is that it's easy to see at a glance which functions promise not to alter the internal state of the plugin (because they're marked `const`).

gedankenexperimenter:
I accidentally broke one testcase.  I was too aggressive with the use of the `always_inline` gcc attribute, which made it impossible for a user override of `isStickable()` to call `isStickableDefault()`.  Removing the `always_inline` attribute only increases PROGMEM by four bytes, so now this PR is +2 bytes PROGMEM, -1 byte RAM.

plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.h
```diff
@@ -187,39 +187,39 @@ class OneShot : public kaleidoscope::Plugin {
 
   // --------------------------------------------------------------------------
   // Configuration variables
-  static uint16_t timeout_;
-  static uint16_t hold_timeout_;
-  static int16_t double_tap_timeout_;
+  uint16_t timeout_{2500};
```
obra, line 190: Generally, I really like the idea behind these changes, but I'd strongly prefer that we stick with more traditional looking initializations for simple values.

gedankenexperimenter: I prefer brace initialization in C++ for all of the reasons listed here: https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives

In C++, it's just the normal way to initialize a variable, it's not some oddball non-traditional alternative way of doing it.  If nothing else, a clear distinction between initialization and assignment is something that I find useful.  What's your reason for preferring the C syntax?

obra: Mostly, readability/human parseability.  I'm coming from a background of having done a lot more work in C-like languages than C++. Braced initialization reads to me like "this is a function returning a static value" - I can learn and adapt, but reading code with brace-style initialization, I have to think harder about what's going on.   Not allowing narrowing *is* a win.

Having gone and poked at the coding style guidelines as currently written, they do say "your choice"  - This isn't a hill I'm going to die on.  I do care about a given chunk of code not mixing multiple styles of initializer, but that's not what's up here.

One of the new initializers, I don't quite understand: (I'm not objecting to it, I just don't understand what it's doing.) 

`  uint16_t stickable_keys_{uint16_t(-1)};`

What's that inner uint16_t do?

gedankenexperimenter: > What's that inner uint16_t do?

Without it, there's a narrowing error, because `-1` is a signed integer.  `uint16_t(-1)` is (I believe, but I could be technically incorrect) a constructor call (but maybe it's a cast operator).  Functionally, it's converting a signed `int` to a `uint16_t`.  If we used old-school initialization, we could drop it.  I do wish there was a better concise way of getting an unsigned integer with all ones in C++.

I do want to convince you that brace initialization is a better standard for C++, but if you still don't like it, I'd prefer it if we used one style consistently throughout the codebase, even if it's not the one I think is best.

> Braced initialization reads to me like "this is a function returning a static value"…

One of the reasons it was adopted in the first place was apparently a phenomenon known as the "most vexing parse", where the compiler recognized something as a function declaration (without its definition) when the programmer intended a constructor.  This was solved by using braces instead of parens for the constructor call to disambiguate.

I reiterate: I'm hoping to convince you to change your mind on which initialization form you want for Kaleidoscope code.  I'm not trying to convince you to allow _me_ to use this style for the code that I contribute.  I will change it if that's what you decide (or I'll leave it if you decide that a mix is fine).  I know we've exchanged ideas about this briefly(?) in the past, too, so I don't intend to keep pressing the issue.

PR #1122, opened 2022-03-04
Fix docker simulator tests

gedankenexperimenter:
With the reorganization of arduino-cli libraries in .arduino, we no longer use a symlink for Kaleidoscope, so don't try to delete it or create it inside the Docker container.


gedankenexperimenter:
Hmm.  This doesn't seem to be working as I expected.  After wiping `.arduino` and re-running `make setup` from master, running `make docker-clean` fails, but with this patch, it succeeds.  However, I get several instances of this error:
```
Error initializing instance: Loading index file: loading json index file /arduino-cli/data/package_kaleidoscope_devel_index.json: open /arduino-cli/data/package_kaleidoscope_devel_index.json: no such file or directory
```
Any ideas?

PR #1125, opened 2022-03-06
docker-bash now mounts Kaleidoscope live and interactive

obra:


gedankenexperimenter:
I've tried this out, and noticed some counterintutive results.  Compared to `master`, I don't see any consistent difference in the time it takes to run a test, either for the first time or for subsequent runs.  What's really surprising is that I'm consistently measuring the time it takes to run a test from the shell inside the docker container at about double the length of time it takes to run the same test from outside via `make docker-simulator-tests`.  This is the opposite of what I would expect, and makes the primary _raison d'être_ of this PR moot (unless we can "fix" it).  Being able to share files from the two domains could certainly be helpful in some situations, though.

obra:
With the most recent change, docker-bash may not be faster than setting up a new docker  runtime every time we want to run a command, but it does enable the workflow of "hack on code and test interactively in docker" better than before, so I'm going to merge.  (Happy to take further bug reports / continue to iterate on this)

PR #1127, opened 2022-03-07
Fix setStickyColor() typo

Simon-Claudius:
Just found this little typo while flashing: `setOnestickyColor` should be `setStickyColor`


obra:
Technically, merging this will 'break backward compatibility', but given the context, I think it's the right thing to do.

PR #1129, opened 2022-03-10
Update MouseKeys testcase for current KeyboardioHID

gedankenexperimenter:
With keyboardio/KeyboardioHID#82, we no longer expect a no-op mouse report after each mouse cursor move report.


gedankenexperimenter:
This PR should pass the MouseKeys testcase once the bundle is updated to point to the newest KeyboardioHID.

PR #1130, opened 2022-03-11
MouseKeys "janitorial" work

gedankenexperimenter:
This is a bundle of changes to MouseKeys that are mostly internal, mainly aimed at conformity to coding style standards for Kaleidoscope.  The most significant change from a user perspective is the deprecation of public configuration variables in favor of setter methods.  Other changes include:

- Removal of trailing underscores from class names
- Changing from `static` to member variables
- Moving the `MouseWrapper` helper class into the `kaleidoscope::plugin::mousekeys` namespace
- Replacing the term "move" with "cursor" to better match "button" and "wheel"
- Various minor odds and ends

This is not an exhaustive cleanup of the MouseKeys code, in part because I'm working on more extensive functional changes to the plugin, but it does make a number of minor improvements to the code.  It should be functionally identical to master while saving a few bytes of PROGMEM and RAM (48 and 2, respectively in the Atreus example sketch).


gedankenexperimenter:
This will pass the test suite when #1129 gets merged.

obra:
There are a couple of things in here that feel like they want a little bit of discussion. Would your preference be for me to cherry pick everything else and for you to move those couple of items to the rewrite or to talk them out now? (In specific: The deprecations don't feel quite like 'janitorial' changes to me and 'cursor' isn't feeling right, but I don't have a solid 'we should do X instead' proposal yet)

gedankenexperimenter:
The "move" _vs_ "cursor" naming certainly warrants discussion.  I think the deprecation of public member variables is merely part of bringing the plugin up to current standards.

I'm okay with doing this whichever way you like, but it *feels* like it would be less work for me to do the discussion first, because I'm going to want to rebase the other one on whatever gets merged into master, and it *seems* like less work to deal with the rebase conflicts just once for that other branch.  I'm quite likely wrong about that, though, so I'm not going to push hard for doing it one way or the other.

To the actual issue: "move" _vs_ "cursor".  My preference, obviously, is to name the three types of (non-absolute) mouse reports after the objects being acted upon: "button", "cursor", and "wheel".  An alternative is action verbs, but though "move" obviously applies to "cursor" and "scroll" applies to "wheel", "button" needs multiple verbs: "press" and "release".  Of course, my preferred terminology isn't perfect, either: "button" and "wheel" are things that the user's hand touches, whereas "cursor" is a thing on the screen.

Either of the above set of terms (all nouns or all verbs) is okay by me, and I'd even be okay with "button", "move", and "scroll", since the button report is fundamentally different from the other two (a no-op for buttons is not necessarily all zeros).  I think the main reason for my preference is that it reduces ambiguity: "speed" alone doesn't clearly tell me if it applies to the cursor or the scroll wheel, but "cursor speed" and "wheel speed" leave me with no question about what we're talking about.

Now that I think about it, even "button", "cursor", and "scroll" would be a better set of terms than what we have now in terms of clarity.

All that being said, I'm okay with removing the deprecations and those name changes from this PR, especially because I'd like to change the whole configuration API of the plugin to something that has more useful parameters.

obra:
I agree that move needs renaming. “cursor” doesn’t feel quite right. Position? Movement? Pointer?

obra:
In terms of workflow, I’d be happy to to rebase the branch to move the commits that may need more discussion to the end, so that the new branch could still depend on the same prior state.

gedankenexperimenter:
> In terms of workflow, I’d be happy to to rebase the branch to move the commits that may need more discussion to the end, so that the new branch could still depend on the same prior state.

The new branch is already disconnected from this one.  Don't worry about it; I'll take care of it.

gedankenexperimenter:
> I agree that move needs renaming. “cursor” doesn’t feel quite right. Position? Movement? Pointer?

I thought "mouse cursor" was a fairly universal formal term for the thing on the screen that's getting moved around, "mouse pointer" being a less formal version.  The former is used in CSS, for example.  It also doesn't suffer from the problem of being confused with other kinds of "pointers" when looking at C++ code.  "Movement speed" is a bit ambiguous, since there are other types of "movement" (scrolling), and "position speed" sounds confusing.

gedankenexperimenter:
Force-pushed without the deprecations and name change.  You've convinced me that those things belong with the rewrite, not this PR.

PR #1131, opened 2022-03-14
Force `EventHandlerResult` values to be a single byte

gedankenexperimenter:
By specifying the type of the `EventHandlerResult` enum as `uint8_t`, it only takes up one byte instead of two, saving a significant amout of PROGMEM.


obra:
Nice catch.

gedankenexperimenter:
I was working on making the MouseKeys plugin configurable via Focus and EEPROM (i.e. Chrysalis), and noticed this in the process of making that more efficient.

Incidentally, this is failing the MouseKeys testcase because #1129 still hasn't been merged.

PR #1133, opened 2022-03-19
fix minor typo in plugin author guide

SylvainJuge:
Self-explanatory PR, only one character away and for once the compiler helped me find what was broken :-).


obra:
Thanks!

PR #1138, opened 2022-03-21
Reorganize header includes with help from `include-what-you-use`

gedankenexperimenter:
This consists of a reorganization and standardization of the header includes for almost all Kaleidoscope source files.  I used [`include-what-you-use`](https://include-what-you-use.org/) via a wrapper script to do most of the work.  In some places I manually added `IWYU pragma` comments to keep headers that it would have discarded, or to mark header files as "private", with alternatives (e.g. including `KeyAddr.h` instead of `MatrixAddr.h` in most places, and including `device/device.h` instead of `Virtual.h` (the hardware header used when running iwyu).

There are some related changes bundled with the header reorg:
- I updated the code style guide with information about the header includes.
- I added the `iwyu.py` wrapper script that we can use on most source files (but unfortunately not all).
- I replaced some Arduino-specific type aliases that had been missed (`boolean` & `byte`), so that those files can include `stdint.h` instead of `Arduino.h`.
- I changed some plugin-specific `Key` constructors from preprocessor macros to `constexpr` functions or variables.
- I replaced some instances of `::Kaleidoscope` with the preferred `::kaleidoscope::Runtime`.
- I attempted to better document the `kaleidoscope/device/device.h` header file to make it clear how it is used.

Files that warrant particular attention:
- `src/kaleidoscope/device/device.h`
- `src/kaleidoscope/hooks.h`

Closes #762

### DIFF
--- a/plugins/Kaleidoscope-Cycle/src/kaleidoscope/plugin/Cycle.h
+++ b/plugins/Kaleidoscope-Cycle/src/kaleidoscope/plugin/Cycle.h
@@ -17,8 +17,15 @@
 
 #pragma once
 
-#include "kaleidoscope/Runtime.h"
-#include <Kaleidoscope-Ranges.h>
+#include <Kaleidoscope-Ranges.h>                // for CYCLE
+#include <stdint.h>                             // for uint8_t
+
+#include "Arduino.h"                            // for PROGMEM
+#include "kaleidoscope/KeyAddr.h"               // for KeyAddr
+#include "kaleidoscope/KeyEvent.h"              // for KeyEvent
+#include "kaleidoscope/event_handler_result.h"  // for EventHandlerResult
+#include "kaleidoscope/key_defs.h"              // for Key
+#include "kaleidoscope/plugin.h"                // for Plugin
 
 constexpr Key Key_Cycle = Key(kaleidoscope::ranges::CYCLE);
 
--- a/plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
+++ b/plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
@@ -14,9 +14,23 @@
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "Kaleidoscope-DynamicMacros.h"
-#include "Kaleidoscope-FocusSerial.h"
-#include "kaleidoscope/keyswitch_state.h"
+#include "kaleidoscope/plugin/DynamicMacros.h"
+
+#include <Arduino.h>                                // for delay, PSTR, strc...
+#include <Kaleidoscope-FocusSerial.h>               // for Focus, FocusSerial
+#include <Kaleidoscope-Ranges.h>                    // for DYNAMIC_MACRO_FIRST
+
+#include "kaleidoscope/KeyAddr.h"                   // for KeyAddr
+#include "kaleidoscope/KeyEvent.h"                  // for KeyEvent
+#include "kaleidoscope/Runtime.h"                   // for Runtime, Runtime_
+#include "kaleidoscope/device/device.h"             // for VirtualProps::Sto...
+#include "kaleidoscope/keyswitch_state.h"           // for INJECTED, IS_PRESSED
+#include "kaleidoscope/plugin/EEPROM-Settings.h"    // for EEPROMSettings
+
+// This is a special exception to the rule of only including a plugin's
+// top-level header file, because DynamicMacros doesn't depend on the Macros
+// plugin itself; it's just using the same macro step definitions.
+#include "kaleidoscope/plugin/Macros/MacroSteps.h"  // for MACRO_ACTION_END
 
 namespace kaleidoscope {
 namespace plugin {
--- a/plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
+++ b/plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
@@ -176,7 +176,7 @@ void DynamicMacros::play(uint8_t macro_id) {
     case MACRO_ACTION_STEP_TAP_SEQUENCE: {
       while (true) {
         key.setFlags(0);
-        key.setKeyCode(pgm_read_byte(pos++));
+        key.setKeyCode(Runtime.storage().read(pos++));
         if (key == Key_NoKey)
           break;
         tap(key);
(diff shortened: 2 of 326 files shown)


gedankenexperimenter:
The current version failed on one of the example sketches, but they all passed when I ran them a few minutes ago, on the same code…

gedankenexperimenter:
I'm pretty sure I'm all done making changes to the commit history now, unless there are things someone's unhappy with.

plugins/Kaleidoscope-Redial/src/kaleidoscope/plugin/Redial.h
```diff
@@ -20,7 +20,7 @@
 #include "kaleidoscope/Runtime.h"
 #include <Kaleidoscope-Ranges.h>
 
-#define Key_Redial Key(kaleidoscope::ranges::REDIAL)
+constexpr Key Key_Redial = kaleidoscope::Key(kaleidoscope::ranges::REDIAL);
```
obra, line 27: The "kaleidoscope::Key" in this commit doesn't match the patterns in other similar header file changes. Was that intentional?

plugins/Kaleidoscope-Cycle/src/kaleidoscope/plugin/Cycle.h
```diff
@@ -20,7 +20,7 @@
 #include "kaleidoscope/Runtime.h"
 #include <Kaleidoscope-Ranges.h>
 
-#define Key_Cycle Key(kaleidoscope::ranges::CYCLE)
+constexpr Key Key_Cycle(kaleidoscope::ranges::CYCLE);
```
obra, line 30: Should this be:

constexpr Key Key_Cycle = Key(kaleidoscope::ranges::CYCLE);

?

plugins/Kaleidoscope-DynamicTapDance/src/kaleidoscope/plugin/DynamicTapDance.cpp
```diff
@@ -41,7 +41,7 @@ void DynamicTapDance::updateDynamicTapDanceCache() {
   map_[0] = 0;
 
   while (pos < storage_base_ + storage_size_) {
-    uint16_t raw_key = Kaleidoscope.storage().read(pos);
+    uint16_t raw_key = Runtime.storage().read(pos);
```
obra, line 53: I'm ok with this here, especially since we were mixing the two, but want to be careful about Kaleidoscope->Runtime changes, since we encourage user code to use the 'Kaleidoscope' name

src/kaleidoscope/driver/led/ws2812/config.h
```diff
@@ -43,6 +43,11 @@
 #define w_fixedhigh   4
 #define w_fixedtotal  8
 
+/* define CPU frequency in Mhz here if not defined in Makefile
```
obra, line 46: This should get pulled to its own commit.

src/kaleidoscope/driver/led/WS2812.h
```diff
@@ -33,8 +33,10 @@
 
 #pragma once
 
-#include "kaleidoscope/hardware/avr/pins_and_ports.h"
+#ifndef KALEIDOSCOPE_VIRTUAL_BUILD
```
obra, line 36: virtual build guards should get pulled to their own commits.

src/kaleidoscope/version.h
```diff
@@ -1,5 +1,12 @@
+// -*- mode: c++ -*-
+
 #pragma once
 
+// We use `include-what-you-use`, which uses `clang` to help manage header
```
obra, line 5: This is good. And should be its own commit.

plugins/Kaleidoscope-Redial/src/kaleidoscope/plugin/Redial.h
gedankenexperimenter: No, it wasn't.  Thanks.  It is, of course, the same thing, since `::Key` is a typedef of `::kaleidoscope::Key`, but I _meant_ to do all of them in the same style.

plugins/Kaleidoscope-DynamicTapDance/src/kaleidoscope/plugin/DynamicTapDance.cpp
gedankenexperimenter: That's what I thought: User code in sketches uses `::Kaleidoscope`, but internal (and plugin?) code uses `::kaleidoscope::Runtime`.  If you'd rather have plugin code also use the `::Kaleidoscope`, that's okay with me, but I'd prefer to use `Runtime` in plugins, since that's the name that matches the file that's being included.  Does that make sense?

src/kaleidoscope/driver/led/ws2812/config.h
gedankenexperimenter: Yikes!  I think it should be removed, actually.  I put in in there so that some file(s) would get successfully processed by IWYU, with the intention of it being temporary.  Sorry!

plugins/Kaleidoscope-DynamicTapDance/src/kaleidoscope/plugin/DynamicTapDance.cpp
obra: I think that us standardizing on Runtime in our plugins is right.

src/kaleidoscope/driver/led/WS2812.h
gedankenexperimenter: I don't recall why I added this guard here.  I'll try simply removing it and see what happens.  It might be necessary to make the new header include changes work, though, and if that's true, I think it belongs in the same commit as those header changes.

plugins/Kaleidoscope-DynamicMacros/src/kaleidoscope/plugin/DynamicMacros.cpp
```diff
@@ -176,7 +190,7 @@ void DynamicMacros::play(uint8_t macro_id) {
     case MACRO_ACTION_STEP_TAP_SEQUENCE: {
       while (true) {
         key.setFlags(0);
-        key.setKeyCode(pgm_read_byte(pos++));
+        key.setKeyCode(Runtime.storage().read(pos++));
```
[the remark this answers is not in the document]
gedankenexperimenter: You're absolutely right.  I actually meant to move this change to a separate PR, because it's only related in that I discovered it by accident (IWYU failed because of it, though I'm still not sure why it fails, but other builds succeeded).

gedankenexperimenter:
I'll be busy most of the afternoon, but I can fix up all the things you pointed out this evening.  If you would prefer the DynamicMacros fix to be submitted in a separate PR, I'm happy to do that.  I can also just leave it in a separate commit in this PR, which would be easier for testing purposes.  Whichever you prefer is fine with me.

src/kaleidoscope/driver/led/WS2812.h
gedankenexperimenter: The simulator tests run just fine without this, so it might be that it's helpful for IWYU, but I'd rather not go searching for why I had it in there right now, so I'm planning to leave it out.

gedankenexperimenter:
Rebased on master, with things that didn't belong pulled out into separate commits, including the DynamicMacros fix, which I plan to leave in this PR unless @obra prefers otherwise.

PR #1141, opened 2022-03-24
Standardize namespace block closing comments

gedankenexperimenter:
This standardizes namespace closing brackets for namespace blocks.  Each one is on its own line, with a comment clearly marking which namespace it closes. Consecutive lines closing namespace blocks have no whitespace between them, but there is one blank line before and after a set of namespace block closing lines.

There should be nothing by comments and whitespace changes in this PR.  I've checked it to verify that, but it changes so many files that someone else should definitely have a look, as well.

### DIFF
--- a/plugins/Kaleidoscope-Colormap/src/kaleidoscope/plugin/Colormap.cpp
+++ b/plugins/Kaleidoscope-Colormap/src/kaleidoscope/plugin/Colormap.cpp
@@ -72,7 +72,7 @@ EventHandlerResult ColormapEffect::onFocusEvent(const char *command) {
                                            map_base_, max_layers_);
 }
 
-}
-}
+}  // namespace plugin
+}  // namespace kaleidoscope
 
 kaleidoscope::plugin::ColormapEffect ColormapEffect;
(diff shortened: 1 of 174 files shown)


obra:
If this was done automatically with a tool, would you mind sticking the command you ran into the commit message?

gedankenexperimenter:
Good point.  It wasn't really done automatically, but I'll summarize what I did both here an in the commit message:

To generate the namespace comments, I used clang-format, with `FixNamespaceComments: true`.  But since clang-format can't exactly duplicate our astyle formatting, it made lots of other changes, too.  To isolate the namespace comments from the other formatting changes, I first ran clang-format with `FixNamespaceComments: false`, committed those changes, then ran it again to generate the namespace comments.  Then I stashed the namespace comments, reset `HEAD` to remove the other changes, applied the stashed namespace comments, and committed the results (after examining them and making a few minor adjustments by hand).

gedankenexperimenter:
I didn't actually run this script, but I made a [gist that summarizes what I did](https://gist.github.com/gedankenexperimenter/258d26d36085aa408c3a4f23dd839b76).

obra:
oof. that's a hassle. but a clever solution.

gedankenexperimenter:
Yeah.  It would be nice if it could just do the one bit of formatting requested (_maybe_ it can, but I couldn't figure out how…?), but I was pretty proud of the trick I came up with when I thought of it. 😁

PR #1143, opened 2022-03-25
Switch from `astyle` to `clang-format` for code formatting

gedankenexperimenter:
I expect this PR will need some further adjustments, so I'm submitting it as a draft for now.

This adds a script for calling `clang-format`, configuration files for it, and updates to the makefile and github workflows to run `make format` instead of `make astyle` and `make check-formatting` instead of `make check-astyle`.

Compared to `astyle`, `clang-format` is quite agressive about formatting things either one way or another; it doesn't tend to leave code alone.  This results in more uniformity across the codebase, but also limits options somewhat.  There were a few things that I wanted to do, but couldn't find a configuration option set that worked (in some cases, I think a future version of `clang-format` will support those style options, but the version I've got on my machine (13) doesn't recognize them, so they're commented out in the `.clang-format` file.  I don't know what version is available for our Ubuntu image.

A few things I think we're likely to want to change:

- I left the line length (`ColumnLimit`) turned off, because it seems to minimize the changes it makes to our existing code.  I think we probably want to set it to some reasonable value, but maybe it's best to leave it as-is.
- I configured it to align consecutive assignments.  In some places, this is a big help, but in others, it's a minor nuisance.  The (I think) desirable (`AlignCompound`) option doesn't seem to work as advertised by the docs, so it's commented out for now.
- Consecutive declarations are not aligned.  I think I like it better with this turned on, but not in places where one of the type names is much longer than others.  I wish there was finer-grained control.


gedankenexperimenter:
Two other options worthy of discussion: pointer and reference alignment:
```c++
LedMode* led_mode_p;  // `PointerAlignment: Left`
LedMode *led_mode_p;  // `PointerAlignment: Right`
// and:
KeyEvent& event;  // `ReferenceAlignment: Left`
KeyEvent &event;  // `ReferenceAlignment: Right`
```
The current style guide says either is fine, and clang-format can be configured to derive the alignment for a particular file from its content (`DerivePointerAlignment: true`), and to use the same alignment for references as pointers (`ReferenceAlignment: Pointer`), but there's no "agnostic" option that will just leave it alone, so a default must be specified, which is currently derived from Google, which is:
```yaml
DerivePointerAlignment: true
PointerAlignment: Left
ReferenceAlignment: Pointer
```
I think most of the code tends to use `Right` alignment now.  It would be easy to add in whichever is decided (if we do end up switching to clang-format).

plugins/Kaleidoscope-LED-AlphaSquare/src/kaleidoscope/plugin/LED-AlphaSquare.h
```diff
@@ -17,30 +17,26 @@
...
-               p00, p01, p02, p03,                                \
-               p10, p11, p12, p13,                                \
-               p20, p21, p22, p23,                                \
-               p30, p31, p32, p33)                               \
-  (uint16_t) (                                                    \
-              p00 <<  0 | p01 <<  1 | p02 <<  2 | p03 <<  3 |     \
-              p10 <<  4 | p11 <<  5 | p12 <<  6 | p13 <<  7 |     \
-              p20 <<  8 | p21 <<  9 | p22 << 10 | p23 << 11 |     \
-              p30 << 12 | p31 << 13 | p32 << 14 | p33 << 15 )
+#define SYM4x4(                                                                   \
```
obra, line 28: This may want some kind of guard comment

plugins/Kaleidoscope-LED-AlphaSquare/src/kaleidoscope/plugin/LED-AlphaSquare/Symbols.h
```diff
@@ -27,10 +27,7 @@ namespace alpha_square {
 namespace symbols {
 
 /* λ */
-static constexpr uint16_t Lambda = SYM4x4(1, 0, 0, 0,
-                                          0, 1, 0, 0,
-                                          0, 1, 1, 0,
-                                          1, 0, 0, 1);
+static constexpr uint16_t Lambda = SYM4x4(1, 0, 0, 0, 0, 1, 0, 0, 0, 1, 1, 0, 1, 0, 0, 1);
```
obra, line 30: needs a guard comment

plugins/Kaleidoscope-LED-Wavepool/src/kaleidoscope/plugin/LED-Wavepool.cpp
```diff
@@ -35,26 +35,85 @@
...
+uint16_t WavepoolEffect::idle_timeout = 5000;                         // 5 seconds
+int16_t WavepoolEffect::ripple_hue    = WavepoolEffect::rainbow_hue;  // automatic hue
 
 // map native keyboard coordinates (16x4) into geometric space (14x5)
 PROGMEM const uint8_t WavepoolEffect::TransientLEDMode::rc2pos[Runtime.device().numKeys()] = {
-  0,  1,  2,  3,  4,  5,  6,     59, 66,    7,  8,  9, 10, 11, 12, 13,
-  14, 15, 16, 17, 18, 19, 34,    60, 65,   35, 22, 23, 24, 25, 26, 27,
-  28, 29, 30, 31, 32, 33, 48,    61, 64,   49, 36, 37, 38, 39, 40, 41,
-  42, 43, 44, 45, 46, 47,     58, 62, 63, 67,    50, 51, 52, 53, 54, 55,
+  0,
```
obra, line 47: Guard comment needed here.

plugins/Kaleidoscope-LED-Wavepool/src/kaleidoscope/plugin/LED-Wavepool.cpp
```diff
@@ -162,11 +219,7 @@ void WavepoolEffect::TransientLEDMode::update(void) {
         uint8_t offset = (y * WP_WID) + x;
 
         int16_t value;
-        int8_t offsets[] = { -WP_WID,    WP_WID,
-                             -1,         1,
-                             -WP_WID - 1, -WP_WID + 1,
-                             WP_WID - 1,  WP_WID + 1
-                           };
+        int8_t offsets[] = {-WP_WID, WP_WID, -1, 1, -WP_WID - 1, -WP_WID + 1, WP_WID - 1, WP_WID + 1};
```
obra, line 222: guard comment?

bin/format-code.sh
```diff
@@ -0,0 +1,16 @@
...
+
+find ./* -type f \( -name '*.h' -o -name '*.cpp' \) \
+     -not \( -path './testing/googletest/*' -o -name 'generated-testcase.cpp' \) \
+     -print0 > "${TMP_SOURCES_FILE}"
+
+xargs -0 clang-format -i \
+      < "${TMP_SOURCES_FILE}"
+
+# shellcheck disable=SC2016
+xargs -0 perl -pi -e 's/^(__attribute__\(\(\w*\)\)) (\w)/$1\n$2/' \
```
obra, line 13: While I agree that the change this transform makes is desirable, I think it's outweighed by the value of having clang-format "just work" , even if it's run by, say, an IDE's tools

obra:
> I think most of the code tends to use `Right` alignment now. It would be easy to add in whichever is decided (if we do end up switching to clang-format).

I have a mild preference for 'Right', but won't die on that hill. I do think that explicitly picking one or the other is better than deriving it per file.

I think we *should* switch to clang-format. I spent some time tonight trying to get clang-format to better match our existing astyle config, but couldn't improve on what you specified.

algernon:
> * I left the line length (`ColumnLimit`) turned off, because it seems to minimize the changes it makes to our existing code.  I think we probably want to set it to some reasonable value, but maybe it's best to leave it as-is.

I'd leave it as-is. Sometimes it is desirable to have longer lines for one reason or the other, and adding guard comments around them feels worse than not limiting line length. On the other hand, no hard stance on this one, so if we do end up with a limit, that's fine too.

> * Consecutive declarations are not aligned. I think I like it better with this turned on, but not in places where one of the type names is much longer than others. I wish there was finer-grained control.

I'd prefer it off, precisely for the reason you listed: if one of the declarations has a much longer type name, this would look awful.

> > I think most of the code tends to use `Right` alignment now. It would be easy to add in whichever is decided (if we do end up switching to clang-format).
> 
> I have a mild preference for 'Right', but won't die on that hill. I do think that explicitly picking one or the other is better than deriving it per file.

I have a medium preference for `Right`, and a strong preference for picking one and sticking with it consistently, rather than per-file.

> I think we _should_ switch to clang-format.

100%  agreed.

plugins/Kaleidoscope-LED-AlphaSquare/src/kaleidoscope/plugin/LED-AlphaSquare.h
gedankenexperimenter: Indeed, I missed this one.  Thanks!

bin/format-code.sh
gedankenexperimenter: I find the `__attribute__((weak))` at the start of the line really hides the declaration from my eyes, so I have a strong preference for putting it on the line above, but I take your point.  Having just played with setting up VS Code to use clang-format on Kaleidoscope last night, I do have two alternative suggestions for your consideration, but I'm okay with whatever you decide.  Here are the options I've thought of:

1. Just let clang-format do its thing, and maybe someday it'll have an option for line breaks after leading attribute macros.
2. Add a single-file version of this script that first calls clang-format, then makes further adjustments that we like, and configure IDEs to call that script instead of clang-format directly.  This is what I would do for my own project.
3. Switch from leading attribute declarations to trailing ones.  I think this works, but the docs I found from gcc are unclear.  I would prefer this to option 1:
```c++
void strongFunc();
bool weakFunction(int test) __attribute__((weak));
// vs
void strongFunc();
__attribute__((weak)) bool weakFunction(int test);
```

gedankenexperimenter: GitHub's syntax highlighter made the second one look worse than it really is, but I still have a strong preference for trailing attributes if they have to be on the same line (assuming that actually works properly).

gedankenexperimenter:
> > > I think most of the code tends to use `Right` alignment now. It would be easy to add in whichever is decided (if we do end up switching to clang-format).
> > 
> > 
> > I have a mild preference for 'Right', but won't die on that hill. I do think that explicitly picking one or the other is better than deriving it per file.
> 
> I have a medium preference for `Right`, and a strong preference for picking one and sticking with it consistently, rather than per-file.

I used to have a preference for `Left` because the `*` or `&` is part of the type, not part of the variable name, but `const`-correctness changed my mind, since there's no way to make it really consistent in every case, and now I pretty much always use `Right`.  So I suggest we go with this strict-right config:
```yaml
DerivePointerAlignment: false
PointerAlignment: Right
ReferenceAlignment: Right
```

gedankenexperimenter:
I'm baffled.  I made the requested changes, force-pushed the branch, and the formatting check failed, apparently because it's now getting an older version of `clang-format` (11 _vs_ 12)?!  Either that, or I'm losing my mind, because I'm certain I had this line well before I submitted the PR the first time:
```yaml
AlignConsecutiveMacros: AcrossEmptyLines
```
Can we get clang-format >= version 12 on the Ubuntu image?  Or should I just format it with v13 locally, and turn off formatting on the `key_defs/*` files, where that option really helps?

bin/format-code.sh
obra: I have one more possible option:
```
void strongFunc();
__attribute__((weak))  //
bool weakFunction(int test);
```

If this doesn't float your boat, I think I lean toward trailing attributes.

obra:
> Can we get clang-format >= version 12 on the Ubuntu image? Or should I just format it with v13 locally, and turn off formatting on the `key_defs/*` files, where that option really helps?

I think that if you call it as 'clang-format-12' in the github action (presumably by setting a `CLANG_FORMAT_CMD` env variable and consuming it in your `format-code.sh`, it will do what you mean.

bin/format-code.sh
gedankenexperimenter: I know I tried that at one point, but I thought there was some problem with it.  I may be confusing it with some other stuff I tried, though.

gedankenexperimenter: Yeah, for some mysterious reason, the comment trick doesn't work in this case.  We get this instead:
```c++
__attribute__((weak))  //
bool
weakFunction(int test);
```

gedankenexperimenter: Regardless, we can kick that style issue down the road.

gedankenexperimenter:
Incidentally, we could get the formatting check (and probably some others) to run much more efficiently by separating them from `make`.  This is especially true for the github actions, because for some reason I can't discern, `make setup` needs to run before `make check-formatting`, and that takes quite a while.  I could easily add a `--check` option to `format-code.sh`, and skip `make` entirely, and it would run much quicker.  I'm certain `check-shellcheck` and `find-filename-conflicts` could also speed up in the same way, and probably `check-cpplint`.

obra:
You want to look at github’s actions documentation. they install a range newer clangs and all sorts of other stuff.

gedankenexperimenter:
> You want to look at github’s actions documentation. they install a range newer clangs and all sorts of other stuff.

Thanks!  I stopped when I found the Ubuntu version number, and didn't look further.  They do include clang-format 12, with clang-format 11 as the default version, so I can also skip the apt-get call, too.  Running it without involving `make` cuts the time to run that check down from ~1 minute to ~5 seconds.  Run locally on my laptop, it's less than one second without using `make`.

bin/format-code.sh
obra: Happy to kick this one down the road.

.github/workflows/build.yml
```diff
@@ -25,13 +26,11 @@ jobs:
...
     - run: make -j $(nproc) simulator-tests
-  check-astyle:
+  check-formatting:
     runs-on: ubuntu-latest
     steps:
-    - run: sudo apt-get install astyle
     - uses: actions/checkout@v2
-    - run: make setup
-    - run: make check-astyle
+    - run: bin/format-code.sh --check
```
obra, line 33: You don't need to run 'make setup', but I'd rather we stick to the makefile `check-formatting` target in the github action. 

The perf issue you flagged is real, but somewhat immaterial for running in CI

gedankenexperimenter: I do have an aversion to using makefiles for anything other than compiling, but that's mostly because the syntax is relatively unfamiliar (and therefore bizarre and mysterious).  If you can make it so `make check-formatting` doesn't fail without first running `make setup`, that's fine with me.  The reason I did it this was was that I _did_ need the `make setup`, or the other target wouldn't run properly.

obra: yup https://github.com/keyboardio/Kaleidoscope/tree/f/no-make-implicit-rules is poorly named but has that fix

algernon (approved):
I'm not sure I'm sold on `AlignConsecutiveAssignments`, I find it awkward more often than not, but it doesn't bother me enough to build a real case against it. Everything else looks good, and switching from astyle to clang-format is certainly a great step forward.

Great work! :shipit:

gedankenexperimenter:
> I'm not sure I'm sold on `AlignConsecutiveAssignments`, I find it awkward more often than not…

I wish there was an `AllowAlignedConsecutiveAssignments`; I agree that it's not always good, but in some cases it's a very big benefit:
```c++
int thing_one   = 1 << 0;
int thing_two   = 1 << 1;
int thing_three = 1 << 3;
int thing_four  = 1 << 4;
// or…
uint8_t WAS_PRESSED = 0b00000001;
uint8_t IS_PRESSED  = 0b00000010;
uint8_t INJECTED    = 0b0100000;
uint8_t FORGOTTEN   = 0b00010000;
```
(I made two genuine errors when writing the above examples, amusingly, but they illustrate my point very well.  When I switched to "Preview" to check alignment (hard to be sure with the variable-width font), those errors jumped out at me, but they weren't obvious at all without them being aligned.)

gedankenexperimenter:
I updated this with another bit of code that should have had formatting turned off (caught by using difftastic).  Also rebased on master.

PR #1147, opened 2022-03-28
A variety of performance and readability improvements to the setup, build and test processes

obra:
I'd love to have at least one other person try this branch before merge, but at least for me (and the simulator), it dramatically cuts time to run the test suite and makes the output more readable.

It also replaces some hairy shell scripts inside our makefiles with proper make dependencies.


gedankenexperimenter:
I'll give it a try later this afternoon.  Kids are home for spring break, and I have errands to run today, so my time is limited.

gedankenexperimenter:
I don't have the expertise to evaluate the code changes, but I have tested this out fairly thoroughly, including running `make setup` from scratch, and everything works, and is significantly faster.  One side benefit: bash completion of makefile targets is now fast enough to be useful in some cases.

gedankenexperimenter (approved):

PR #1155, opened 2022-04-06
Various minor build infrastructure improvements

gedankenexperimenter:
[Much of what had been in this PR was spun out into other PRs.  I'm keeping the original comments below for context.]

This is a collection of loosely-associated minor changes to the build system and code quality checking.

- Reorganize the main Kaleidoscope makefile meant to make it easier to understand.
- Remove obsolete `set-timestamps-from-git`.
- Replace `find-filename-conflicts` with something a bit more comprehensive.
- Change the way we call `shellcheck` to (hopefully) make it less cryptic.

These things all make more sense to _me_ as improvements, but I don't want to give the impression that I think any of it is perfect, or that I'm especially attached to any particular change here.

---

~~The main part of this PR is the `include-what-you-use` wrapper script that can now be run in an automated fashion to verify (to a certain extent) that source files follow the style guide.  While working on that, I found a few errors that slipped past in the previous round of IWYU (none of which seem to be harmful at the moment).  One important feature is the `.iwyu_ignore` file, which lets us exclude the files for which it just won't work properly.~~

~~I also expanded the use of `clang-format` to include example sketches (and updated their comments to use `clang-format` instead of `astyle` directives).~~

~~There are a bunch of other improvements, centered around the support scripts in `bin/` and the makefiles.~~

~~I also included a change to allow coloring of output to the terminal when using `make -j6 simulator-tests`, and sending output from ktest-generated testcases to the same place as output from the rest of gtest.~~


obra (commented):
So I don't think we...need the more featureful python script. 

But, more to the point, we no longer need adjust-git-timestamps at all. It was used as part of the old travis infrastructure. It can just be removed.

obra (commented):
This is good.

bin/check-cpplint.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+: "${KALEIDOSCOPE_DIR:=$(pwd)}"
+cd "${KALEIDOSCOPE_DIR}"
+
+OPTIONS+=(
```
obra, line 6: This doesn't match the old logic.  --noisy complained about more stuff than "normal" - it was an aspirational goal we aren't yet hitting.

But also, cpplint supports 'CPPLINT.cfg' -instead of a wrapper script, let's put the standard config in the config file?

obra:
Definitely break out the IWYU stuff to its own PR and the clang-format stuff to its own PR. The rest might be reviewable as one PR pretty reasonably?

bin/check-cpplint.sh
gedankenexperimenter: This is the one I had looked into the least; I hadn't even checked if it could be done with a config file.  That definitely sounds better than what I did!

gedankenexperimenter:
> So I don't think we...need the more featureful python script.
> 
> But, more to the point, we no longer need adjust-git-timestamps at all. It was used as part of the old travis infrastructure. It can just be removed.

I've been avoiding pestering you with "why is this thing here?" questions about this stuff; I'm perfectly happy to remove it!  Just for the record, it wasn't really that I was after something more featureful.  I was taking a close look at everything in the `bin/` dir, and noticed that, because the plugins had been moved, they weren't getting targeted by the old script anymore, but it wasn't easy to see that because it was hard-coded in the middle of a long line of code.  My goal with everything in this PR is clarity.

gedankenexperimenter:
> Definitely break out the IWYU stuff to its own PR and the clang-format stuff to its own PR. The rest might be reviewable as one PR pretty reasonably?

Sure!  Just to save myself from repetition (and minimize chances for errors when resolving any conflicts that come up during rebases), which of those do you want first?  I'm guessing it makes the most sense to do the applying `clang-format` to the example sketches first, then IWYU, then the rest?

bin/check-cpplint.sh
gedankenexperimenter: Ugh.  Apparently, `cpplint.py`'s config file(s) take precedence over command-line options, and the config file name is hard-coded, so if want a "noisy" mode and a "quiet" mode, you can only specify the filters that they have in common in the config file.  Or we could use a wrapper that symlinks to the appropriate config file.  Or we could edit `cpplint.py`, which leads me to ask where this version of it came from, because I can't find a version of it that looks like ours in the history here: https://github.com/google/styleguide/tree/gh-pages/cpplint)

gedankenexperimenter: Ah, I think I found it here: https://github.com/cpplint/cpplint

obra:
> I've been avoiding pestering you with "why is this thing here?" questions

Those kinds of questions are *great*, especially if my response ends up resulting in a removal or something getting better documented.

obra:
> Sure! Just to save myself from repetition (and minimize chances for errors when resolving any conflicts that come up during rebases), which of those do you want first? I'm guessing it makes the most sense to do the applying `clang-format` to the example sketches first, then IWYU, then the rest?

That ordering seems sensible.

gedankenexperimenter:
Obviously, I've still got at least one thing to fix up here.  When I get it working properly, I'll mark the PR ready for review.

gedankenexperimenter:
> Those kinds of questions are _great_, especially if my response ends up resulting in a removal or something getting better documented.

I somehow missed this message earlier.  What I really meant was that I was avoiding posing those questions while it seemed you were busy with far more important matters, rather than that I would avoid asking them in general.

obra:
Oh, I’m very capable of ignoring things when I need to. You should always feel free to ask stuff. 

jesse

gedankenexperimenter:
One thing I want to call out from this set of changes: the `.DEFAULT_GOAL` makefile variable (vs the `DEFAULT_GOAL` target hack) appears to have been introduced way back in ~2006 (I think; that's from memory of when I looked it up a couple of days ago) in version 3.81 of GNU make—which happens to be the version that Mac OS is still using (WTF, Apple?!).  It does work, however, even without installing a version that's less than ten years old.

obra:
> One thing I want to call out from this set of changes: the `.DEFAULT_GOAL` makefile variable (vs the `DEFAULT_GOAL` target hack) appears to have been introduced way back in ~2006 (I think; that's from memory of when I looked it up a couple of days ago) 

Great.

> in version 3.81 of GNU make—which happens to be the version that Mac OS is still using (WTF, Apple?!).

It's a licensing thing. Apple won't ship GPLv3 code.

>  It does work, however, even without installing a version that's less than ten years old.

Great

Makefile
```diff
@@ -97,17 +106,23 @@ build-gtest-gmock:
 	(cd testing/googletest && cmake -H. -Bbuild -DCMAKE_C_COMPILER=$(call _arduino_prop,compiler.path)$(call _arduino_prop,compiler.c.cmd) -DCMAKE_CXX_COMPILER=$(call _arduino_prop,compiler.path)$(call _arduino_prop,compiler.cpp.cmd)  .)
 	$(MAKE) -C testing/googletest
 
+.PHONY: adjust-git-timestamps
```
obra, line 100: I think  this wants to get removed?

gedankenexperimenter: Yes it does; thanks!  (A mistake born of too many rebases…)

gedankenexperimenter:
> It's a licensing thing. Apple won't ship GPLv3 code.

Ah, I see (presumably the same reason they want me to use zsh instead of bash).  But even so, I stand by my exasperated remark.

PR #1156, opened 2022-04-06
Update cpplint, with config file changes

gedankenexperimenter:
This updates cpplint to the latest version from https://github.com/cpplint/cpplint, with the addition of the ability to set the config file name on the command line (cpplint/cpplint#198).  Now, when we call cpplint from the makefile, it reads the filters from the specified file (either `.cpplint` or `.cpplint-noisy`), rather than the command line (which I botched in an earlier attempt at making this clearer).

This wouldn't have been possible without the alternate config filename changes because the config file overrides the command-line options.  It's also nice to have the config file named `.cpplint` instead of `CPPLINT.cfg`.

### DIFF
--- a/.cpplint
+++ b/.cpplint
@@ -1,0 +1,9 @@
+set noparent
+
+extensions=cpp,h,ino
+
+filter=-build/include
+filter=-legal/copyright
+filter=-readability/namespace
+filter=-runtime/references
+filter=-whitespace
--- a/.cpplint-noisy
+++ b/.cpplint-noisy
@@ -1,0 +1,9 @@
+set noparent
+
+extensions=cpp,h,ino
+
+filter=-build/include
+filter=-legal/copyright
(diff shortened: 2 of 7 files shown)


obra:
It looks like this change makes our code fail some cpplint tests.

gedankenexperimenter:
I'll get into the details later this evening, and get it fixed up.

gedankenexperimenter:
Ah.  It failed because I asked it to check the plugins, which are no longer under `src/`.  I'll fix those things up and add that to this PR when I get a chance.

obra:
It’s already demonstrating its utility!

gedankenexperimenter:
> It’s already demonstrating its utility!

Actually, this is the second time I discovered this (and I forgot about it after the first time).  I've reworked a bunch of stuff more than once in the last week or two…

PR #1158, opened 2022-04-06
Automate updating of header includes

gedankenexperimenter:
This is mainly a wrapper script that enables us to run `include-what-you-use` automatically, skipping files that it can't parse correctly (or will mangle for one reason or another — see include-what-you-use/include-what-you-use#1022).

Unfortunately, it can't easily be put into practice yet in our current git workflows, because it requires the latest version of IWYU (0.18) to get useful exit codes, and to build that, I needed to clone the repo and do some cherry-picking.  Eventually, we should be able to just plug it in and automatically enforce the policy, but for now it's improved from a very laborious manual process to a single command on a machine that has been set up for it, so it's still a big improvement.


obra:
If I could automate grabbing 0.18 from git, what kind of cherrypicking do we need to know?

gedankenexperimenter:
> If I could automate grabbing 0.18 from git, what kind of cherrypicking do we need to know?

I'll check on what I did.  They have different branches depending on which version of clang/llvm you're using.  I think I checked out v0.17 for the clang-13 branch, then cherry-picked the commits for the improved exit codes.  I doubt this will continue to be necessary for long.

gedankenexperimenter:
Here's the version that I've been using (with llvm 13.0.1 from homebrew on mac OS): https://github.com/gedankenexperimenter/include-what-you-use/tree/clang_13

obra:
Oh. 0.18 has shipped. https://github.com/include-what-you-use/include-what-you-use/releases

gedankenexperimenter:
It has, but I wasn't able to get it working for me as-is, which is why I did the cherry-picking I did.  Hopefully the homebrew version of IWYU will get caught up soon, and maybe even the Ubuntu packages for our Github workflows.

PR #1159, opened 2022-04-07
Improve output of simulator tests

gedankenexperimenter:
This restores the utility of coloring the output to the terminal when running `make -j<N> simulator-tests`, and sends the relevant output from ktest-generated tests to the same output stream as the rest of the gtest output (I didn't bother to color it, though).


gedankenexperimenter:
I forgot to mention the text coloring applies to GNU make >=4.0 with `--output-sync`. (Just making a note for future reference if anyone is searching this PR in the history.

PR #1161, opened 2022-04-08
Rename `TEST()` macro to avoid conflicts with gtest

gedankenexperimenter:
The `TEST()` macro defined in `macro_helpers.h` shared a name with a critical macro used by gtest code in the simulator, making simulator code very sensitive to the order of header includes, with rather unhelpful error messages when it failed.  This change renames the offending macro, and a related one, for good measure.  Neither renamed macro was directly used by any Kaleidoscope code, so this doesn't affect any APIs.

### DIFF
--- a/src/kaleidoscope/macro_helpers.h
+++ b/src/kaleidoscope/macro_helpers.h
@@ -198,12 +198,12 @@ int array[] = { A, B, RESTRICT_ARGS_COUNT(C, 3, B_MACRO, ##__VA_ARGS__) };
 
 #define GLUE2_AUX(...) GLUE2(__VA_ARGS__)
 
-#define TEST1(UNUSED, A, B, C, D,                                              \
-                      E, F, G, H,                                              \
-                      I, J, K, L,                                              \
-                      M, N, O, P,                                              \
-                      Q, ...) Q
-#define TEST(...) TEST1(__VA_ARGS__)
+#define HAS_MORE_THAN_2_ARGS_1(UNUSED, A, B, C, D,                             \
+                                       E, F, G, H,                             \
+                                       I, J, K, L,                             \
+                                       M, N, O, P,                             \
+                                       Q, ...) Q
+#define HAS_MORE_THAN_2_ARGS(...) HAS_MORE_THAN_2_ARGS_1(__VA_ARGS__)
 #define CHOICE(UNUSED, ...) ,##__VA_ARGS__, 1, 1, 1, 1,                        \
                                             1, 1, 1, 1,                        \
                                             1, 1, 1, 1,                        \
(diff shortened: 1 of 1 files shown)


obra:
do you have thoughts on a name that is not just appending a numeral?  given the infrequent use, something long and self documenting ought to be fine.

gedankenexperimenter:
> do you have thoughts on a name that is not just appending a numeral? given the infrequent use, something long and self documenting ought to be fine.

I'm all in favour of such a name, but since I have absolutely no idea what these macros are for, or what they do, I'll have to leave that in someone else's hands.

Anderium:
Hi both, I'm new and a total noob, but thought I'd take a look because heck it.  Maybe the name `IS_EMPTY_SIGNATURE_TEST[1]` would be appropriate?

gedankenexperimenter:
> Hi both, I'm new and a total noob, but thought I'd take a look because heck it. Maybe the name `IS_EMPTY_SIGNATURE[_TEST]` would be appropriate?

That sounds fine to me, if that's a fair description of what it does.  Is it?  I just don't have the patience to try to analyze and understand these ridiculous nested preprocessor macros.  Whatever it's called, it's going to remain mysterious and impenetrable to me; I just want the names of all these macros to be obscure enough to not cause conflict with other macros, since they can be properly namespaced.  I'd be just as happy to assign each of them a name with 20 randomly generated capital letters (or better yet, done away with entirely).

Anderium:
Alright I've looked at the whole file some more and yes, that does seem to be what it's meant for.

Since it took me some time to figure out how it works, perhaps this can be useful to someone later…:  The macros `MAKE_TEMPLATE_SIGNATURE`, `ADD_TEMPLATE_BRACES` and `TEMPLATE_KEYWORD` have two implementations, `_0` and `_1`.  The one to execute depends on whether there are 1 or 2+ arguments passed.  If I understand correctly, this is to expand `MAKE_TEMPLATE_SIGNATURE(ARG1, ARG2, ARG3, ...)` into `template<ARG2, ARG3, ...>` only if there are >= 2 arguments.  This is done by checking if `__VA_ARGS__` in `UNUSED, ...` contains any arguments, but this number should be <= 17.  (This limit exists because only the 2nd-17th arguments get gobbled.)  The `TEST` macro then uses its 18th input which is used to construct the correct macro name.  If there were no arguments, the 18th argument keeps a `0` from the `CHOICE` macro and the first implementation of the macro is used.  If there were arguments the output would be `1` and the second implementation is used.  (However, if you had an 18th argument, that would be used and you'd select `MAKE_TEMPLATE_SIGNATURE_ARG18` which doesn't exist.  And if `ARG18:=0` this would give a very difficult bug.)

gedankenexperimenter:
I have to admit to not really being able to follow that, but I'm happy to change the name to anything that doesn't conflict.  I do have two proposals:

1. Give all of these macros a `KALEIDOSCOPE_` prefix as a poor man's namespacing.
2. Add a `macro_helpers_footer.h` that undefines all of the macros in `macro_helpers.h`, and is included at the bottom of any header file that includes `macro_helpers.h`.

obra:
Perhaps change `TEST to `HAS_MORE_THAN_2_ARGS` and include @Anderium's explanation in a comment on the macro?

gedankenexperimenter:
I've changed the name as suggested, but both the macros and the explanation are essentially complete gibberish to me, so I'd prefer it if someone else wrote whatever comment goes into the file.  I've got another change that expands IWYU to work on the test simulator files, but it's blocking on this change, so it would be nice if this could get merged even if it's imperfect.

PR #1192, opened 2022-06-01
Improvements to CycleTimeReport

gedankenexperimenter:
The existing CycleTimeReport plugin uses a weighted average, where each cycle gets twice the weight of the preceding one.  As a result, only the few most recent cycles have any significant impact on the reported average cycle time.  On a keyboard like the Model01, some cycles take much longer than others (in particular, when LED updates happen), but by design those long cycles occur rarely.  This means that CycleTimeReport is unlikely to reflect the impact of those longer cycles in any given report, usually underreporting the average, but occasionally reporting a much higher value.  Also, because it only counts the microseconds between its own `beforeEachCycle()` and `afterEachCycle()` handlers in the elapsed time for a given cycle, it entirely misses any time taken up by `afterEachCycle()` handlers that are registered after it in `KALEIDOSCOPE_INIT_PLUGINS()` (and `beforeEachCycle()` handlers registered before it, though there aren't many of these).

This new version instead computes an unweighted mean cycle time for each report interval, and only does its computation once per report, minimizing its own impact on the value reported.  It's possible to be even more efficient and accurate by reporting every 256 cycles instead of using a configurable interval, but then it makes sense to simply add the code to the main `loop()` function rather than using a plugin.

I think this new plugin is clearer, and does a better job of delivering what it advertises.  It also has a significantly smaller PROGMEM footprint (but it does use a few more bytes of RAM).  This is probably because the new computation method only requires one event handler, rather than two.

### Other minor changes

- I changed the default report function to include the units ("µs") in the output for clarity.
- I deprecated the plugin's public variable, but given the nature of the plugin, maybe it would be better to simply remove it.
- I moved the overridable function from the global namespace to make it a member function, and gave it a parameter to avoid storing the computed average.

### DIFF
--- a/examples/Features/CycleTimeReport/CycleTimeReport.ino
+++ b/examples/Features/CycleTimeReport/CycleTimeReport.ino
@@ -40,11 +40,20 @@ KEYMAPS(
 )
 // clang-format on
 
+// Override CycleTimeReport's reporting function:
+void kaleidoscope::plugin::CycleTimeReport::report(uint16_t mean_cycle_time) {
+  Serial.print(F("average loop time = "));
+  Serial.println(mean_cycle_time, DEC);
+}
+
 KALEIDOSCOPE_INIT_PLUGINS(CycleTimeReport);
 
 void setup() {
   Kaleidoscope.serialPort().begin(9600);
   Kaleidoscope.setup();
+
+  // Change the report interval to 2 seconds:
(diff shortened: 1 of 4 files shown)


obra:
For a dev tool like this, I don't think I'd bother with the overhead of the deprecation. But everything else is <3

PR #1214, opened 2022-07-05
make format

gergelypolonkai:
Signed-off-by: Gergely Polonkai <gergely@polonkai.eu>


obra:
Thanks for this, but I'm going to close it out. Generally "make format" is
something that probably doesn't make a good PR, since any code updates will
end up conflicting with this.

gedankenexperimenter:
I think something may have changed in clang-format 14 _vs_ 13.  I will investigate.

PR #1215, opened 2022-07-17
[Doc] MouseKeys: fix example warp key mappings

miyagawa:
Example key mappings for the warp keys get east/west mixed up.


obra:
Thanks!

PR #1253, opened 2022-10-01
implement HostPowerManagement for Model 100

### LINKED ISSUE #1249: I want the Model 100 LED to shut off when I shut down my computer.
I shutdown my Windows 11 desktop to go to sleep but the bright LED on my Model 100 stayed on.

I'd like for the LED to turn off when my computer is off.

None.

**Additional context**
When I woke up, the LED were off, so eventually they do turn off, but I don't know when.

algernon had this to say in discord chat (#model-100):
"it's also the fault of the HostPowerManagement plugin, which has an ifdef for AVR, but no GD32 code at all
so even if the GD32 usb core would support usb suspend, the plugin doesn't make use of it
(I have no clue if the core supports it)
HostPowerManagement is the thing that is supposed to help with turning the LEDs off"

tlyu:
This requires the corresponding change in the GD32 Arduino core that adds USBCore_::isSuspended.

Fixes #1249.


tlyu:
I expect smoke-sketches to fail until the prerequisite GD32 core change is merged.

PR #1265, opened 2022-10-05
Add helper scripts to backup/restore eeprom contents

noerw:
When flashing firmware with changed eeprom layout, the stored data may become corrupted. Using these scripts before & after flashing allows to restore a working eeprom config.

Signed-off-by: Norwin <git@nroo.de>

### DIFF
--- a/bin/eeprom-backup.sh
+++ b/bin/eeprom-backup.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# creates a backup of eeprom configuration in a subdirectory of $pwd.
+# depends on focus-send. 
+
+focus="focus-send"
+subdir=kaleidoskope-eeprom-$(date -Iseconds)
+mkdir "$subdir"
+CMDS=$(focus-send help | grep -vE '^(help|eeprom|device|plugins|version)')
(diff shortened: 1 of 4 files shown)


obra:
Thanks so much for the contribution, Norwin.

To get this into the Kaleidoscope core, I'd really like the process of
using them documented or integrated into the Makefiles.

noerw:
I agree that making this a part of the build process would be great, but after taking a first look at the several makefiles I must say integrating this is way out of my scope and skill, as I have no understanding how these makefiles interact + in what contexts they are supposed to work. Also my quick&dirty scripts probably lack error checking etc needed for automated usage. After all I just wrote this to scratch an itch and thought this might help other users too.

I can add it to the docs though, I guess [this page](https://kaleidoscope.readthedocs.io/en/latest/build_default_firmware.html#install-the-firmware) wouldbe a good place.

ivanjonas:
@noerw Hi there, thanks for sharing these useful scripts. Would you mind confirming how I can use them? Is it as simple as:
1. On command line, run the backup script
2. In Arduino IDE, flash the firmware
3. On command line, run the restore script

PR #1269, opened 2022-10-07
Layer system bug fixes

gedankenexperimenter:
This fixes a couple of bugs introduce by #1039, but which had gone unnoticed because no core plugins made use of the affected functions.

- `Layer.mostRecent()` — Because shifted (momentary) layers are now represented with an offset on the layer stack, this function was returning unexpected results when a shifted layer was on top of the stack.  This bug has been fixed by always returning the "unshifted" layer number, with the offset removed.
- `Layer.deactivateMostRecent()` — I had erroneously "unshifted" the value of the top layer before calling `deactivate()` on it.  This was clearly an error, and could have resulted in the wrong layer being deactivated anywhere in the stack.

In addition to fixing these bugs, I've deprecated both `deactivateMostRecent()` and `activateNext()` on the grounds that they are not used, and don't really make sense any longer, now that the layer system is no longer index-based.

[Thanks to @sboesebeck for discovering the problem with `mostRecent()`.]

### DIFF
--- a/src/kaleidoscope/layers.cpp
+++ b/src/kaleidoscope/layers.cpp
@@ -248,7 +248,7 @@ void Layer_::activateNext() {
 }
 
 void Layer_::deactivateMostRecent() {
-  uint8_t layer = unshifted(active_layers_[active_layer_count_ - 1]);
+  uint8_t layer = active_layers_[active_layer_count_ - 1];
   deactivate(layer);
 }
 
--- a/src/kaleidoscope/layers.h
+++ b/src/kaleidoscope/layers.h
@@ -102,7 +102,8 @@ class Layer_ {
   static void move(uint8_t layer);
 
   static uint8_t mostRecent() {
-    return active_layers_[active_layer_count_ - 1];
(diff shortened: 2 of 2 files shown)


gedankenexperimenter:
This brings up a further change that might be called for.  `Layer.mostRecent()` will now report the correct value of the top shifted layer to clients, but lower shifted layers would still require the client to call `layer.unshifted(layer_number)` to decode.  Those lower-level layers are not as likely to be of interest to plugins, but perhaps we should implement an index operator and/or an iterator that gives more useful values for layers at any position in the stack.

obra:
Awesome. Thank you. Is this stuff that *could* be tested with the test suite?

gedankenexperimenter:
> Awesome. Thank you. Is this stuff that _could_ be tested with the test suite?

Yes, it can be.  It would just need a minimal custom plugin for the one bug, and a carefully-designed sequence of layer key events for the other one.  I plan to write the tests, too, but I wanted to get the fix out there and available to the user who found the bug first, and since the fix only affects code that doesn't get called anywhere in Kaleidoscope or any core plugins, it was quite safe to merge even before tests were added for it.

PR #1281, opened 2022-10-16
Fix Kal{ie,ei}doscope typos

OdyX:


obra:
(We don't need a signoff for typo fixes.)

Thanks so much for this.

OdyX:
Will consider the Signed-off-By: for next commits. Thanks for the quick merge!

PR #1282, opened 2022-10-17
Switch to clang-format-15 for CI

gedankenexperimenter:
Now that clang-format-15 is available on GitHub Actions, I think we can upgrade.  I've made the necessary changes to the clang-format config file, and run `make check-code-style` locally.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -26,11 +26,11 @@ AttributeMacros:
   - __attribute__((unused))
 BinPackArguments: false
 BinPackParameters: false
-# BraceWrapping:
-#   SplitEmptyFunction: false
-#   SplitEmptyRecord: true
-#   SplitEmptyNamespace: true
-# BreakBeforeBraces: Custom
+BraceWrapping:
+  SplitEmptyFunction: false
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBraces: Custom
 ColumnLimit: 0
 ConstructorInitializerIndentWidth: 2
 ContinuationIndentWidth: 2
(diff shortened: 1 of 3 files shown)


gedankenexperimenter:
I'm stumped.  GitHub claims to support clang-format-15 now (https://github.com/marketplace/actions/clang-format-check), but I can't figure out how to get it installed on the VM that runs the CI.

gedankenexperimenter:
Och, I'm an idiot.  It's notthing to do with GitHub's `clang-format-check`; it's the Linux VM that needs it installed…

gedankenexperimenter:
Aha! `ubuntu-latest` is not actually the latest, apparently.  If I specify `ubuntu-22.04`, it does have `clang-format-14` (which is still older than what I really want, but it might solve the formatting discrepancy that makes `make check-code-style` fail every time I run it locally because of some bug in `clang-format-12`, which I don't have a version of.

gedankenexperimenter:
> That's because `ubuntu-latest` is confusingly not 22.04, but 20.04: https://github.com/actions/runner-images#available-images, ad 20.04 only has clang-format-10, as far as I see.

The GitHub VM for 20.04 has versions of clang-format up to 12, actually.

gedankenexperimenter:
Well, I've tried it with clang-format-14, and unfortunately, I haven't found a way to configure it at doesn't munge some of our code unnecessarily, so until clang-format-15 actually becomes available, we're stuck.

obra:
Now that we're on clang-format-15 in CI, I extracted the remaining bits of this PR and merged them as #1375

PR #1283, opened 2022-10-17
improve some Focus timeout behavior

tlyu:
Add a timeout to `Focus::isEOL` to mitigate truncation that might happen during chunked writes.

Change the timeout to 10ms, which is hopefully long enough that Chrysalis won't have a problem, and short enough to mitigate some of the lockups caused by probing by third-party software.


tlyu:
> Can you do a `make format`? I'll happily merge as soon as the format check is happy.

Done, thanks!

PR #1284, opened 2022-10-18
add timeout to Focus::isEOL

tlyu:
This removes a possible source of truncated settings when chunked writes are being used.

Fixed a logic bug from the last iteration.


tlyu:
Also omits the timeout reduction, because apparently Chrysalis chunked writes delay for 50ms, which was less than the 10ms reduced timeout of the previous iteration.

PR #1291, opened 2022-10-20
Add some cross-links for EEPROM stuff

blm768:
This should make it a little more likely that newcomers find the documentation on how to deal with EEPROM layout changes. The current wording is the best my sleep-deprived brain can manage at the moment, but there's probably room for improvement. I also haven't built the docs yet since there's some messing around I'll have to do to get that working on my NixOS machine.


obra:
Thanks! Merged.

PR #1304, opened 2022-10-30
fallback keymap

tlyu:
Make `EEPROMSettings` fall back to a usable key map if it detects a possibly incompatible EEPROM layout. This forces `keymap.onlyCustom 0` and `settings.defaultLayer 0`, but does not write them to EEPROM.

A new `settings.acceptInvalid` Focus command may be used to accept the current settings as valid, updating the checksum.

### DIFF
--- a/docs/UPGRADING.md
+++ b/docs/UPGRADING.md
@@ -442,6 +442,13 @@ As a developer, one can continue using `millis()`, but migrating to `Kaleidoscop
 
 ## Breaking changes
 
+### `EEPROMSettings` key layout fallback
+
+When detecting stored settings that might be incompatible with the current firmware, `EEPROMSettings` will attempt to fall back to a usable key layout by temporarily forcing `settings.defaultLayer 0` and `keymap.onlyCustom 0`.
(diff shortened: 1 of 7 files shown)


tlyu:
This should probably not get merged until Chrysalis gets modified to do something reasonable when it encounters a keyboard in the fallback state. At least #1217 should be fixed.

Also, Chrysalis should do display a useful warning if it finds invalid settings. (It should check for mismatched CRCs in `settings.crc`; `settings.valid?` is basically useless for older firmware.) Ideally, it would give an opportunity to review the key layout, and possibly use the new `settings.acceptInvalid` command if the user approves, or else offer a factory reset.

tlyu:
OK, this unfortunately breaks the keymap toggle magic combo on the Model 100, because it unconditionally resets the keymap to `EEPROMSettings::ignoreHardcodedLayers()` on each cycle.

tlyu:
This now exactly fits into flash on the Atreus. I had to remove the redundant `Layer.move(EEPROMSettings.default_layer())` from the Atreus example sketch. I think we need to do something to free up flash on the Atreus if we want to implement a proper EEPROM directory.

PR #1315, opened 2022-11-27
Fix link to key coord header files

valrus:
...and add a separate one for the Model 100. The coords are the same for the two keyboards but the separate header file link may be useful for other reasons.

The links are to a specific commit so as to avoid having them break due to code reorganization.


obra:
Thanks so much for the PR. I think you might want to modify those links to link specifically to /master/ rather than a specific revision.

valrus:
OK. I wasn't sure which would be better; linking to /master/ seems to be how they ended up broken when the files were moved at some point in the past. Linking to a specific commit is more stable but does risk becoming out of date in a less detectable way, so maybe that's worse. I'll update the links.

PR #1317, opened 2022-12-02
Further separate layer shifts and locks

gedankenexperimenter:
When a layer lock key is pressed, ignore shifted layers in the stack when determining if the target layer is on the top of the layer stack.  This lets users put a `LockLayer(N)` on a layer that is accessed by holding `ShiftToLayer(M)`, and have more consistent behaviour.  Previously, toggling layer `N` on worked as expected, but it couldn't be toggled off with a single tap because the shifted layer `M` would be on top when the `LockLayer(N)` was tapped.

For example: A user has a base layer `0` with a `ShiftToLayer(1)` key (<kbd>fn</kbd>).  On layer `1`, there is a `LockLayer(2)` key (<kbd>X</kbd>).   On layer `2`, there is the same `ShiftToLayer(1)` key.  The idea is to use the combination <kbd>fn</kbd> + <kbd>X</kbd> to toggle layer `1` on and off, but without this change, when the user tries to turn layer `1` off, the shifted version of layer `2` is on top of the stack by the time <kbd>X</kbd> toggles on, and that causes it to simply be put on the top of the stack.  With this change, the shifted layer `2` is ignored, and layer `1` is treated as if it were on top of the layer stack, and therefore deactivated.

### DIFF
--- a/src/kaleidoscope/layers.cpp
+++ b/src/kaleidoscope/layers.cpp
@@ -118,7 +118,19 @@ void Layer_::handleLayerKeyEvent(const KeyEvent &event) {
     // LockLayer()/UnlockLayer()
     target_layer = key_code;
 
-    if (stackPosition(target_layer) == active_layer_count_ - 1) {
+    // First, we find the top non-shifted layer in the stack:
+    int8_t top_locked_layer = -1;
+    for (uint8_t i = 0; i < active_layer_count_; ++i) {
+      if (active_layers_[i] < LAYER_SHIFT_OFFSET) {
+        top_locked_layer = active_layers_[i];
+      }
+    }
+
+    // If the top locked layer is the target layer, we remove it from the stack.
+    // Otherwise, we activate it.  We disregard shifted layers so that it's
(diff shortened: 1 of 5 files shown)


obra:
Is this something we want tests for?

gedankenexperimenter:
> Is this something we want tests for?

Yes, definitely.  I should write a few, and add them to this PR.  And maybe a note in NEWS or UPGRADING?

But I'm also not completely convinced that there are no similar corner cases that this PR would make worse.  And I'm fairly convinced that it would be better (and to some degree simpler and clearer in the code) to do away with using the layer shift offset to separate shifted and locked layers and instead store two separate active layer stacks, with "shifted" layers always considered to be on top of "locked" layers.

gedankenexperimenter:
Testcase finally added for this.  Sorry it took so long.

obra:
No apology necessary!

gedankenexperimenter:
Oops.  I forgot to record here that this PR fixes #1337

PR #1319, opened 2022-12-20
Fix Model100 hardware plugin library name

gedankenexperimenter:
This was copied from the Model01 hardware plugin.  It doesn't seem to cause any problems, but it should still get corrected.


obra:
ty

PR #1321, opened 2022-12-24
Fix `KeyAddrEventQueue::shouldAbort()` (qukeys getting stuck)

gedankenexperimenter:
This fixes a problem that was causing key events to get dropped in some cases when the event id rolled over from positive to negative values while a plugin had events queued spanning both sides of the overflow.  This would generally result in event `127` getting dropped.  If the dropped event was a keyswitch toggling on, the key would be unresponsive.  If it was a key release, the key would get "stuck on".

The problem was that the 8-bit integer `KeyEventId` values were getting promoted to 16-bit integers when computing the difference between the event being flushed from the queue and the event at the head of the queue.  The resulting 16-bit value (`-128 - 127`) would be negative, whereas its 8-bit counterpart would be positive, resulting in a false positive return value from `shouldAbort()`.   The fix is to convert the difference back to an 8-bit integer before doing the comparison with zero.

I've included a testcase that fails without the fix.

### DIFF
--- a/src/kaleidoscope/KeyAddrEventQueue.h
+++ b/src/kaleidoscope/KeyAddrEventQueue.h
@@ -154,9 +154,20 @@ class KeyAddrEventQueue {
     return KeyEvent{addr(i), state, Key_Undefined, id(i)};
   }
 
-  // Only call this after `EventTracker::shouldIgnore()` returns `true`.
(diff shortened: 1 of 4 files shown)


obra:
Thank you!

PR #1323, opened 2023-01-11
reset HID protocol on reset

### LINKED ISSUE #1314: reset protocol at bus reset, not GetDescriptor(Report)
## Describe the bug

One symptom: When Windows 10 resumes from hibernation, it doesn't send a SetProtocol(Report) to the Boot Keyboard. If the UEFI/BIOS sets the Boot Keyboard to Boot Protocol, this will persist through the bus reset that occurs when Windows regains control.

The Boot Keyboard resets its protocol when its Report Descriptor is queried, assuming that it's part of enumeration. During hibernation, Windows appears to cache the Report Descriptors, and doesn't retrieve them again after resuming, even though it resets the USB bus upon regaining control from the UEFI/BIOS.

## To Reproduce

- Connect a Model 100 with custom USB trace logging code to a USB-3.0 port on a Windows host.
- Hibernate Windows, then resume.
- Observe that the keyboard sends Boot Keyboard reports, not NKRO Keyboard reports, after resume, even though it had been sending NKRO Keyboard reports before.

## Expected behavior

The Model 100 should send NKRO reports after resuming, if that's the mode that was previously active.

The Boot Keyboard should reset its protocol on bus reset, not GetDescriptor(Report), because GetDescriptor(Report) is not a reliable indicator of a bus reset having occurred.

## Environment

 - OS: Windows 10
 - Version: 22H2
 - Device: Keyboardio Model 100
 - Host: Lenovo Ideapad 110-15ACL (only tested in UEFI mode)

## Additional context

Windows appears to assume that Boot Devices will reset to Report Protocol on a USB bus reset, as required by the HID specification (version 1.11, section F.3). The HID spec also says that a host should not assume the device state, and should always explicitly set the desired protocol. Windows does not seem to follow that recommendation.

This behavior might be specific to USB-3.0 ports on Windows. There is some evidence that the Windows xHCI (USB-3.0 host controller) driver is completely separate code from the USB-2.0 host controller drivers, and makes more efforts to cache device state whenever possible. It's also notable that when waking from sleep (not hibernation), Windows does not reset the USB-3.0 port, even though it resets the USB-2.0 port.

tlyu:
Poll the keyboard device to check for a USB bus reset, and reset the protocol if so. This is necessary because polling the USB configuration state is the most portable way to detect a bus reset.

This requires keyboardio/KeyboardioHID#93, and keyboardio/ArduinoCore-GD32-Keyboardio#20 on GD32. The CI will probably fail until those are merged.

Fixes #1314.

### DIFF
--- a/src/kaleidoscope/Runtime.cpp
+++ b/src/kaleidoscope/Runtime.cpp
@@ -65,6 +65,9 @@ void Runtime_::setup(void) {
 void Runtime_::loop(void) {
   millis_at_cycle_start_ = millis();
 
+  if (device().pollUSBReset()) {
+    device().hid().onUSBReset();
+  }
   kaleidoscope::Hooks::beforeEachCycle();
 
   // Next, we scan the keyswitches. Any toggle-on or toggle-off events will
--- a/src/kaleidoscope/device/Base.h
+++ b/src/kaleidoscope/device/Base.h
@@ -305,6 +305,16 @@ class Base {
   /** @} */
 
   /**
+   * Poll the USB device for a USB reset.
+   *
+   * @return true if a USB reset has occurred
+   * @return false if a USB reset has not occurred
+   */
+  bool pollUSBReset() {
+    return mcu_.pollUSBReset();
+  }
+
+  /**
    * @defgroup kaleidoscope_hardware_keyswitch_state Kaleidoscope::Hardware/Key-switch state
    *
    * These methods offer a way to peek at the key switch states, for those cases
--- a/src/kaleidoscope/driver/hid/Base.h
+++ b/src/kaleidoscope/driver/hid/Base.h
@@ -48,6 +48,10 @@ class Base {
     keyboard().setup();
   }
 
+  void onUSBReset() {
+    keyboard().onUSBReset();
+  }
+
   auto keyboard() -> decltype(keyboard_) & {
     return keyboard_;
   }
--- a/src/kaleidoscope/driver/hid/base/Keyboard.h
+++ b/src/kaleidoscope/driver/hid/base/Keyboard.h
@@ -66,6 +66,8 @@ class NoBootKeyboard {
   uint8_t getLeds() {
     return 0;
   }
+
+  void onUSBReset() {}
 };
 
 class NoNKROKeyboard {
@@ -279,6 +281,10 @@ class Keyboard {
   }
   void setDefaultProtocol(uint8_t protocol) {
     boot_keyboard_.setDefaultProtocol(protocol);
+  }
+
+  void onUSBReset() {
+    boot_keyboard_.onUSBReset();
   }
 
  private:
--- a/src/kaleidoscope/driver/hid/keyboardio/Keyboard.h
+++ b/src/kaleidoscope/driver/hid/keyboardio/Keyboard.h
@@ -90,6 +90,10 @@ class BootKeyboardWrapper {
   uint8_t getLeds() {
     return BootKeyboard().getLeds();
   }
+
+  void onUSBReset() {
+    BootKeyboard().onUSBReset();
+  }
 };
 
 class NKROKeyboardWrapper {
--- a/src/kaleidoscope/driver/mcu/ATmega32U4.h
+++ b/src/kaleidoscope/driver/mcu/ATmega32U4.h
@@ -73,6 +73,10 @@ class ATmega32U4 : public kaleidoscope::driver::mcu::Base<_Props> {
     if (_Props::disable_clock_division)
       disableClockDivision();
   }
+
+  bool USBConfigured() {
+    return USBDevice.configured();
+  }
 };
 #else
 template<typename _Props>
--- a/src/kaleidoscope/driver/mcu/Base.h
+++ b/src/kaleidoscope/driver/mcu/Base.h
@@ -43,6 +43,31 @@ class Base {
    * Must restore the link detachFromHost severed.
    */
   void attachToHost() {}
+
+  virtual bool USBConfigured() {
+    return true;
+  }
+
+  /**
+   * Poll the USB device for a bus reset.
+   *
+   * This default implementation uses a change in USBConfigured() as a proxy
+   * for actually detecting a bus reset.
+   */
+  bool pollUSBReset() {
+    static bool was_configured;
+    if (was_configured) {
+      if (!USBConfigured()) {
+        was_configured = false;
+        return true;
+      }
+    } else {
+      if (USBConfigured()) {
+        was_configured = true;
+      }
+    }
+    return false;
+  }
 };
 
 }  // namespace mcu
--- a/src/kaleidoscope/driver/mcu/GD32.h
+++ b/src/kaleidoscope/driver/mcu/GD32.h
@@ -38,6 +38,9 @@ class GD32 : public kaleidoscope::driver::mcu::Base<_Props> {
   void attachToHost() {
     USBCore().connect();
   }
+  bool USBConfigured() {
+    return USBCore().configured();
+  }
 
 
   void setup() {
--- a/src/kaleidoscope/driver/mcu/SAMD.h
+++ b/src/kaleidoscope/driver/mcu/SAMD.h
@@ -34,6 +34,9 @@ class SAMD : public kaleidoscope::driver::mcu::Base<BaseProps> {
   void attachToHost() {
     USBDevice.attach();
   }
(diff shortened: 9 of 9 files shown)


obra:
I'm not thrilled with the naming here (and was a little quick on the draw with the merge in keyboardioHID, which wants the same rename.

is it `detectUsbBusReset()`?  I'd love to see some inline comments explaining the 'why' of this code in Kaleidoscope.

Also, I think it probably wants to happen *before* we run the before-cycle hooks.

tlyu:
Yeah, I'm not thrilled about the naming, either. It's actually both detecting the USB reset and acting on it. We could make a `onUSBReset` method, but then we'd have to figure out where to put the code that calls it, so the polling code would live elsewhere.

On GD32, we could hook into the USBCore reset handler. I think there's no equivalent on AVR, so we'd have to poll somewhere, if we don't want to modify the AVR core. Maybe the higher-level Kaleidoscope keyboard object gets a `beforeEachCycle` that does the polling and state tracking, but that puts some device-dependent lower-level code higher in the stack than we might want. Maybe not actually `beforeEachCycle`, because I don't know if we want to hook it into the plugin system.

obra:
On Wed, Jan 11, 2023 at 3:34 PM Taylor Yu ***@***.***> wrote:

> Yeah, I'm not thrilled about the naming, either. It's actually both
> detecting the USB reset and acting on it. We could make a onUSBReset
> method, but then we'd have to figure out where to put the code that calls
> it, so the polling code would live elsewhere.
>

Separating out into detection and action as different methods makes sense
to me.


> On GD32, we could hook into the USBCore reset handler. I think there's no
> equivalent on AVR, so we'd have to poll somewhere, if we don't want to
> modify the AVR core. Maybe the higher-level Kaleidoscope keyboard object
> gets a beforeEachCycle that does the polling and state tracking, but that
> puts some device-dependent lower-level code higher in the stack than we
> might want. Maybe not actually beforeEachCycle, because I don't know if
> we want to hook it into the plugin system.
>

Thinking about it, at the Kaleidoscope level, we use the results of this
decision in `boot_keyboard_.getProtocol()` in
`kaleidoscope/driver/hid/base/keyboard.h`. Does it make sense to do our
check *there*?

tlyu:
> Separating out into detection and action as different methods makes sense to me.

I agree. That also allows us to create a reset hook on GD32, while using polling on AVR as a fallback.

> Thinking about it, at the Kaleidoscope level, we use the results of this decision in `boot_keyboard_.getProtocol()` in `kaleidoscope/driver/hid/base/keyboard.h`. Does it make sense to do our check *there*?

The protocol change needs to be reflected in the GetProtocol control requests, so it has to be polled/tracked at the lower layer. Also, I think that querying the protocol in Kaleidoscope is the wrong direction in the long run. The Boot Keyboard and the NKRO Keyboard probably want to be in the same lower-level object. State changes in the Boot Keyboard affect whether we send key reports on the NKRO keyboard, so it seems to make more sense to consolidate them, and have Kaleidoscope talk to that consolidated object. Consolidation will also help us with the problem that the keyboard protocol can change while keys are held down, which currently means that they'll be "stuck" in the non-active keyboard object.

obra:
Ok. Given that, where would you want to put it?

tlyu:
I think I'll end up putting a polling function in the MCU driver. I could also have the MCU driver on GD32 register a bus reset handler hook. That seems to be how some of the platform-specific stuff is abstracted in the Kaleidoscope runtime. What are your reasons for wanting the polling to happen prior to the `beforeEachCycle` hooks, as opposed to inside `scanMatrix`?

obra:
On Jan 15, 2023, at 10:11 AM, Taylor Yu ***@***.***> wrote:﻿
I think I'll end up putting a polling function in the MCU driver. I could also have the MCU driver on GD32 register a bus reset handler hook. That seems to be how some of the platform-specific stuff is abstracted in the Kaleidoscope runtime. What are your reasons for wanting the polling to happen prior to the beforeEachCycle hooks, as opposed to inside scanMatrix?I believe that there is a chance that some plug-in could perform some key event in that hook and that we would want to have consistent information about the state of the USB device between all parts of a single cycle. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>

tlyu:
>I believe that there is a chance that some plug-in could perform some key event in that hook and that we would want to have consistent information about the state of the USB device between all parts of a single cycle.

Oh, I see. Right now, I wasn't planning to expose this more generally to plugins. I was planning for the bus reset hook to be a simple function pointer, not the kind of hook that the Kaleidoscope plugin system deals in. My intention is to make this the minimum abstraction necessary to make the necessary boot keyboard behavior change possible.

I can put the call to the polling function in front of `beforeEachCycle` to make it easier for future changes.

If you would like me to additionally expose USB reset detection to plugins using the hook system, I could work on that. Right now, I'm not aware of a need for it, and I would hesitate to add more hooks to the plugin system without a use case. (I'm not convinced that the HID drivers need to be plumbed into the plugin system, because it looks like so much of their functionality needs to work before the hook system gets initialized.)

obra:
it’s not that I want to plug-ins to be aware of USB reset detection but that I’m worried that if a plugin performs a keystroke in a before-cycle hook, it might do something inconsistent with the rest of the cycle, especially if it also does something in an after cycle hook. On Jan 15, 2023, at 11:32 AM, Taylor Yu ***@***.***> wrote:﻿

I believe that there is a chance that some plug-in could perform some key event in that hook and that we would want to have consistent information about the state of the USB device between all parts of a single cycle.

Oh, I see. Right now, I wasn't planning to expose this more generally to plugins. I was planning for the bus reset hook to be a simple function pointer, not the kind of hook that the Kaleidoscope plugin system deals in. My intention is to make this the minimum abstraction necessary to make the necessary boot keyboard behavior change possible.
I can put the call to the polling function in front of beforeEachCycle to make it easier for future changes.
If you would like me to additionally expose USB reset detection to plugins using the hook system, I could work on that. Right now, I'm not aware of a need for it, and I would hesitate to add more hooks to the plugin system without a use case. (I'm not convinced that the HID drivers need to be plumbed into the plugin system, because it looks like so much of their functionality needs to work before the hook system gets initialized.)

tlyu:
> it’s not that I want to plug-ins to be aware of USB reset detection but that I’m worried that if a plugin performs a keystroke in a before-cycle hook, it might do something inconsistent with the rest of the cycle, especially if it also does something in an after cycle hook.
>I believe that there is a chance that some plug-in could perform some key event in that hook and that we would want to have consistent information about the state of the USB device between all parts of a single cycle.

Oh. I think we already have that problem because of how the host can change the protocol of the boot keyboard in between `beforeEachCycle` and `afterEachCycle`. The reset hook adds another (possibly less common?) way that it can happen, but I think it's an existing problem, and one we should eventually solve. It's related to the problem of keys being "stuck down" in the boot keyboard if the host switches the protocol while some keys are physically pressed.

The reason I think the reset hook is a less common path is that USB resets often involve other behaviors that mean the user isn't likely to touch the keys for a few seconds. (host reboots, hotplug events, etc) Things like USB selective suspend are possibly more frequent, and might cause hidden protocol switching to occur without the user noticing.

obra:
I guess, also, it feels like a violation of the social contract to me that we're doing some of our "set up for user code" after running some user code. Is there a reason you *want* to do that check after running the 'before' hook point?

tlyu:
> I guess, also, it feels like a violation of the social contract to me that we're doing some of our "set up for user code" after running some user code. Is there a reason you _want_ to do that check after running the 'before' hook point?

I didn't have any particular reason to put the check after running the "before" hook, other than it seemed logical to fold it into the matrix scan, which seems like the "talk to the hardware" step of the main loop. It turns out that adding it to the matrix scan code is tricky, because the different sub-objects of the device driver don't have a good way to get handles on instances of each other, so we have to add some plumbing in `Runtime.cpp` anyway.

In my current implementation, the `onUSBReset` hook gets called via the ISR on GD32, so those changes will occur asynchronously with respect to the the main loop. They kind of have to, if we want to give the host a consistent view of the HID protocol state, but we don't have a way to do the same on AVR without patching the AVR core, so AVR gets to poll, and some race conditions will exist because of that.

tlyu:
Moved the USB reset polling to the runtime loop. There is an abstraction in the MCU driver to do the platform-specific parts of the polling. A separate PR will use a USB reset hook on GD32.

PR #1325, opened 2023-01-15
make BootKeyboard first HID

tlyu:
Update for KeyboardioHID using a singleton for BootKeyboard. This helps ensure that it's the first HID interface. Some UEFIs or BIOSes seem to only recognize boot keyboards that have this ordering.

This depends on keyboardio/KeyboardioHID#94.


tlyu:
I think CI will fail until the prerequisite changes in KeyboardioHID are integrated.

tlyu:
I did confirm that the LTO still works to delete the AbsoluteMouse HID interface on the Model 01, if unused. (Model 100 builds don't appear to have LTO turned on.)

tlyu:
Hm, does this also need a submodule update before the CI will succeed?

PR #1327, opened 2023-01-24
add USB reset hook on GD32

tlyu:
This is based on the revised #1323.


tlyu:
This also depends on keyboardio/ArduinoCore-GD32#21, so CI will fail until that is merged.

PR #1332, opened 2023-02-22
Fix defining multiple prefix layers

### LINKED ISSUE #1324: PrefixLayer does not work with more than one prefix layer.
With this config snippet, only the first prefix layer works:
image

image
Manually editing the `prefix_layers_length_` in the highlighted line of code to 3 fixes this issue for me so I believe the length is not being calculated correctly.

I'm using a Model 100 and I'm on the latest master branch of Kaleidoscope.

gedankenexperimenter:
The function `setPrefixLayers()` had a flaw: its parameter was a pointer to an array, without a length, resulting in the `prefix_layers_length_` variable not being set correctly.  It was meant to compute the length by dividing the size of the whole user-defined `prefix_layers` array by the size of a single element, but was instead using the size of a pointer in place of the size of the whole array.  The included testcase needs three prefix layers in order to trigger the bug because the size of a pointer in the simulator is 8 bytes.  Using a template function that takes a fixed-size array reference instead allows setting the length correctly without requiring the user to supply it as a separate argument.

I'm still a bit uncertain how prefix layers were working at all before this change.  I'm pretty sure that the size of a pointer on AVR is 2 bytes, but the size of the `PrefixLayer::Entry` data structure is 3 bytes, which would have resulted in a computed length of zero regardless of how many prefix layers were defined.

Fixes #1324

### DIFF
--- a/plugins/Kaleidoscope-PrefixLayer/src/kaleidoscope/plugin/PrefixLayer.cpp
+++ b/plugins/Kaleidoscope-PrefixLayer/src/kaleidoscope/plugin/PrefixLayer.cpp
@@ -63,11 +63,6 @@ EventHandlerResult PrefixLayer::onAddToReport(Key key) {
   return EventHandlerResult::OK;
 }
 
-void PrefixLayer::setPrefixLayers(const PrefixLayer::Entry *prefix_layers) {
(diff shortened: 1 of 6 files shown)


obra:
Thanks!

PR #1334, opened 2023-03-13
Fix "next" and "previous" layer shift keys

### LINKED ISSUE #1333: Key_KeymapNext_Momentary is no longer stacking like it used to
**Describe the bug**
`Key_KeymapNext_Momentary` no longer (new M100) seems to be functioning as it used to (on my M01s which I haven't updated in a long time).

I have set both fn keys to do `Key_KeymapNext_Momentary`. While holding both I am not ending up two layers down as I expect to but only one layer down, same as if I hold either one alone.

**To Reproduce**
1. Base firmware (dev rather than stable)
2. Firmware sketch such as http://sprunge.us/RfcIZd which sets more than one key as `Key_KeymapNext_Momentary` on the base layer and has at least two more than base
3. Flash firmware
4. Run `xev` or similar
5. Hold both fn keys and press numbers
6. See F1, F2 etc

**Expected behavior**
High F-numbers (F21, F22, etc)

 - OS: Ubuntu linux
 - Version 20.04.5 LTS
 - Device Keyboardio Model 100

**Additional context**
Similar config is working fine on my M01s, but I haven't updated them in some time.

More context in this thread: https://community.keyboard.io/t/both-palm-keys-together-to-trigger-a-separate-layer/669

gedankenexperimenter:
When shifted layers were given an offset to separate them from locked layers on the layer stack, we neglected to apply that offset correctly to the "next" and "previous" layer shift keys, as well as the more common numbered layer shift keys.

Fixes #1333.

### DIFF
--- a/src/kaleidoscope/layers.cpp
+++ b/src/kaleidoscope/layers.cpp
@@ -81,7 +81,7 @@ void Layer_::handleLayerKeyEvent(const KeyEvent &event) {
     case KEYMAP_NEXT:
     case KEYMAP_PREVIOUS:
       if (keyToggledOn(event.state)) {
-        uint8_t top_layer = active_layers_[active_layer_count_ - 1];
+        uint8_t top_layer = unshifted(active_layers_[active_layer_count_ - 1]);
         if (target_layer == KEYMAP_NEXT) {
           // Key_KeymapNext_Momentary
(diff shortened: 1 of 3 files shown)


obra:
Thank you!

PR #1336, opened 2023-03-14
Fix premature release of OneShot keys

### LINKED ISSUE #1335: Macros can trigger premature release of OneShot keys
Depending on the plugin initialization order, Macros (and DynamicMacros) can cause keys in the one-shot state to be released before they should, such that a one-shot modifier will only be applied to the first event in a macro sequence.  This happens because OneShot is ignoring events that include the `INJECTED` flag, but not ones with an invalid `KeyAddr` (which should also be ignored).

Whether or not `INJECTED` events should be ignored by OneShot is debatable, but it should definitely ignore any event without a valid `KeyAddr`.  Those events are simply not meaningful to OneShot; it has no mechanism for tracking those keys.

Original discussion where this bug was discovered: https://community.keyboard.io/t/macro-that-respects-one-shot-modifier-keys/6163

gedankenexperimenter:
If either the Macros or DynamicMacros plugin was included after OneShot in `KALEIDOSCOPE_INIT_PLUGINS`, playback of a macro containing a sequence of key presses would result in keys in the one-shot state being released during macro playback, such that only the first key in a Macros sequence would have  one-shot modifier applied to it.

This was happening because OneShot's `afterReportingState()` handler was failing to perform any checks, but simply unconditionally calling the code that could release a key under certain conditions.  This change introduces more robust guards around both key event handlers such that OneShot will now only respond to events corresponding to physical key addresses, and without the `INJECTED` flag.

Fixes #1335.

### DIFF
--- a/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
+++ b/plugins/Kaleidoscope-OneShot/src/kaleidoscope/plugin/OneShot.cpp
@@ -279,6 +279,8 @@ EventHandlerResult OneShot::onKeyEvent(KeyEvent &event) {
 
 // ----------------------------------------------------------------------------
 EventHandlerResult OneShot::afterReportingState(const KeyEvent &event) {
+  if (keyIsInjected(event.state))
+    return EventHandlerResult::OK;
   return afterEachCycle();
 }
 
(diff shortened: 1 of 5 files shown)


obra:
<3

PR #1339, opened 2023-03-28
Add Chord plugin

lugray:
This Kaleidoscope plugin allows you to define a chord of keys on your keyboard which, when pressed simultaneously, produce a single keycode. This differs from [MagicCombo](https://github.com/keyboardio/Kaleidoscope/tree/master/plugins/Kaleidoscope-MagicCombo) in that the individual keys making up a chord are suppressed, producing only the singular result.

### DIFF
--- a/plugins/Kaleidoscope-Chord/src/kaleidoscope/plugin/Chord.h
+++ b/plugins/Kaleidoscope-Chord/src/kaleidoscope/plugin/Chord.h
@@ -1,0 +1,81 @@
+/* -*- mode: c++ -*-
+ * Kaleidoscope-Chord -- Respond to chords of keys as a single keystroke
+ * Copyright (C) 2023 Lisa Ugray
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <Arduino.h>  // for PROGMEM
+#include <stdint.h>   // for uint8_t, uint16_t, int8_t
+
+#include "kaleidoscope/KeyAddr.h"               // for KeyAddr
+#include "kaleidoscope/KeyEvent.h"              // for KeyEvent
+#include "kaleidoscope/KeyEventTracker.h"       // for KeyEventTracker
+#include "kaleidoscope/event_handler_result.h"  // for EventHandlerResult
+#include "kaleidoscope/plugin.h"                // for Plugin
+#include "kaleidoscope/key_defs.h"              // for Key, Key_Transparent
+#include "kaleidoscope/progmem_helpers.h"       // for cloneFromProgmem
+
+
+namespace kaleidoscope {
+namespace plugin {
+class Chord : public kaleidoscope::Plugin {
+ public:
+  EventHandlerResult onKeyswitchEvent(KeyEvent &event);
+  EventHandlerResult afterEachCycle();
+  void setTimeout(uint8_t timeout);
+
+  template<uint8_t _chord_defs_size>
+  void configure(Key const (&chords)[_chord_defs_size]) {
+    chord_defs_      = chords;
+    chord_defs_size_ = _chord_defs_size;
+  }
+
+ private:
+  void resolveOrArpeggiate();
+  void resolve(Key target_key);
+  bool inChord(uint8_t index, Key key);
+  bool isChordStrictSubset();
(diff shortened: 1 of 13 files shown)


obra:
Hi Lisa! This is pretty neat. The two things I'd like to get added are an example sketch that shows off the code in use and a (tiny) test suite to make sure we don't break the functionality down the road.

The best place to crib from for the tests would likely be https://github.com/keyboardio/Kaleidoscope/tree/master/tests/plugins/MagicCombo, but I'd be happy to help out if you have any trouble.

-jesse

lugray:
The tests seemed quite brittle to timing changes. Have I done something wrong, or is that intended/unavoidable?

obra:
@lugray Can you tell me a bit more about what you're seeing? The tests *should* fail if the timings specified don't match the code's behavior, but it's quite possible that we're doing something wrong.

lugray:
> @lugray Can you tell me a bit more about what you're seeing? The tests _should_ fail if the timings specified don't match the code's behavior, but it's quite possible that we're doing something wrong.

That's my answer. Everything is working as intended. I was just surprised that the timing is tested so exactly - most of the places where I needed to `RUN 1 cycle` felt more like an implementation detail than desired behaviour of the code, and I'm used to testing only for what's explicitly desired.

gedankenexperimenter:
> It _should_ be possible to press multiple keys in the same cycle, although in practice, since keyscans happen every 1-2ms (iirc) humans aren't quite so good at hitting keys with that level of precision and testing with a bit of delay between presses isn't wrong either.

I'm pretty sure typical cycle times are less than 1ms even on AVR processors (presumably faster on GD32).  But when an LED update occurs on a Model01 (I have no idea about the Model100), there can be a significant delay, so it's definitely possible to have multiple switches change state in a single cycle.

obra:
On Sat, Apr 1, 2023 at 5:30 PM Michael Richters ***@***.***>
wrote:

> It *should* be possible to press multiple keys in the same cycle,
> although in practice, since keyscans happen every 1-2ms (iirc) humans
> aren't quite so good at hitting keys with that level of precision and
> testing with a bit of delay between presses isn't wrong either.
>
> I'm pretty sure typical cycle times are less than 1ms even on AVR
> processors (presumably faster on GD32). But when an LED update occurs on a
> Model01 (I have no idea about the Model100), there can be a significant
> delay, so it's definitely possible to have multiple switches change state
> in a single cycle.
>


absolutely. But it’s also possible for it to not happen. Also, the keys
changed in a single key scan is different than a cycle on the main device.
Because updates get recorded and queued by the key scanners.

plugins/Kaleidoscope-Chord/src/kaleidoscope/plugin/Chord.h
```diff
@@ -0,0 +1,81 @@
...
+}  // namespace kaleidoscope
+
+extern kaleidoscope::plugin::Chord Chord;
+
+#define CHORDS(chord_defs...)                            \
+  {                                                      \
+    static Key const chord_def[] PROGMEM = {chord_defs}; \
+    Chord.configure(chord_def);                          \
+  }
+#define CHORD(chord_keys...) chord_keys, Key_NoKey
```
gedankenexperimenter, line 81: I'd like to recommend changing the end-of-chord marker from `Key_NoKey` to `Key_Transparent`.  `Key_NoKey` is actually a potentially useful value in a chord, which can appear in the live keymap, whereas `Key_Transparent` is effectively not (I *think* it can be returned when looking up a keymap entry, but only if every active layer has a transparent entry, probably including the default layer).  If someone tries to define a `CHORD(Key_X, Key_NoKey)` or `CHORD(Key_X, Key_NoKey, Key_Y)`, not realizing what this macro does, what happens?

lugray: This PR has already merged. I'm not a maintainer in this repo, but as the original author have no objection to your opening a PR to make a change. The only thing I'd say is that if you're worried about reuse of the key, it's probably more correct to use ranges to create a `Key_ChordEnd` specifically for this.

gedankenexperimenter: It's not common that we make changes to plugins without first at least discussing those changes with the author.  At this point, I don't feel like it warrants creating an issue to discuss it, so despite the PR having been merged, I'm bringing it up here, where you'll still get notifications about it.

Your idea for dedicating a `Key` value to use as the delimiter is clearly better than my idea of using `Key_Transparent`, but perhaps we should create one that is not specific to each plugin that might have a use for it.  `Key_ListEnd`, perhaps?

lugray: I like that. If we make it that general though, we should probably consider what behaviour should happen if it makes its way somehow into an event, either by getting included in a key map, or by being injected by a plugin.

gedankenexperimenter: Unless a plugin chooses to look for that `Key` value, it would effectively get ignored by Kaleidoscope, just like any other `Key` value that isn't recognized by anything.  It would result in a spurious call to the HID layer to send a Keyboard report, but that report is almost certain to be redundant, and suppressed by the HID library.  So it *should* be harmless as a keymap entry.

lugray: Sounds good then. I'm happy to review a PR bringing that change to Chord.

PR #1341, opened 2023-04-13
Gitignore bin/arduino-cli

hupfdule:
Calling `make setup` creates the arduino-cli binary which is necessary for building Kaleidoscope. But as this file is not ignored by git it “taints” the local working copy as it looks like there are local changes which is not really the case.


obra:
Thanks!

PR #1355, opened 2023-10-02
Add plugin for applying LED colors to keys in layer

EvyBongers:


plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```c++
+#include <Kaleidoscope.h>
+#include <Kaleidoscope-LEDControl.h>
+#include <Kaleidoscope-LED-ActiveLayerKeys.h>
+
+KALEIDOSCOPE_INIT_PLUGINS(LEDControl,
+                          LEDActiveLayerKeysEffect);
+
+void setup () {
```
obra, line 22: Could you add comment here that explains that this is an example color map for a two-layer layout that sets keys on layer 0 to $color and keys on layer 1 to $other_color?

(As it's a contrived example, it might make sense to make the example show "layer 0 is red, layer 1 is green, and layer 2 is blue" or something like that.

plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Plugin properties
+
+The plugin provides the `LEDActiveLayerKeysEffect` object, which has the following
+method:
+
+### `.setColormap(colormap)`
+
+> Sets the colormap to the supplied map. Each element of the map should be a
+> `cRGB` color, and the array must have the same amount of items as there are
```
obra, line 41: What happens if the user gets the number of layers wrong?   I wonder if it would make sense to set up "default" colors for up to a dozen layers so that users don't need to override the color map unless they want to change them?

obra:
Thanks so much for submitting this draft! I left a couple notes on a few quick items that jumped out at me.

plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
EvyBongers: This behaviour is [exactly the same](https://github.com/keyboardio/Kaleidoscope/blob/master/plugins/Kaleidoscope-LED-ActiveLayerColor/README.md?plain=1#L40-L42) as with the ActiveLayerColor plugin.

It isn't too hard to change that in the [getActiveColor](https://github.com/keyboardio/Kaleidoscope/pull/1355/files#diff-bf4d14835645db8edc064486f531d7939ccfefe3c43c09a3799b842da814b96fR43-R53) method, but perhaps it would be better to have a separate PR to fix that for both plugins?

plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
EvyBongers: I'm up for it. Should I also include an example in the PR?

EvyBongers (commented):
Small note: this plugin is, for a large part, a copy of the ActiveLayerColor plugin and apart from borrowing logic from the Numpad plugin, I did little more than `s/LEDActiveLayerColor/LEDActiveLayerKeys/g` after copying the ActiveLayerColor source files. It might be worth reviewing the logic for that plugin, as both your comments apply there too :)

EvyBongers:
I'm playing with the idea to have this plugin show colours for lower layers as well (for keys that are reachable below transparent keys). The way things are how, this doesn't seem to be possible, because `layers.h` only exposes [the most recent layer](https://github.com/keyboardio/Kaleidoscope/blob/master/src/kaleidoscope/layers.h#L118-L121).

Would you be open to exposing the active layers as a read-only member to make this possible?

plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
obra: An example would be awesome. I *believe* you can just create an 'examples' directory inside the plugin and arduino will do the right thing.. I've been meaning to start moving plugin examples inside the plugins.

plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
obra: My preference would be to do it right in the new plugin and a PR to fix the old one, but I don't feel super strongly.

obra:
> Would you be open to exposing the active layers as a read-only member to make this possible?

Sure! (I'd definitely want the core change as a separate PR.)

plugins/Kaleidoscope-LED-ActiveLayerKeys/README.md
EvyBongers: I *think* it's done, but please check my logic as it comes right from StackOverflow

plugins/Kaleidoscope-LED-ActiveLayerKeys/src/kaleidoscope/plugin/LED-ActiveLayerKeys.cpp
```diff
@@ -0,0 +1,85 @@
...
+  : parent_(parent) {}
+
+void LEDActiveLayerKeysEffect::setColormap(const cRGB colormap[]) {
+  colormap_ = colormap;
+}
+
+cRGB LEDActiveLayerKeysEffect::TransientLEDMode::getLayerColor(uint8_t layer) {
+  cRGB color;
+
+  if (layer < sizeof(parent_->colormap_)/sizeof(*parent_->colormap_)) {
```
EvyBongers, line 45: This seems to break it for all but the first layer in my sketch 🤔 How should I check if the color map has an entry for a given layer?

obra:
Sorry for the radio silence. I've been in China for work with the factory for the last couple weeks. Just got home and am slowly starting to dig out from under a giant pile of email.

obra:
Thanks! Sorry for the long delay

PR #1356, opened 2023-11-02
Implement refreshAt for ActiveLayerKeys

EvyBongers:


plugins/Kaleidoscope-LED-ActiveLayerKeys/src/kaleidoscope/plugin/LED-ActiveLayerKeys.cpp
```diff
@@ -53,16 +56,22 @@ void LEDActiveLayerKeysEffect::TransientLEDMode::onActivate() {
...
-    }
+void LEDActiveLayerKeysEffect::TransientLEDMode::refreshAt(KeyAddr key_addr) {
+  uint8_t top_layer = ::Layer.mostRecent();
+  Key k             = Layer.lookupOnActiveLayer(key_addr);
+  Key layer_key     = Layer.getKey(top_layer, key_addr);
+
+  if ((k == layer_key) && (k != Key_NoKey) && (k != Key_Transparent)) {
+    ::LEDControl.setCrgbAt(KeyAddr(key_addr), active_color_);
+  } else {
+    ::LEDControl.setCrgbAt(KeyAddr(key_addr), {0, 0, 0});
```
EvyBongers, line 74: `::LEDControl.setCrgbAt(KeyAddr(key_addr), {0, 0, 0});` is needed to reset the color of keys not in the current layer.

Some time ago, I mentioned on discord that I'm thinking about implementing an option to support painting keys for all layers in the stack. This can almost be achieved by adding a condition on the `else` clause like this
```suggestion
  if ((k == layer_key) && (k != Key_NoKey) && (k != Key_Transparent)) {
    ::LEDControl.setCrgbAt(KeyAddr(key_addr), active_color_);
  } else if (!remember_past_colors) {
    ::LEDControl.setCrgbAt(KeyAddr(key_addr), {0, 0, 0});
```

The downside of this is that unshifting a layer won't cause the leds to be reset, except for keys in the now-topmost layer. If you're interested, I'm happy to add a commit to implement that, even if it isn't how I envision it.

EvyBongers:
> I feel like the not-resetting would lead to surprise and confusion for
users? It *seems* like it's worth doing.

I'm going to leave it out and look for a better way to achieve what I want.

EvyBongers:
That was easier than I expected. I had completely overlooked `Layer.lookupActiveLayer` 😄

PR #1359, opened 2023-11-14
Add Colormap-Overlay plugin

EvyBongers:


plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "kaleidoscope/plugin/LEDControl.h"     // for LEDControl
+#include <Kaleidoscope-LED-Palette-Theme.h>     // for LEDPaletteTheme
+
+namespace kaleidoscope {
+namespace plugin {
+uint16_t ColormapOverlay::map_base_;
+
+void ColormapOverlay::setup() {
+  map_base_ = ::LEDPaletteTheme.reserveThemes(1);
+}
```
EvyBongers, line 35: As stated above, I would prefer to remove this and move that single line to [`ColormapOverlay::onSetup()`](https://github.com/keyboardio/Kaleidoscope/pull/1359/files#diff-b0e992db6b7067570f61bfaaf670c1995074526312e862f9fca49be6f04053bbR53), but that doesn't work as I expect it to.

I'd appreciate some help

Issue #1305, opened 2022-11-07
high-level design changes for Boot/NKRO switching


tlyu:
## What do you want to change?

Clarify the Boot vs NKRO functionality and implementation. This is mostly me thinking out loud and trying to capture a few things before possibly suggesting some more narrowly scoped changes.

- Implement something similar to keyboardio/KeyboardioHID#50, but with the state variable maintained in the Kaleidoscope wrappers. This effectively puts the protocol switching entirely under user control.
- Consider starting up with only the Boot Keyboard sending reports, and only turn on reports on the NKRO Keyboard if we receive an explicit Set Protocol request to turn on Report Protocol. This might help with BIOSes and KVMs that assume that a Boot Keyboard is always in Boot Protocol mode. This needs some investigation about what USB-capable OSes actually do. (Maybe some of them assume Report Protocol, as it is the HID default.)
- Track key events on both the Boot Keyboard and the NKRO Keyboard, but have Kaleidscope select which ones to send reports on.
- If we choose to use the host turning on Report Protocol as a signal to stop sending Boot Keyboard reports, we should only stop after all keys currently pressed on the Boot Keyboard have been reported as released. The host might consider them to be two completely independent keyboards that happen to reside at the same device address, so if the protocol switch happens while keys are held down, the key up events would never be sent on the Boot Keyboard, if we didn't make this change.
- We might want to consider timeouts on sending HID reports, and shut down an interface if its reports are blocked from being sent. (This means the host isn't polling the endpoint, which might happen if it's a BIOS that only talks to Boot Keyboards, and the interface is the NKRO/Multi-report interface.)

### Background

The host probably considers the Boot Keyboard and the NKRO Keyboard to be two independent keyboard interfaces (sub-devices) within a single USB device. (Technically, the NKRO Keyboard is a single report within a multi-report HID device interface.)

The Boot Keyboard only produces HID Boot Protocol reports, regardless of what protocol the host has requested with the Set Protocol request. This is fine, and probably best for backwards compatibility. The NKRO Keyboard only produces NKRO reports, regardless of what protocol the host has requested with the Set Protocol request. This is probably wrong from a HID compliance standpoint. (The NKRO Keyboard shouldn't even honor requests to switch to Boot Protocol, because it currently lacks code to produce Boot Protocol reports. Also, being a multi-report HID interface, it couldn't send any of its other reports, because Boot Protocol reports can't have Report IDs.)

[...]

PR #1360, opened 2023-11-26
unconditionally send boot report

tlyu:
This is a partial proof of concept fix for #1305. It always sends Boot Keyboard reports.

It requires a coordinated change keyboardio/KeyboardioHID#99. It will check for a macro that is only present in a new enough version of KeyboardioHID.

### DIFF
--- a/plugins/Kaleidoscope-USB-Quirks/src/kaleidoscope/plugin/USB-Quirks.cpp
+++ b/plugins/Kaleidoscope-USB-Quirks/src/kaleidoscope/plugin/USB-Quirks.cpp
@@ -28,12 +28,8 @@ namespace kaleidoscope {
 namespace plugin {
 
 void USBQuirks::toggleKeyboardProtocol() {
-  uint8_t new_protocol = !Runtime.hid().keyboard().getProtocol();
-
-  Runtime.detachFromHost();
-  Runtime.hid().keyboard().setDefaultProtocol(new_protocol);
-  delay(1000);
-  Runtime.attachToHost();
+  // Do nothing, because the old behavior might cause HID-aware hosts to
+  // receive no input, now that we mark the Boot Keyboard report as padding
 }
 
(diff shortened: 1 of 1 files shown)


obra:
when everything is all done, will we have a way to disable the NKO report so that only the boot protocol report goes out? I suspect that there may be environments where that is the actual quirk we need to work around.

tlyu:
> when everything is all done, will we have a way to disable the NKO report so that only the boot protocol report goes out?

I think that's trickier than it sounds. With the current draft PRs, if a host sets the Boot Keyboard to Boot Protocol, the non-boot interfaces won't send any reports. If a non-HID-aware host expects Boot Protocol, but doesn't explicitly set it as required by the HID spec, we will send the other reports. The host probably won't poll those endpoints, so those sends will time out (250ms, currently, which is probably user-visible). I'm very curious whether FreeBSD falls into this category.

We could make an option that is effectively "don't send any non-boot reports unless the host has taken some action that implies that it will understand them". I'm wary of doing that by default, because nothing requires a HID-aware host to do anything other than configure the device and start polling the interrupt endpoints.

In practice, HID-aware hosts tend to do at least one of the following before polling HID interfaces:

* GetDescriptor(Report)
* SetReport(Output) (to set the LEDs)
* SetIdle
* SetProtocol(Report) on the boot interfaces

I think even Windows 10, which doesn't send SetProtocol(Report) or request GetDescriptor(Report) after reset when coming out of hibernation, still sends SetIdle and SetReport(Output) at that time. Almost any host that does its own key repeat management tends to send SetIdle(0) to minimize bus traffic.

tlyu:
The feedback we have on the OpenBSD behavior suggests that maybe we want to try the approach of prepending the boot report as padding to the NKRO report.

I don't know whether this will cause problems for pre-boot or KVM environments that don't request Boot Protocol, but do expect only 8 bytes.

PR #1361, opened 2023-11-30
hybrid boot keyboard

### LINKED ISSUE #1305: high-level design changes for Boot/NKRO switching

tlyu:
Make the Boot Keyboard the only keyboard, and make it have a hybrid
report format in Report Protoocol, while still ending Boot Reports in
Boot Protocol. The hybrid reports have the Boot Report as a prefix,
which might help hosts that don't request Boot Protocol but still
expect to see it.

Adjust USBQuirks to switch between always Boot Protocol and the
hybrid report.

Add visual feedback about the target mode to USBQuirks.

~Requires keyboardio/KeyboardioHID#100, and is probably not safe to apply independently of it.~
edit: no longer requires it, because KeyboardioHID has been merged into Kaleidoscope, and I updated the patches accordingly.

Fixes #1305.


tlyu:
I would appreciate it of someone for whom the simulator tests work could try running them on this PR. (Workflows will fail because there's no way to pull in the matching KeyboardioHID commit.)

I'm having trouble getting the simulator tests to work on macOS, possibly due to XCode linker bugs. I _think_ I fixed the compilation bugs with the simulator, but due to linker failures, I can't run the tests locally.

tlyu:
Fixed a few things I missed in the simulator tests.

Dirty fixes added for macOS linker stuff. Will split and make separate PRs later.

PR #1368, opened 2023-12-12
Focus sendbytes

tlyu:
Not for merging! This is for testing purposes, and might make some sketches too large to fit on AVR platforms.

This adds a `sendbytes` command to test certain ZLP sending issues.

For example, `sendbytes 59` will send a 64-byte packet (subtract 5 bytes for the CR LF Period CR LF sent at command completion).


obra:
I'm going to close out this PR, since it's not something that's ever intended to be merged as-is. But it feels like it might someday make sense as a standalone plugin.

PR #1374, opened 2023-12-19
Add tests for the Cycle plugin

justincmoy:
This commit verifies a baseline behavior for the Cycle plugin.

Honestly, I want to combine the Cycle plugin with the Qukeys plugin and this will help ensure neither plugin breaks.


obra:
Justin, 

You are my new favorite person. I absolutely adore pull requests that just add tests. Thank you!

PR #1385, opened 2024-01-24
convert report descriptors to TinyUSB macros

tlyu:
As part of a project to abstract the HID interface code away from particular device back end drivers, import a limited subset of the TinyUSB headers, and use their much nicer report descriptor macros to convert the existing KeyboardioHID report descriptors.

Other than a separate commit that slightly optimizes the Mouse descriptor, the resulting descriptors are byte-wise identical to the originals.


obra (commented):
It feels like extra work that will lead to more extra work later to disable unused tusb definitions.   I'd love to hear a bit about why that feels like the right path here.

tlyu:
There were conflicts with some KeyboardioHID definitions, so I disabled as much stuff as possible that wasn't directly needed by the report descriptor work. I didn't take detailed notes of what all the conflicts were.

I expect that I'll re-enable some of those definitions later, as I continue to improve abstractions and move away from the existing KeyboardioHID code.

tlyu:
It's not especially clean, but the most recent commit has most of `tusb_hid.h` re-enabled.

obra:
Thanks so much for this

PR #1386, opened 2024-02-07
PrefixLayer: read layers from PROGMEM

### LINKED ISSUE #1351: PrefixLayer doesn't work on my Model 01
**Describe the bug**
The [PrefixLayer plugin](https://github.com/keyboardio/Kaleidoscope/tree/master/plugins/Kaleidoscope-PrefixLayer) doesn't seem to work on my Model 01, while the same sketch builds and works as expected on my Model 100. (I'm trying to bring my Model 01s up to my current sketch so I can use them at other workstations.)

**To Reproduce**
1. Build and upload [the PrefixLayer example sketch](https://github.com/keyboardio/Kaleidoscope/tree/master/examples/Keystrokes/PrefixLayer) to a Model 01.
2. Press and release Butterfly+L while monitoring key inputs.

**Expected behavior**
The expected behavior of this sketch is to see Ctrl+B pressed and released, then the right arrow key pressed and released.

**Actual behavior**
Only the right arrow key is pressed and released. (This seems to imply that the layer transition is being handled fine, but the usual events that should fire within the PrefixLayer plugin are not.)

 - OS: macOS; Arch Linux
 - Version: 12.6.x; N/A
 - Device Keyboardio Model 01

**Additional context**
When initially upstreaming this plugin I was developing against a Model 100. I'm unsure how best to go about debugging this issue; I'm not sure what differences in the hardware or firmware between the Model 01 and Model 100 might be causing this.

iliana:
The documentation and examples for the plugin specify that the prefix layers are PROGMEM, but we didn't actually read from PROGMEM, resulting in the plugin working improperly on AVR hardware.

Tested on a Model 01.

Fixes #1351.

### DIFF
--- a/plugins/Kaleidoscope-PrefixLayer/src/kaleidoscope/plugin/PrefixLayer.cpp
+++ b/plugins/Kaleidoscope-PrefixLayer/src/kaleidoscope/plugin/PrefixLayer.cpp
@@ -44,8 +44,8 @@ EventHandlerResult PrefixLayer::onKeyEvent(KeyEvent &event) {
     return EventHandlerResult::OK;
 
   for (uint8_t i = 0; i < prefix_layers_length_; i++) {
-    if (Layer.isActive(prefix_layers_[i].layer)) {
-      current_prefix_ = prefix_layers_[i].prefix;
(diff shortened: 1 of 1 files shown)


obra:
Amazing! Thank you so much

PR #1387, opened 2024-02-09
Add focus command for retreiving a list of installed LED effects

EvyBongers:


plugins/Kaleidoscope-LED-ActiveLayerKeys/src/kaleidoscope/plugin/LED-ActiveLayerKeys.cpp
```diff
@@ -37,6 +38,10 @@ LEDActiveLayerKeysEffect::TransientLEDMode::TransientLEDMode(
   : parent_(parent),
     active_color_{0, 0, 0} {}
 
+EventHandlerResult LEDActiveLayerKeysEffect::onLedEffectQuery() {
+  return ::Focus.sendName(F("LEDActiveLayerKeysEffect"));
```
EvyBongers, line 42: I'm abusing `::Focus.sendName` as it's what `onNameQuery` uses and it gets the job done. I should probably rename it to `::Focus.sendString` or something similar.

Any preference?

EvyBongers (commented):
On a side note, I noticed that there are several plugins that don't implement `onNameQuery`. Is that something to be fixed sometime, or is there a specific use case where that's needed for certain plugins?

plugins/Kaleidoscope-LED-ActiveLayerKeys/src/kaleidoscope/plugin/LED-ActiveLayerKeys.cpp
obra: maybe sendLine()?  since what's special about it is that it sends a linefeed afterward?

obra:
> On a side note, I noticed that there are several plugins that don't implement onNameQuery. Is that something to be fixed sometime, or is there a specific use case where that's needed for certain plugins?

I wanted the answer to be "check `event-handler-hooks.md`".  So I made that true: https://github.com/keyboardio/Kaleidoscope/blob/master/docs/api-reference/event-handler-hooks.md#onnamequery

EvyBongers:
> > On a side note, I noticed that there are several plugins that don't implement onNameQuery. Is that something to be fixed sometime, or is there a specific use case where that's needed for certain plugins?
> 
> I wanted the answer to be "check `event-handler-hooks.md`". So I made that true: https://github.com/keyboardio/Kaleidoscope/blob/master/docs/api-reference/event-handler-hooks.md#onnamequery

I had that change ready to be committed 😅 Doesn't change that I can't tell in what scenarios it is or isn't important for the know to know about it. I'll leave that up to you

plugins/Kaleidoscope-LED-ActiveLayerKeys/src/kaleidoscope/plugin/LED-ActiveLayerKeys.cpp
EvyBongers: Makes sense. Do you want that in a separate PR?

EvyBongers:
The basics are implemented in all LED mode plugins.

Example output using the M100 [example sketch](https://github.com/keyboardio/Kaleidoscope/blob/master/examples/Devices/Keyboardio/Model100/Model100.ino):
```
$ ../_firmware/kaleidoscope/bin/focus-send led_effects
LEDRainbowEffect
LEDChaseEffect
SolidColor
SolidColor
SolidColor
SolidColor
SolidColor
SolidColor
SolidColor
LEDBreatheEffect
AlphaSquareEffect
StalkerEffect
ColormapEffect
```

For the plugin to be useful for displaying LED effect names in Chrysalis, some configuration options might need to be included, for the `SolidColor` plugin for example.

obra:
Nice! It does seem like it makes sense to be able to (optionally) specify an additional string for at least the SolidColor LED effect.  I wonder if it's worth implementing something in the base class to let the user pass in a descriptive string in the constructor or with another method call and then having the base class able to respond with that in preference to the hardcoded name.

Given that, it might make sense to refactor onNameQuery into the base class and to pull the name from a string baked into the plugin so we can reuse it for this

EvyBongers:
> Nice! It does seem like it makes sense to be able to (optionally) specify an additional string for at least the SolidColor LED effect. I wonder if it's worth implementing something in the base class to let the user pass in a descriptive string in the constructor or with another method call and then having the base class able to respond with that in preference to the hardcoded name.
> 
> Given that, it might make sense to refactor onNameQuery into the base class and to pull the name from a string baked into the plugin so we can reuse it for this

I like the idea of implementing it in the base class. Right now every plugin has a few nearly identical lines, just to make this work. Moving the LED effect name to a variable allows for deduplicating lines as well as adds flexibility for including parameters, like for `SolidColor`, so it's win/win.

Will look into that soon.

EvyBongers:
There's something I have yet to figure out about the constuctor arguments, but this new design seems to be working great.
```
$ ../_firmware/kaleidoscope/bin/focus-send led_effects
LEDOff
LEDRainbowEffect
LEDRainbowWaveEffect
LEDActiveLayerKeysEffect
LEDChaseEffect
SolidColorEffect: /* TODO */
SolidColorEffect: /* TODO */
SolidColorEffect: /* TODO */
SolidColorEffect: /* TODO */
SolidColorEffect: /* TODO */
SolidColorEffect: /* TODO */
SolidColorEffect: /* TODO */
LEDBreatheEffect
AlphaSquareEffect
StalkerEffect
ColormapEffect
HeatmapEffect
LEDActiveLayerColorEffect
WavepoolEffect
DigitalRainEffect
```

Two questions:
* is there any autoformatting config I could use to pass the code style checks?
* should led modes that haven't got a name specified return a default (empty?) string? With this new design, that can be easily achieved in [LEDModeInterface.h](https://github.com/keyboardio/Kaleidoscope/pull/1387/files#diff-a9a02be3d5a99fb56747558c1fb2dd3be51eef53848be0fdd06cd1dac67f1feeR34)

obra:
Since the goal is to match up these names to the LED effect list in a desktop app like Chrysalis, I think we want a straight one to one mapping between effects and names ordered the same as they are on the device

src/kaleidoscope/plugin/LEDControl/LED-Off.h
```diff
@@ -27,7 +27,8 @@ namespace plugin {
 //
 class LEDOff : public LEDMode {
  public:
-  LEDOff(void) {}
+  LEDOff(void) { ledModeName = "LEDOff"; }
```
EvyBongers, line 30: In all led modes that didn't have a constructor yet, I added constructors as `<class_name>()` rather than `<class_name>(void)`. I couldn't find anything about it in the code style docs and I believe I've seen zero-argument constructors w/o `void` elsewhere.
Should I add the `void` everywhere else? Should I remove it here?

EvyBongers (commented):
I implemented and tested the led mode name in just about all the relevant plugins now. The only exceptions are Jukebox and Miami plugins, because with my limited understanding of C++, I failed to figure out how to implement it in those. 
On a side note: when I tried getting Tricolor to work, I got errors that I failed to resolve and at this point I'm not even sure if Tricolor is even meant to be used as a plugin.

src/kaleidoscope/plugin/LEDModeInterface.h
```diff
@@ -16,13 +16,28 @@
...
+    : ledModeName{0} {}
+  explicit LEDModeInterface(char *_ledModeName)
+    : ledModeName{_ledModeName} {}
   void activate();
 
+  kaleidoscope::EventHandlerResult onLedEffectQuery() {
+    if (ledModeName == 0) {
+      // If no name was defined, don't return anything
+      return EventHandlerResult::OK;
+    }
```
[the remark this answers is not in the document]
obra: I think that sticking it in this clause is probably going to be clearer.

plugins/Kaleidoscope-FocusSerial/src/kaleidoscope/plugin/FocusSerial.cpp
```diff
@@ -78,17 +78,22 @@ EventHandlerResult FocusSerial::afterEachCycle() {
 }
 
 EventHandlerResult FocusSerial::onFocusEvent(const char *input) {
-  const char *cmd_help    = PSTR("help");
-  const char *cmd_reset   = PSTR("device.reset");
-  const char *cmd_plugins = PSTR("plugins");
+  const char *cmd_help      = PSTR("help");
+  const char *cmd_reset     = PSTR("device.reset");
+  const char *cmd_led_modes = PSTR("led_modes");
```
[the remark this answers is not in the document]
obra: I think that dropping the word 'effect' makes a ton of sense.

Looking harder at existing code, I'd also probably further amend my naming request to make it 'led.modes' to match with the 'led.brightness'.  some day, I'll shave the yank of changing 'led_mode.default' to 'led.default_mode' or something.

src/kaleidoscope/plugin/LEDControl/LED-Off.h
obra: I *think* the idea is that the "void" just improves clarity. But it does appear that we use 'without void' more often than 'with void'.   I think removing it here is ok.

EvyBongers:
I'm not sure what I need to change in order to get those failing checks to pass. I would appreciate some help :)

obra:
>  I'm not sure what I need to change in order to get those failing checks to pass. I would appreciate some help :)

Oh. I think the issue here is that you're essentially introducing a focus dependency in the core that has not previously existed. I think you probably want to have onLedEffectQuery return a pointer
and have the code inside focus do the output.

plugins/Kaleidoscope-LED-ActiveLayerKeys/src/kaleidoscope/plugin/LED-ActiveLayerKeys.cpp
EvyBongers: @obra in case you missed this: should this change be done in a separate PR? It might be breaking for custom plugins

obra: I think it probably makes sense to stick a symbol alias in place so it doesn't break on people?

EvyBongers:
> Oh. I think the issue here is that you're essentially introducing a focus dependency in the core that has not previously existed. I think you probably want to have onLedEffectQuery return a pointer and have the code inside focus do the output.

The api expects hooks to return an `EventHandlerResult`, so it can't return a pointer. I tried to solve it by passing a pointer as reference and read it after the LED mode plugin set the value:

```c++
if (inputMatchesCommand(input, cmd_led_modes)) {
  const char *ledModeName;
  kaleidoscope::Hooks::onLedEffectQuery(ledModeName);
  ::Focus.send(ledModeName, '\n');
}
```

This resulted in only a single LED mode being sent back. After hours of debugging, the simplicity of the problem struck me: `kaleidoscope::Hooks::onLedEffectQuery(ledModeName);` only returns once the hook has been called on all plugins.

I could try to make it a list of strings that gets updated, but that seems far from efficient. It also doesn't solve the other problem that is reported in CI, that the sketch becomes too large for the Model01.

plugins/Kaleidoscope-FocusSerial/src/kaleidoscope/plugin/FocusSerial.cpp
```diff
@@ -77,18 +77,27 @@ EventHandlerResult FocusSerial::afterEachCycle() {
   return EventHandlerResult::OK;
 }
 
+void ledModeHandler_(const __FlashStringHelper *name) {
+  Runtime.serialPort().println(name);
+}
```
EvyBongers, line 82: Shouldn't this return `EventHandlerResult::OK`, similar to `::Focus.sendName`?
Even if it doesn't, with the current signature of `onLedEffectQuery` we might as well rename `::Focus.sendName` to `::Focus.sendLine` and use that, rather than having a handler method that does almost exactly the same :)

plugins/Kaleidoscope-LEDEffects/src/kaleidoscope/plugin/Jukebox.cpp
```diff
@@ -20,10 +20,12 @@
 #include "kaleidoscope/device/device.h"    // for CRGB
 #include "kaleidoscope/plugin/TriColor.h"  // for TriColor
 
-kaleidoscope::plugin::TriColor JukeboxEffect(CRGB(0xc8, 0xe8, 0xee),  /* TM */
+kaleidoscope::plugin::TriColor JukeboxEffect(F("Jukebox"),
```
EvyBongers, line 23: To allow for setting a custom name for the led mode I would change this to a constructor argument, but I'm not sure how to update the header files to support that. The same holds tru for JukeboxAlternateEffect and MiamiEffect.

src/kaleidoscope/plugin/LEDModeInterface.h
```diff
@@ -16,13 +16,32 @@
...
+    : ledModeName{_ledModeName} {}
   void activate();
 
+  EventHandlerResult onLedEffectQuery(LedModeCallback callback) {
+    if (ledModeName == 0) {
+      // If no name was defined, return a default string
+      callback(F("[unnamed led mode]"));
+    } else {
+      callback(ledModeName);
+    }
```
EvyBongers, line 41: I'm wondering if there might an argument for moving this logic into the callback function. I think that this is the more obvious place to do a default-if-unset test, but I'm curious what you think.

EvyBongers:
All of them are implemented now. Only thing is that the Tricolor effects (Jukebox and Miami) don't yet support setting a custom name.
```
$ ../_firmware/kaleidoscope/bin/focus-send led.modes
Off
Rainbow
RainbowWave
Wavepool
ActiveLayerKeys
Chase
solidRed
solidOrange
solidYellow
solidGreen
solidBlue
solidIndigo
solidViolet
Breathe
AlphaSquare
Stalker
Colormap
Heatmap
ActiveLayerColor
DigitalRain
Jukebox
JukeboxAlternate
Miami
```

plugins/Kaleidoscope-FocusSerial/src/kaleidoscope/plugin/FocusSerial.cpp
obra: EventHandlerResult::OK is something that gets returned by the actual EventHandler, which, in this case, is `onLedEffectQuery` 

This is made more confusing than it should be by me misnaming the callback that prints out the LED mode `ledModeHandler_` rather than `sendLedModeCallback_` or something like that.

And I think I agree more with your comment down-conversation about moving the 'unnamed mode' guard into the callback than about reusing sendName as the callback. 

I didn't realize that with the signature `const __FlashStringHelper *name` we could *also* pass in regular in-memory strings and that everything would work.)

(I do think sendName should get refactored, but let's tackle that outside this PR)

plugins/Kaleidoscope-LEDEffects/src/kaleidoscope/plugin/Jukebox.cpp
obra: I'm ok with not tackling that as part of this PR.

src/kaleidoscope/plugin/LEDModeInterface.h
obra: That makes sense to me.

obra:
Reading through the code, it's weird to me that we're ending up with an underscore prefixed lowerCamelCase _ledModeName pointer as an argument to the constructor, when all the other constructor params are snake_case and that we're ending up with a lowerCamelCase protected `const __FlashStringHelper *ledModeName;` in our `LEDModeInterface` class.

The style guide isn't super helpful for the constructor args, but for members, it says we do want 'em snake cased and that if they're not public, they should have trailing _s. `AccessTransientLEDMode.h`is the first example in the core I can find with an actual member variable that's protected:

` protected:
  uint8_t led_mode_id_ = 255; /* 255 means uninitialized */
};
`

I think I'd have expected the constructors to look more like:

`TriColor::TriColor(const __FlashStringHelper *led_mode_name, cRGB base_color, cRGB mod_color, cRGB esc_color) {
    led_mode_name_ = led_mode_name;
` 

and that we should end up with the member looking like this:
`
protected:
const __FlashStringHelper *led_mode_name_;
`

plugins/Kaleidoscope-FocusSerial/src/kaleidoscope/plugin/FocusSerial.cpp
EvyBongers: > This is made more confusing than it should be by me misnaming the callback that prints out the LED mode `ledModeHandler_` rather than `sendLedModeCallback_` or something like that.

I agree. I'll rename the method

EvyBongers:
> Reading through the code, it's weird to me that we're ending up with an underscore prefixed lowerCamelCase _ledModeName pointer as an argument to the constructor, when all the other constructor params are snake_case and that we're ending up with a lowerCamelCase protected `const __FlashStringHelper *ledModeName;` in our `LEDModeInterface` class.

I'll update that. I mostly focussed on getting code to work and for the parts of the style guide I was already aware of, I've tried to stick to it, but I certainly haven't read the full style guide. Thank you for pointing that out.

> The style guide isn't super helpful for the constructor args, but for members, it says we do want 'em snake cased and that if they're not public, they should have trailing _s. `AccessTransientLEDMode.h`is the first example in the core I can find with an actual member variable that's protected:
> 
> `protected: uint8_t led_mode_id_ = 255; /* 255 means uninitialized */ };`
> 
> I think I'd have expected the constructors to look more like:
> 
> `TriColor::TriColor(const __FlashStringHelper *led_mode_name, cRGB base_color, cRGB mod_color, cRGB esc_color) { led_mode_name_ = led_mode_name; `
> 
> and that we should end up with the member looking like this: `protected: const __FlashStringHelper *led_mode_name_;`

👍🏻

src/kaleidoscope/plugin/LEDModeInterface.h
```diff
@@ -16,13 +16,32 @@
...
+  LEDModeInterface()
+    : led_mode_name_{0} {}
+  explicit LEDModeInterface(const char *led_mode_name)
+    : led_mode_name_{led_mode_name} {}
   void activate();
 
+  EventHandlerResult onLedEffectQuery(LedModeCallback callback) {
+    if (led_mode_name_ == 0) {
+      // If no name was defined, return a default string
+      callback("[unnamed led mode]");
```
obra, line 38: What's the case when we would want both this  string *and* the protected const char?

obra:
@EvyBongers  - just that one question from this last review. Other than that, I just want to run it on hardware, then I'm happy to merge. Thanks so much for your work on this!

src/kaleidoscope/plugin/LEDModeInterface.h
EvyBongers: I'm not sure I understand what you mean. Do you mean that I should assign this as the default value to the protected const char?

EvyBongers: Oh, right. Now that I'm looking at the diff I see what you mean. It used to be that I didn't assign a default value, which would cause this clause to be entered. I changed the default value to "[unnamed led mode]" at some point during my tests to see if I could reduce the size and it seems I forgot to remove the default value.
I'm happy with either option, though having a default empty value and this clause might make it move obvious what happens. Just let me know what you prefer.

obra: I'd probably default to an uninitialized value as the default and sending that string if the value is empty. It make it easier for other future code to reason about empty values.

PR #1388, opened 2024-02-15
Stalker: keep LED bright while key is held

### LINKED ISSUE #1342: Stalker LED effect no longer stays full brightness on key hold
**Describe the bug**
The behaviour of the stalker LED effect (my favourite!) has changed at some point between last time I updated my Model 01s and when I finally got around to setting up my Model 100 a month or two ago.

Digging in the code it looks like the change was semi-intentionally made by @gedankenexperimenter in 2073c4f8 while adapting the plugins to KeyEvent handlers; see his comment here:

https://github.com/keyboardio/Kaleidoscope/blob/master/plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp#L54L57

I think it's a shame! This is my favourite LED plugin, and I use it by default. I'm hoping the above was meant as a temporary change so those updates could be made swiftly. But I wonder if there's a way to make it behave like it used to?

**To Reproduce**
Enable the stalker effect. (I'm using the blazing trail variant.)
Press and hold a key. See it start at full brightness and decay over a few seconds.
Release the key. See it jump back up to full brightness and start decaying again.

**Expected behavior**
Key lights up full brightness on press and stays full brightness on hold.
Brightness decay starts on release.

 - OS: Ubuntu
 - Version 20.04
 - Device Keyboardio Model 100

**Additional context**

tremby:
This closes #1342.

Thanks to @algernon for guidance. But any bugs are mine not his.

Questions:

- Is it worth removing the event handler? If we are setting the intensity to max while the key is held, I'm not sure whether the event handler (which sets it to max on press and again on release) is adding anything. Maybe it adds a fraction more responsiveness?
- Is `update` the right place for this? I've read also of `beforeSyncingLeds` and `afterEachCycle`, and I don't know which is most appropriate.

### DIFF
--- a/plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
+++ b/plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
@@ -22,6 +22,7 @@
 
 #include "kaleidoscope/KeyAddr.h"                     // for MatrixAddr, KeyAddr, MatrixAddr<>::...
 #include "kaleidoscope/KeyEvent.h"                    // for KeyEvent
+#include "kaleidoscope/LiveKeys.h"                    // for live_keys
 #include "kaleidoscope/Runtime.h"                     // for Runtime, Runtime_
 #include "kaleidoscope/device/device.h"               // for cRGB, CRGB
 #include "kaleidoscope/event_handler_result.h"        // for EventHandlerResult, EventHandlerRes...
@@ -41,24 +42,6 @@ StalkerEffect::TransientLEDMode::TransientLEDMode(const StalkerEffect *parent)
     step_start_time_(0),
     map_{} {}
 
-EventHandlerResult StalkerEffect::onKeyEvent(KeyEvent &event) {
-  if (!Runtime.has_leds)
-    return EventHandlerResult::OK;
-
-  if (!event.addr.isValid())
-    return EventHandlerResult::OK;
-
-  if (::LEDControl.get_mode_index() != led_mode_id_)
-    return EventHandlerResult::OK;
-
-  // The simplest thing to do is trigger on both press and release. The color
-  // will fade while the key is held, and get restored to full brightness when
-  // it's released.
-  ::LEDControl.get_mode<TransientLEDMode>()->map_[event.addr.toInt()] = 0xff;
-
-  return EventHandlerResult::OK;
-}
-
 void StalkerEffect::TransientLEDMode::update() {
   if (!Runtime.has_leds)
     return;
@@ -71,6 +54,12 @@ void StalkerEffect::TransientLEDMode::update() {
 
   for (auto key_addr : KeyAddr::all()) {
     uint8_t step = map_[key_addr.toInt()];
+
+    // If key is active (held), set its animation position to the start
+    if (live_keys[key_addr] != Key_Inactive) {
+      step = 0xff;
+    }
+
     if (step) {
       ::LEDControl.setCrgbAt(key_addr, parent_->variant->compute(&step));
     }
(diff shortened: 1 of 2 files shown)


tremby:
Forgot to mention: this is how it used to be before https://github.com/keyboardio/Kaleidoscope/commit/2073c4f85537e07b14d44b589ef44d98a003776d#diff-8ad5917f582836d9d7f56d572e4f08b9d7c15728ac75ec22bb1fab809f3d2ce9

plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
```diff
@@ -71,6 +69,11 @@ void StalkerEffect::TransientLEDMode::update() {
 
   for (auto key_addr : KeyAddr::all()) {
     uint8_t step = map_[key_addr.toInt()];
+
```
obra, line 72: @tremby would you mind adding a  comment here explaining what's happening?

plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
```diff
@@ -51,9 +52,6 @@ EventHandlerResult StalkerEffect::onKeyEvent(KeyEvent &event) {
   if (::LEDControl.get_mode_index() != led_mode_id_)
     return EventHandlerResult::OK;
 
-  // The simplest thing to do is trigger on both press and release. The color
-  // will fade while the key is held, and get restored to full brightness when
-  // it's released.
   ::LEDControl.get_mode<TransientLEDMode>()->map_[event.addr.toInt()] = 0xff;
```
obra, line 55: I do feel like, for clarity, if this is setting the key to full brightness on release, even when we don't need it to, it's worth a comment explaining what's going on, so that future hackers don't need to go spelunking too deeply for intent.

tremby: This comment was added when the API changed such that `onKeyEvent` only fires on press and release, whereas before that change it used to fire continuously while held too. So that comment was added to describe the degraded behaviour it had since that change.

I could certainly add a comment here saying this lights it up on press and release, but isn't that just what `onKeyEvent` means anyway, so wouldn't that be obvious to someone who knows the API?

But what if instead we move the entire `onKeyEvent` handler? I don't think it adds anything. I just tried with it entirely commented out in the cpp and h file and it's working just fine.

plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
tremby: Sure.

plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
tremby: I added a commit removing the handler. Let me know what you think.

plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
tremby: I added a comment. Honestly I am not 100% certain that `live_keys` and `Key_Inactive` mean exactly what I think they do, so please do review! I am not well versed in the Kaleidoscope API.

tremby:
Oops, I'm not used to signing commits. Squashed and signed.

plugins/Kaleidoscope-LED-Stalker/src/kaleidoscope/plugin/LED-Stalker.cpp
obra: If it still works right without the handler, then I approve :)

PR #1389, opened 2024-02-15
Turbo: keep sticky toggle key lit while active

tremby:
With this changeset, if turbo is set to sticky mode, whichever key is the toggle will stay lit up while turbo mode is active.

My C++ and my understanding of the Kaleidoscope API is not strong, so please scrutinize. The changeset works well for me, but I don't know if I've done things the best way.

### DIFF
--- a/plugins/Kaleidoscope-Turbo/src/kaleidoscope/plugin/Turbo.cpp
+++ b/plugins/Kaleidoscope-Turbo/src/kaleidoscope/plugin/Turbo.cpp
@@ -83,6 +83,9 @@ EventHandlerResult Turbo::onKeyEvent(KeyEvent &event) {
   if (event.key != Key_Turbo)
     return EventHandlerResult::OK;
 
+  // Remember the address so we can pin its LED on each cycle when necessary
+  turbo_key_addr_ = event.addr;
+
 
   if (active_) {
     // If Turbo is active, and in "sticky" mode, we abort the event when a Turbo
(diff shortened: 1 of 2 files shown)


ashkitten:
thanks for the ping. i don't currently use the turbo key (due to it not being available within chrysalis, and not having the motivation to recompile the firmware).

i don't recall what it means for a turbo key to be sticky. does this mean that the key that is pressed after the turbo key will light up continuously while rapidly sending keypresses?

assuming my interpretation is correct, i have two nits:
- is it possible for multiple keys to be stuck in turbo mode simultaneously? this only seems to account for one
- i think it would be more clear if the key flickers like the turbo key itself does when active

tremby:
> i don't recall what it means for a turbo key to be sticky. does this mean that the key that is pressed after the turbo key will light up continuously while rapidly sending keypresses?
>
> assuming my interpretation is correct...

Nope. Sticky mode (which was already there by the way; I didn't add that) makes it a toggle.

With sticky, the turbo key is a toggle. So you press the turbo key once and then it stays active until pressed again. While active, pressing (and holding) other keys makes them repeat fast.

I wanted to be certain of whether turbo mode is active or inactive so this patch makes the turbo toggle key light up while active.

Keys being turbo-repeated already flashed while held, and still do.

(Without sticky, I think the idea is that you hold down the turbo key, and whenever that's held, pressing (and holding) any other key will make that key repeat fast. I never tried this mode since I didn't like the sound of it, but also I don't think my changes could have affected it. If you'd like me to test to make sure I haven't broken it I will.)

>     * is it possible for multiple keys to be stuck in turbo mode simultaneously? this only seems to account for one

As above, nothing gets stuck. I'd never actually tried doing two keys turbo at the same time but just did and they alternate. Both flash. My turbo toggle key is lit up continuously throughout.

>     * i think it would be more clear if the key flickers like the turbo key itself does when active

It could I guess? I'll wait your your response before trying to implement that.

PR #1390, opened 2024-02-17
Add warning about needing Rosetta2 on Apple Silicon

jamesnvc:
When attempting to build on an Apple Silicon mac that does not have Rosetta2 installed, attempting to build gives a somewhat inscrutable error message, so trying to head that off by giving people a heads-up here. Maybe there's a better place for it though?

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ $ cd examples/Devices/Keyboardio/Atreus
 $ make compile
 ```
 
+If you're running an Apple Silicon Mac and you get an error on this step about "bad CPU type in executable", install Rosetta2 by running `softwareupdate --install-rosetta` and try again.
+
 5.  Install your firmware
 
 ```sh


obra:
Huh. I was pretty sure we were good without Rosetta. What's the error you're getting?

jamesnvc:
> What's the error you're getting?

```
keyboardio:gd32 1.0.0   /Users/jamescash/src/Kaleidoscope/.arduino/user/hardware/keyboardio/gd32
Error during build: fork/exec /Users/jamescash/src/Kaleidoscope/.arduino/data/packages/keyboardio/tools/xpack-arm-none-eabi-gcc/9.3.1-1.3/bin/arm-none-eabi-g++: bad CPU type in executable
```

It looks like a bunch of the Arduino executables that get downloaded are x86_64; maybe something was detected improperly with `make setup`? I was setting up the toolchain on a new machine, just cloned the repo, ran `make setup`, and got that error when trying to compile the firmware. After installing Rosetta, it worked fine...

tlyu:
Yeah, it looks like the xpack toolchains specified in the keyboardio package index are only for `x86_64-apple-darwin`.

obra:
@jamesnvc I'd love to know if https://github.com/keyboardio/Kaleidoscope/pull/1391 works for you. (Where I can, I'd generally rather bake the checks and advice into the code where folks will run into it, than add notes to the docs where they might be missed)

jamesnvc:
> @jamesnvc I'd love to know if #1391 works for you. 

Unfortunately I already installed Rosetta on that machine and there doesn’t appear to be an easy way to remove it. Agreed that the linked approach is definitely better, looks good to me.

PR #1408, opened 2024-03-12
Extend Qukeys to support adding Qukeys in the keymap 

EvyBongers:


plugins/Kaleidoscope-Ranges/src/Kaleidoscope-Ranges.h
```diff
@@ -87,6 +88,8 @@ enum : uint16_t {
   OS_CANCEL,
   CS_FIRST,
   CS_LAST = CS_FIRST + MAX_CS_KEYS,
+  QK_FIRST,
+  QK_LAST = QK_FIRST + MAX_QK_KEYS,
```
EvyBongers, line 92: Asked a few questions about this on discord, so might change this depending on answers there

plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -244,3 +244,10 @@ extern kaleidoscope::plugin::Qukeys Qukeys;
       qukey_defs};                                                  \
     Qukeys.configureQukeys(qk_table);                               \
   }
+
+/// This defines a `Key` object that will be handled by the Qukey plugin.
+/// The argument `n` is the index number of the `Qukey` in the array (starting
+/// at zero).
+constexpr kaleidoscope::Key QK(uint8_t n) {
+  return kaleidoscope::Key(kaleidoscope::ranges::QK_FIRST + n);
+}
```
EvyBongers, line 253: Mostly copy/paste from CharShift. I noticed that `LayerTapKey` and `ModTapKey` are defined within `kaleidoscope::plugin` above and that macros are used to expose those in a user friendly way. Any preference on either of those?

EvyBongers (commented):
Just a first draft of what I think might be a good way to approach a new Qukeys api. Please let me know about anything that should change before I continue with this.

gedankenexperimenter:
I don't feel like I have time to help much (I haven't even looked at Kaleidoscope code in what seems a very long time), but I can recall that the idea was to have it work in much the same way a CharShift, with essentially the same (or equivalent) `KeyPair` structure in an array stored in PROGMEM or optionally in EEPROM (yeah, okay, preferably in EEPROM so it can be configured with Chrysalis, but it would be best if that wasn't *required*…).

It seems like you've got the same idea I had, so 👍.

As for how to do deprecation of the old stuff…I'm sure I could remind myself how to go about it at some point. When I did that in the past, I tried to use preprocessor directives to isolate the code that would later be removed, both to make it easier to do the removal correctly, and to make it possibly to compile a smaller binary with the deprecated code stripped out. I'm sure there are a few old PRs that demonstrate how it was done then.

I'm sure I can manage to find time to do an actual code review at an appropriate time, if that would be helpful.

plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -55,6 +57,12 @@ constexpr Key LayerTapKey(uint8_t layer, Key tap_key) {
              (layer << 8) + tap_key.getKeyCode());
 }
 
+constexpr Key Qkey(Key primary_key, Key secondary_key) {
```
EvyBongers, line 60: Gonna need to find a way to fix the name collision with `constexpr Qukey(int8_t layer, KeyAddr k, Key alternate_key)` below, unless you're okay with keeping it like this.

plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -158,6 +167,10 @@ class Qukeys : public kaleidoscope::Plugin {
   EventHandlerResult afterEachCycle();
 
  private:
+  // An array of Qkeys in PROGMEM.
+  Key const *qkeys_ PROGMEM;
```
EvyBongers, line 171: I'm not sure if this needs to be initialized somehow.

plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -234,6 +247,12 @@ bool isModifierKey(Key key);
 
 extern kaleidoscope::plugin::Qukeys Qukeys;
 
+constexpr Key Qkey(Key tap_key, Key qkey) {
+  uint8_t qkey_index = Qukeys.storeQkey(qkey);
+  // TODO(EvyBongers): store the qkey index somewhere, somehow
+  return Key(kaleidoscope::ranges::QK_FIRST + tap_key.getRaw());
+}
```
EvyBongers, line 254: My first thought was to use the same approach as with the ModTap and LayerTap keys, but if I grokked the logic of storing key code and flags in the `Key` class correctly, that would not be possible without functionality. Now I just need to find some way to store the reference.

EvyBongers: On second thought, simply using the sum of these values seems bound to go wrong. I might have to store both key values in PROGMEM (or EEPROM) and return the sum of `QK_FIRST` and the returned index.

EvyBongers (commented):
I think the idea I have for making this work is about finished now. There's two TODOs left for figuring out how to keep track of the alternate values and one error to fix.

I would appreciate suggestions on how to best approach this.

PR #1410, opened 2024-03-13
refactor KeyboardioHID more abstractly

tlyu:
Refactor the platform-independent front end bits out out of KeyboardioHID, so they can be used by other back ends that aren't based on PluggableUSB.


obra:
Cool. Thanks. Does this feel ready to merge to you?

tlyu:
I think so. It's been stable for quite some time. There is one behavior change that I did to work around what seems like a bug in clang-format-15 that's not present in my local version. I think it was trying to align assignments that it shouldn't have, maybe across different blocks, or even different functions? Hopefully the commit message makes it somewhat clear.

obra:
Thanks! Merged.

PR #1412, opened 2024-03-22
implement TinyUSB HID driver back end

tlyu:
Implement a TinyUSB HID driver back end. Some future work may need to happen around buffering HID reports when resuming from suspend.

This depends on the new getProtocol method of Adafruit_USBD_HID, which has been released in Adafruit_TinyUSB_HID, but not yet  incorporated into the Adafruit nRF52 core. (CI builds should succeed anyway, because no example sketch or hardware plugin uses this driver yet.)


obra:
Thanks! Merged.

Issue #1416, opened 2024-03-23
The Numpad plugin should only highlight actual numpad keys


EvyBongers:
**What do you want to change?**

The way the Numpad plugin currently works, it checks if the numpad layer is active and that keys on the layer don't have any modifiers (key flags) applied to them. For the default layouts in the provided firmware builds this works, but for Chrysalis users who repurpose layer it's far from convenient. In order to improve user experience in Chrysalis, I see a few options:

* Store the numpad layer setting in EEPROM (allow setting a different layer as numpad layer through Chrysalis)
  * Optionally allow specifying `-1` as value to disable the numpad plugin
* Add an option (EEPROM and/or Focus) to disable the numpad plugin (instead of or in addition to checking if the numpad layer is active)
* Update the numpad plugin to only set the led color for keys on the keypad according to the HID mappings (`Key_Keypad*`)
  * Caveat is that the function of the keys is influenced by the state of numlock, which might be unexpected for users. Possibly the plugin can be extended to automatically enable/disable numlock when the numpad layer is (de)activated.

**How will it make Kaleidoscope better?**

It's not so much Kaleidoscope that would get better, but this would set the stage for improving user experience for Chrysalis users.

**What trouble might users have in adapting to the new functionality?**

In case of options 1 and 2, I don't see any trouble. Chrysalis users who use the numpad layer and plugin as is wouldn't notice any difference, as long as the default values are kept. Kaleidoscope users might have to update their sketch to match how the plugin parameters are set.
In case of option 3, users would have to update their keymap to use only keypad keys, where the factory sketches for both M01 and M100 use the number row keys and 'normal' enter rather than their keypad counterparts. As stated above, using the keymap variants of the numbers and dot would be affected by the numlock state, though it might be possible to enable/disable numlock through the plugin.

**What trouble might developers have in adapting to the new functionality?**

I don't think there would be any differences outside of the plugin itself, except for now the existing plugin parameters are set.

obra:
I'd actually really love to entirely get rid of the Numpad plugin in favor of a pre-configured color theme for that layer. It's never been *good* and dates back to before we had the option of custom per-layer LED color themes..

EvyBongers:
Oh, that should be easy enough to do. Would that be for all example sketches or just for Keyboardio hardware?

obra:
We should figure out what the upgrade scenario for the Model 100 sketch
is.  How do we make it not hurt for users flashing a new firmware onto
their existing model 100?

[...]

PR #1418, opened 2024-04-19
Expand palette to 24 colors

EvyBongers:
Following up on https://github.com/keyboardio/Kaleidoscope/issues/1416, this PR updates the led palette to support 24 colors

### DIFF
--- a/plugins/Kaleidoscope-LED-Palette-Theme/src/kaleidoscope/plugin/DefaultPalette.cpp
+++ b/plugins/Kaleidoscope-LED-Palette-Theme/src/kaleidoscope/plugin/DefaultPalette.cpp
@@ -31,7 +31,7 @@ __attribute__((weak)) extern bool palette_defined = false;
 void DefaultPalette::setup() {
   if (!ledpalette::palette_defined) return;
 
-  for (uint8_t i = 0; i < 16; i++) {
+  for (uint8_t i = 0; i < 24; i++) {
     cRGB color;
 
     color.r = pgm_read_byte(&(ledpalette::palette[i].r));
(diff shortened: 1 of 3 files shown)


EvyBongers:
So far it's just the bare minimum to support 24 colors. We should probably use a variable to define the palette size and and update the macro to accept a list of arguments rather than hard coding a fixed number of arguments.

obra:
Indeed, that little bit of refactoring is probably worth doing.  And hm. Is
the macro change going to break existing sketches?

EvyBongers:
The way it is structured now, probably. I'm thinking that there should be a way to refactor it that supports any number of colors (like how qukeys are handled). That should supply a default to black (off) in case the sketch supplies less colors that the palette supports.

obra:
Sorry for my radio silence. I think the only thing left on this PR is the macro change not breaking existing sketches - https://github.com/keyboardio/Kaleidoscope/actions/runs/8772302322/job/24071073901?pr=1418#step:5:996

obra:
Looks like tests are still a little sad:

https://github.com/keyboardio/Kaleidoscope/actions/runs/8929088241/job/24526089322#step:5:1009

EvyBongers:
I noticed. The weird thing is that the M100 sketch I have locally (that uses the same macro) does build just fine. I just haven't included that because it includes most (if not all) the other changes that we discussed in the redesign issue. I have yet to figure out what the relevant difference is.

obra:
It *may* be an issue with the older GCC used for AVR Arduino.

EvyBongers:
> It _may_ be an issue with the older GCC used for AVR Arduino.

That will be the issue.

> The extra comma in printf("hello",) is a syntax error in C and C++. C++20 solved this problem by adding a new special identifier, __VA_OPT__. The sequence __VA_OPT__(x), which is only legal in the substitution list of a variable-argument macro, expands to x if __VA_ARGS__ is non-empty and to nothing if it is empty.

(https://www.scs.stanford.edu/~dm/blog/va-opt.html#variable-argument-macros)

EvyBongers:
I found a way to get the build passing. I have yet to verify if the code functions as expected, which is why I converted it back to draft

PR #1422, opened 2024-05-07
Have the makefile print the path to the resulting bin/hex file

EvyBongers:
To make the resulting bin/hex file easier to find for users who prefer to flash their build using Chrysalis


obra:
Thanks!

PR #1427, opened 2024-05-26
Have ColormapOverlay behave more like Colormap

### LINKED ISSUE #1424: ColormapOverlay should work more like colormap
This is a follow-on from #1420 - I think that ColormapOverlay would fit better in Kaleidoscope if it were configured the same way as Colormap.

@EvyBongers's argument against:

> It can certainly be done, but I feel that that defeats the purpose of the plugin. This is something that Colormap does already and does really well. I feel that Colormap-Overlay would typically apply colors to just a few keys, at they are applied on top of the active LED effect.

What I'm thinking about is mostly about making the API easier for users to understand. Having the two plugins configured with equivalent datastructures means there's less syntax for folks to learn, we don't need to provide the big comment showing off the key map layout, etc. If the colormaps and overlay colormaps worked the same, it'd also make it more straightforward to add support for colormap overlays to chrysalis.

I don't really object to having a "just overlay this one key" function, but for use cases like how we're handling numpad, I think it'd be nice to have the whole-keyboard level config.

EvyBongers:
Adding support for specifying color overlays in a similar was as Colormap

Yet to do:
- [x] Find a way to convert key index to key address
- [x] Implement a way for ColormapOverlay to not override the led color
- [x] Implement validation of the colormap index specified (is that needed?)

Fixes #1424

### DIFF
--- a/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
+++ b/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
@@ -87,7 +87,6 @@ class ColormapOverlay : public kaleidoscope::Plugin {
 #define COLORMAP_OVERLAYS_MAP(layers...)                                \
   namespace kaleidoscope {                                              \
   namespace plugin {                                                    \
-  namespace defaultcolormap {                                           \
   const uint8_t overlays_[][kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns] = { \
       layers                                                            \
   };                                                                    \
--- a/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
+++ b/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
@@ -24,6 +24,7 @@
 #include "kaleidoscope/key_defs.h"              // for Key, KEY_FLAGS, Key_NoKey, LockLayer
 #include "kaleidoscope/layers.h"                // for Layer, Layer_
 #include "kaleidoscope/plugin/LEDControl.h"     // for LEDControl
+#include <Kaleidoscope-LED-Palette-Theme.h>     // for LEDPaletteTheme
 
 namespace kaleidoscope {
 namespace plugin {
@@ -60,7 +61,8 @@ class ColormapOverlay : public kaleidoscope::Plugin {
     for (int layer_ = 0; layer_ < _layer_count; layer_++) {
       for (int key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
         int8_t color_index_ = overlays[layer_][key_index_];
-        if (color_index_ != no_color_overlay) {
+        if (color_index_ >= 0 && color_index_ < ::LEDPaletteTheme.getPaletteSize() &&
+            color_index_ != no_color_overlay) {
           overlays_[overlay_count_] = Overlay(layer_, KeyAddr(key_index_), color_index_);
           overlay_count_++;
         }
--- a/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
+++ b/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
@@ -53,6 +53,17 @@ class ColormapOverlay : public kaleidoscope::Plugin {
     overlay_count_ = _overlay_count;
   }
 
+  template<uint8_t _layer_count>
+  void configureOverlays(uint8_t const (&overlays)[_layer_count][kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns]) {
+    overlays_      = Overlay[];
+    overlay_count_ = 0;
+    for (int layer_ = 0; layer_ < _layer_count; layer_++) {
+      for (int key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
+        overlays_[overlay_count_] = Overlay(layer_, /* TODO(EvyBongers) */, overlays[layer_][key_index_]);
+        overlay_count_++;
+      }
+    }
+  }
   // A wildcard value for a qukey that exists on every layer.
   static constexpr int8_t layer_wildcard{-1};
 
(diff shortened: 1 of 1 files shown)


plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -53,6 +53,19 @@ class ColormapOverlay : public kaleidoscope::Plugin {
...
 
+  template<uint8_t _layer_count>
+  void configureOverlays(uint8_t const (&overlays)[_layer_count][kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns]) {
+    overlays_      = Overlay[];
+    overlay_count_ = 0;
+    for (int layer_ = 0; layer_ < _layer_count; layer_++) {
+      for (int key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
+        color_index_ = overlays[layer_][key_index_];
+        // TODO(EvyBongers): validate the color index?
+        overlays_[overlay_count_] = Overlay(layer_, /* TODO(EvyBongers) */, color_index_);
```
EvyBongers, line 64: @obra I haven't been able to find a convenience function to translate a key index to a key address. I'm also unsure if I can construct a `KeyAddr` from the key index. I probably overlooked a constructor for it?

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -53,6 +53,19 @@ class ColormapOverlay : public kaleidoscope::Plugin {
...
   }
 
+  template<uint8_t _layer_count>
+  void configureOverlays(uint8_t const (&overlays)[_layer_count][kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns]) {
+    overlays_      = Overlay[];
+    overlay_count_ = 0;
+    for (int layer_ = 0; layer_ < _layer_count; layer_++) {
+      for (int key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
+        color_index_ = overlays[layer_][key_index_];
+        // TODO(EvyBongers): validate the color index?
```
EvyBongers, line 63: For implementing a way to specify a key to not have the led color overriden, I'm thinking of using `-1` as a value. I'm wondering if there's a need to validate the upper boundary of the palette. The plugin as it currently exist doesn't do that and I believe that Colormap doesn't to this either, so I feel like I should only verify that lower boundary to skip overriding the color.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -69,6 +82,36 @@ class ColormapOverlay : public kaleidoscope::Plugin {
...
+    MAP_LIST(__IDENTITY__, PER_KEY_DATA_STACKED(0, __VA_ARGS__))        \
+  }
+#endif
+
+#ifdef PER_KEY_DATA
+#define COLORMAP_OVERLAY(...)                            \
+  {                                                      \
+    MAP_LIST(__IDENTITY__, PER_KEY_DATA(0, __VA_ARGS__)) \
+  }
+#endif
```
EvyBongers, line 111: This is mostly copy/paste from DefaultColormap. I'm wondering if the namespaces need to be specified like they are 🤔

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
obra: KeyAddr is currently always implemented as a MatrixAddr.

The key index is described as `offset_` inside MatrixAddr.

Here's what we've got for constructors in there:

```

  constexpr MatrixAddr(uint8_t row, uint8_t col)
    : offset_(row * cols + col) {}

  explicit constexpr MatrixAddr(uint8_t offset)
    : offset_(offset) {}

```

The actual calculation of an offset looks like this: (from `src//kaleidoscope/device/key_indexes.h`)

```
constexpr uint8_t keyIndex(uint8_t row, uint8_t col) {
  return KeyAddr(row, col).toInt() + 1;
}
```

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
obra: -1 does not feel unreasonable to me, as long as we define it as a const rather than using a magic number in the code ;)

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
obra: Hm. I guess test it?

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -53,8 +53,23 @@ class ColormapOverlay : public kaleidoscope::Plugin {
...
 
-  // A wildcard value for a qukey that exists on every layer.
+  template<uint8_t _layer_count>
+  void configureOverlays(uint8_t const (&overlays)[_layer_count][kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns]) {
+    overlays_      = Overlay[];
+    overlay_count_ = 0;
+    for (int layer_ = 0; layer_ < _layer_count; layer_++) {
+      for (int key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
+        int8_t color_index_ = overlays[layer_][key_index_];
+        if (color_index_ != no_color_overlay) {
```
EvyBongers, line 63: Should I validate the lower and upper boundaries for `color_index_`?

obra: I think it's generally good practice to do so, yeah.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
EvyBongers: I haven't felt motivated to pursue this and messing with it could open a can of worms, so I just marked the PR ready to be merged. If I end up pursuing this, I'll do so in a separate PR.

PR #1432, opened 2024-06-15
Have ColormapOverlay support blank keys

### LINKED ISSUE #1425: ColormapOverlay overlays don't seem to apply to transparent keys
I was a little surprised, when testing out the replacement for the Numpad plugin, that setting an overlay on a key that didn't have anything defined on the numpad layer...just didn't light up the key. I'd expect that if I've explicitly defined an overlay, it wouldn't skip lighting up just because there was a transparent key on the keymap.

@EvyBongers - is this an intentional design? I'd love to understand the design intent better.

EvyBongers:
Apply color overlays regardless of the value of the key in the current (top most) layer.
This change aims to maintain backwards compatibility to apply color overlays on lower layers, provided that the top most layer does not specify a color for a key.

Due to limitations of the API, there is no guarantee that the color applied matches the highest layer matching the key except for the top most layer.

Fixes #1425


obra:
Thanks! Merged

PR #1434, opened 2024-07-12
Add focus support to ColormapOverlay

EvyBongers:
Quick attempt to implement focus support in ColormapOverlay. I feel this should work, but I don't have my keyboard with me right now, so any debugging I'll have to do later.

### DIFF
--- a/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
+++ b/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
@@ -29,9 +29,10 @@
 namespace kaleidoscope {
 namespace plugin {
 uint16_t ColormapOverlay::map_base_;
+const uint8_t no_themes_ = 1;
 
 void ColormapOverlay::setup() {
-  map_base_ = ::LEDPaletteTheme.reserveThemes(1);
+  map_base_ = ::LEDPaletteTheme.reserveThemes(no_themes_);
 }
 
 bool ColormapOverlay::hasOverlay(KeyAddr k) {
@@ -71,6 +72,10 @@ EventHandlerResult ColormapOverlay::beforeSyncingLeds() {
   setLEDOverlayColors();
 
   return EventHandlerResult::OK;
+}
+
+EventHandlerResult ColormapOverlay::onFocusEvent(const char *input) {
+  return ::LEDPaletteTheme.themeFocusEvent(input, PSTR("colormap.overlay"), map_base_, no_themes_);
 }
 
 }  // namespace plugin
--- a/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
+++ b/plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
@@ -58,6 +58,7 @@ class ColormapOverlay : public kaleidoscope::Plugin {
 
   EventHandlerResult onSetup();
   EventHandlerResult beforeSyncingLeds();
+  EventHandlerResult onFocusEvent(const char *input);
 
  private:
   static uint16_t map_base_;


plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -73,6 +74,10 @@ EventHandlerResult ColormapOverlay::beforeSyncingLeds() {
   return EventHandlerResult::OK;
 }
 
+EventHandlerResult ColormapOverlay::onFocusEvent(const char *input) {
+  return ::LEDPaletteTheme.themeFocusEvent(input, PSTR("colormap.overlay"), map_base_, no_themes_);
```
EvyBongers, line 78: `colormap.overlay` seems like the most logical focus command to me. Let me know what you think.

obra:
@EvyBongers did you end up testing this on hardware?

EvyBongers:
I completely forgot about this. I'll see about testing it today

EvyBongers:
So, I finally got around to testing this. The command returns the palette as expected, but setting the palette doesn't yet work 🤔

EvyBongers:
So, I dug a little deeper to see what I was missing. I thought I had a quick fix by calling on `::LEDPaletteTheme.themeFocusEvent()`, but the internal logic of the plugin never actually used LEDPaletteTheme to store its data because it was never intended to work as a theme.
Looking at #1416, it seems to me that the third PR that we discussed there was never created, so unless we want so implement a read only version here, this effort is going to have to wait until EEPROM support is added to Colormap-Overlay.

The confusing part is that Colormap-Overlay (seemed to) need to call `::LEDPaletteTheme.reserveThemes()` in order to work with the palette, even though it never uses any of the theme logic. It's been ages since I implemented that part and I don't remember why I needed to add that call to reserve themes nor whether it's actually needed for the plugin to work with the palette at all.

Closing this for now

EvyBongers:
Just found a comment about this [on Discord](https://discord.com/channels/492408953041321984/504378238647402507/1267867168091930694). Cross-posting to have all of this together.
> #1434 will need quite a bit of work yet, I think. My first attempt to make focus work for colormap overlay settings simply passed through the handling of the focus event to ::LEDPaletteTheme.themeFocusEvent(). I had a proper look at that today and I'm pretty sure that isn't going to work, or at least it isn't sufficient. That's partly because (iirc) I needed to reserve at least one theme to get ::LEDPaletteTheme.lookupPaletteColor() to work. However, the plugin doesn't store any other data in a theme. That means that I will either have to find a way to have the plugin work with themes or I'm gonna have to implement a second or different means for handing the focus command

obra:
Just pull it out to a separate commit and I'm good :)

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -30,13 +30,20 @@
...
 namespace plugin {
 uint16_t ColormapOverlay::map_base_;
 
 void ColormapOverlay::setup() {
+  // It appears that a call to ::LEDPaletteTheme.reserveThemes() is needed
+  // because it's where palette_base_ gets initialized. Since this plugin (and
+  // possibly others) don't actually use themes, requesting memory for storing
+  // themes doesn't make much sense. Maybe initialisation of palette_base_
+  // could be moved to a setup() method, though maybe the palette should be
+  // split from palette theme altogether?
```
EvyBongers, line 46: @obra I don't really like having to reserve memory for a theme that I don't even use. I'm happy to remove the comment and just accept it if that's what you prefer, but as stated in the comment I feel it might be better to split initialisation of the palette and logic for themes.

obra: I don't know that it should be in a 'setup' method? But possibly in a 'reservePalette' method (that the existing 'reserveThemes' would be refactored to call)?

EvyBongers: That's fine by me. Want a separate PR for that too?

obra:
Sure!

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
EvyBongers: #1482

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -30,14 +30,39 @@
...
+#include <Kaleidoscope.h>                       // for Kaleidoscope
+#include <Kaleidoscope-FocusSerial.h>           // for Focus
 #include <Kaleidoscope-LED-Palette-Theme.h>     // for LEDPaletteTheme
+#include <Kaleidoscope-EEPROM-Settings.h>       // for EEPROMSettings
 
 namespace kaleidoscope {
 namespace plugin {
 uint16_t ColormapOverlay::map_base_;
+uint16_t ColormapOverlay::overlays_base_;
+uint8_t ColormapOverlay::max_overlays_ = 64;  // TODO(EvyBongers): figure this out. How determine a good maximum?
```
EvyBongers, line 42: @obra do you have a way to determine a good value here, or is this pure guess work?

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -30,14 +30,39 @@
...
+
+  overlays_base_ = ::EEPROMSettings.requestSlice(max_overlays_ * sizeof(Overlay));
+
+  ::EEPROMSettings.seal();
+
+  if (!::EEPROMSettings.isValid()) {
+    // TODO(EvyBongers): figure this out. What to do when the settings are out of sync...
+
+    return;
+  }
```
EvyBongers, line 63: I don't really have any idea how to handle this case. I suppose I might change the condition to 'if(isValid()) { /* get overlays from storage */ }', but that doesn't help me with informing the user that there's corruption in EEPROM

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -84,6 +109,60 @@ EventHandlerResult ColormapOverlay::beforeSyncingLeds() {
...
+    // -> No overload for signed integers
+    // Ref: src/kaleidoscope/device/Base.h:90-92
+    // -> parseInt() seems to support signed values?
+    ::Focus.read(color_index_);
+    if (color_index_ >= 0) {
+      uint8_t key_index_ = i % Runtime.device().numKeys();
+      uint8_t layer_     = (i - key_index_) / Runtime.device().numKeys();
+
+      overlays_[overlay_count_] = Overlay(layer_, KeyAddr(key_index_), color_index_);
+      overlay_count_++;  // TODO(EvyBongers): how to make sure that we don't exceed max_overlays_
```
EvyBongers, line 156: Is there any way to communicate over focus that an error occurred? I could just `break` the loop when the limit is reached, but I wonder if there are ways to report back why it happens. Of course, if the limit above were to be set to `max_layers * numKeys` this wouldn't happen, but that doesn't make any sense because if you're gonna set led colors on every key you should be using Colormap instead of Colormap-Overlay

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -58,7 +59,7 @@ class ColormapOverlay : public kaleidoscope::Plugin {
   // takes as its sole argument an array reference of size `_overlay_count`, so
   // there's no need to use `sizeof` to calculate the correct size, and pass it
   // as a separate parameter.
-  template<uint8_t _overlay_count>
+  template<uint8_t _overlay_count>  // TODO(EvyBongers): how to make sure that we don't exceed max_overlays_
```
EvyBongers, line 62: Similar question as above, though this is when building a sketch. Should I validate this to the maximum and throw an error?

EvyBongers (commented):
Functionally it now works in that the numpad keys in the default firmware always light up, regardless of the assigned keys and that the overlay colors can be changed/disabled over focus.

One thing I noticed is that, if an effect is active on layer 0, that effect keeps playing on all keys not in the numpad. To work around this, I could change the default sketch to apply color index 22 (for example, 16-22 should be black once users upgrade), so that this doesn't happen.

However, that isn't the only way the sketch now differs from the Numpad plugin. My tests showed that, when changing from the numpad back to layer 0, the overlay colors persisted until the point that the active color mode applied a different color. That is:
* With any mode that lights all leds, the leds follow the color mode logic and the colors applied by Colormap-Overlay don't persist.
* with modes like stalker of chase, the red color persists until the effect hits those keys and changes the leds
* with LEDOff the color of the overlay persists indefinitely

Please let me know what you think about this, as well as the comments I left on my TODOs

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
obra: there isn't currently.  There *should* be.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
obra: I think that when eeprom settings isn't valid, it probably makes sense to wipe the config.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
obra: do we want to be setting it to layer_count? that might (or might not) need an accessor.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
obra: could you static assert so it throws a build error?

obra:
> Functionally it now works in that the numpad keys in the default firmware always light up, regardless of the assigned keys and that the overlay colors can be changed/disabled over focus.
> 
> One thing I noticed is that, if an effect is active on layer 0, that effect keeps playing on all keys not in the numpad. To work around this, I could change the default sketch to apply color index 22 (for example, 16-22 should be black once users upgrade), so that this doesn't happen.

I don't really feel strongly here.

> 
> However, that isn't the only way the sketch now differs from the Numpad plugin. My tests showed that, when changing from the numpad back to layer 0, the overlay colors persisted until the point that the active color mode applied a different color. That is:

This one, you should be able to fix by forcing the active led mode to do a full repaint, no?

> 
> * With any mode that lights all leds, the leds follow the color mode logic and the colors applied by Colormap-Overlay don't persist.
> * with modes like stalker of chase, the red color persists until the effect hits those keys and changes the leds
> * with LEDOff the color of the overlay persists indefinitely
> 
> Please let me know what you think about this, as well as the comments I left on my TODOs

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
EvyBongers: `layer_count` is far too few, as an overlay only applies to a single key. Of course we could set it to `layer_count * Runtime.device().numKeys()`, but if anyone is gonna need that number of overlays, they're better off using Colormap rather than ColormapOverlay.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -82,6 +101,69 @@ EventHandlerResult ColormapOverlay::beforeSyncingLeds() {
...
+      uint8_t key_index_ = i % Runtime.device().numKeys();
+      uint8_t layer_     = (i - key_index_) / Runtime.device().numKeys();
+
+      overlays_[overlay_count_] = Overlay(layer_, KeyAddr(key_index_), color_index_);
+      overlay_count_++;
+
+      if (overlay_count_ >= max_overlays_) {
+        // TODO(anyone): For now we just break the loop. Once focus implements
+        // support for reporting back errors, this should report why we break
+        // the loop
```
EvyBongers, line 153: Left a TODO to follow up on this discussion: https://github.com/keyboardio/Kaleidoscope/pull/1434#discussion_r1991593501

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -127,8 +131,10 @@ class ColormapOverlay : public kaleidoscope::Plugin {
   }
 
  private:
-  Overlay *overlays_;
-  uint8_t overlay_count_;
+  static uint16_t overlays_base_;
+  static const uint8_t max_overlays_ = 64;  // TODO(EvyBongers): figure this out. How determine a good maximum?
```
EvyBongers, line 135: Following up on https://github.com/keyboardio/Kaleidoscope/pull/1434#discussion_r1991570807:

`layer_count` is far too few, as an overlay only applies to a single key. We could set it to `layer_count * Runtime.device().numKeys()`, but if anyone is gonna need that number of overlays, they're better off using Colormap rather than ColormapOverlay.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
```diff
@@ -79,12 +81,12 @@ class ColormapOverlay : public kaleidoscope::Plugin {
...
     uint8_t count = 0;
-    for (int layer_ = 0; layer_ < _layer_count; layer_++) {
-      for (int key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
+    for (uint8_t layer_ = 0; layer_ < _layer_count; layer_++) {
+      for (uint8_t key_index_ = 0; key_index_ < kaleidoscope_internal::device.matrix_rows * kaleidoscope_internal::device.matrix_columns; key_index_++) {
         int8_t color_index_ = overlays[layer_][key_index_];
         if (color_index_ >= 0 && color_index_ < ::LEDPaletteTheme.getPaletteSize() &&
             color_index_ != no_color_overlay) {
-          count++;
+          count++;  // TODO(EvyBongers): how to make sure that we don't exceed max_overlays_
```
EvyBongers, line 89: Per your suggestion, I added a `static_assert` in the method above, but since I need to count the number if overlays here, it seems impossible to use that here. Any ideas on how I could validate that here? Or should I move the logic for finding all specified overlays to the macro below?

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
```diff
@@ -30,20 +30,39 @@
...
 void ColormapOverlay::setup() {
   ::LEDPaletteTheme.reservePalette();
+
+  overlays_base_ = ::EEPROMSettings.requestSlice(max_overlays_ * sizeof(Overlay));
+
+  ::EEPROMSettings.seal();
+
+  if (!::EEPROMSettings.isValid()) {
+    // When settings are invalid, better to clean it all up to prevent unwanted things from happening
+    ::EEPROMSettings.invalidate();
```
EvyBongers, line 53: Re: https://github.com/keyboardio/Kaleidoscope/pull/1434#discussion_r2038431488

Is this the correct way to handle this?

EvyBongers:
> This one, you should be able to fix by forcing the active led mode to do a full repaint, no?

Just implemented this and verified that it works.

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.cpp
obra: I don't believe so.  the seal is taken care of automatically. and isValid is a whole-system level thing.
This should be the right pattern
```
EventHandlerResult Keyclick::onSetup() {
  // Request storage space and load settings
  bool success = ::EEPROMSettings.requestSliceAndLoadData(&settings_base_, &settings_);

  // If settings weren't loaded successfully, use defaults
  if (!success) {
    settings_.enabled = true;
    settings_.version = 0x01;
    updateSettings();
  }
```

plugins/Kaleidoscope-Colormap-Overlay/src/kaleidoscope/plugin/Colormap-Overlay.h
obra: Understood. I think that starting at a fixed # like 64 seems fine. if we find folks needing it increased, we can do that.

PR #1477, opened 2025-02-21
Fix stalker docs link in Model100

TurekBot:
I noticed that the previous link was broken

image

so I tracked down the [current link](https://github.com/keyboardio/Kaleidoscope/blob/master/plugins/Kaleidoscope-LED-Stalker/README.md) and figured I'd pass it on!

I also did a search and didn't see any other instances of this link, or others like it, in this repo.


obra:
thanks so much!

PR #1498, opened 2025-04-13
Include examples in plugins in smoke_sketches

EvyBongers:
Just noticed that not all examples are built with the smoke sketches CI :)


EvyBongers:
Fixed most failed builds already. KeyboardioHID examples are still failing on a missing import and Kaleidoscope-Hardware-Keyboardio-Alto fails because the board is unknown.
Assuming you want to include these missing exambles in CI, could you look into fixing those errors?

PR #1515, opened 2025-07-22
Piezo Speaker Tune/Melody Functionality

jackhumbert:
Makes some methods virtual for overriding, and allows initializer lists for nice calls like this:

```c++
kaleidoscope::Runtime.device().speaker().playTune(
    {440, 880,  440,  880}, 
    {50,  50,   200,  50}
);
```


jackhumbert:
Ah, I didn't see that! I should have dug deeper. It's funny how we ended up with similar `playNextNote` methods :)

This is great, though - I love the phonemes! I'll check it out and play around with it.

PR #1517, opened 2025-07-28
Fix udev ENV assignment operator syntax

ohno418:
Changed ENV{} assignments from ':=' to '=' operator in udev rules to resolve systemd-udevd warnings. The ':=' operator is not valid for ENV keys in udev rules - they only accept '==', '\!=', '=', or '+=' operators.

This fixes the following warning:

systemd-udevd[474]: /etc/udev/rules.d/60-kaleidoscope.rules:13 ENV key takes '==', '!=', '=', or '+=' operator, assuming '='.


obra:
Thanks for this. It appears that this is caused by systemd's udev parser not actually being compatible with udevd.

PR #1518, opened 2025-08-06
Fix Preonic layout rendering for bluetooth

gglanzani:
Some leftover from an old firmware?


obra:
Indeed. That was an internal tool that was useful enough to keep around and share. But it did need that update. Thank you!

PR #1539, opened 2026-02-07
Add support for ignoring Qukeys activation on the same hand

### LINKED ISSUE #1407: Qukeys support for same-hand/location-specific inactivation

HRMs (home row mods) are a fairly common feature of layouts in the ergo keyboard community, especially for users who need to minimize lateral thumb movement.  With the current implementation of Qukeys, they are prone to frequent misfires with quick rolls of e.g. adjacent home row keys (which are a common feature of many alpha layouts).

Both QMK (via a userspace solution, [achordion](https://getreuer.info/posts/keyboards/achordion/index.html) being the best known) and ZMK (via [hold-key-trigger-positions](https://zmk.dev/docs/behaviors/hold-tap#advanced-configuration)) allow the user to customize, for any given qukey, a set of keys that it could possibly act as the hold value for versus always acting as a tap.  Most commonly, this is used to make HRMs only apply cross-hand (i.e. a right hand qukey always resolves as tap for other right hand keys), but it can also be used for more detailed configuration (I personally disable my dual-function keys for same-row presses on the same hand only). 

There isn't much alternative currently feasible in Kaleidoscope, other than learning to type "cleaner", but this necessarily becomes a problem at high speeds.

**Configuration thoughts**

At its simplest, this could simply be a flag to disable same-hand interaction; that would get a large portion of the functionality of this feature for minimum effort.

However, as mentioned above, sometimes more granular control is desirable, in which case, there are two possible approaches I can think of:

* When defining a qukey, one can optionally pass a list of secondary key coordinates for which it is active.  IMO, coordinates make the most sense here, rather than key codes since you generally care about the physical relation of the keys.  This is how ZMK does it and feels more "kaleidoscope-y" to me.
* Have a function that is passed the details of the qukey and secondary keys for each interaction, the return value of which determines whether the qukey can be active or not.  This is how achordion works for QMK, but to me doesn't feel like a "kaleidoscope-y" solution, since it tends to prefer customizing plugin behavior rather than hacking on a `process_record` and the like.

eirvandelden:
This fixes #1407 

Disclaimer: I'm a web developer and this is too low level for me. This is the code Claude Opus generated for me to be able to use Qukeys, but ignore activations from the same hand. I.e. having homerow mods, it will ignore a held key when activating with another key on the same hand (output 2 different keys), but if I hold one key and then press another key on the other cluster, it does trigger the alt mode.

### DIFF
--- a/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.cpp
+++ b/plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.cpp
@@ -194,6 +194,15 @@ bool Qukeys::processQueue() {
       // Otherwise, we've found a subsequent key press, so we record it for the
       // overlap comparison later, unless we've already done so.
       if (next_keypress_index == 0) {
+        // If the opposite-hands rule is enabled and this subsequent key is on
+        // the same hand as the qukey, immediately flush as primary. There's no
+        // need to wait for the overlap comparison because same-hand rollover
+        // is never treated as a modifier chord.
+        if (require_opposite_hands_ &&
+            !oppositeHands(queue_head_addr, event_queue_.addr(i))) {
+          flushEvent(queue_head_.primary_key);
+          return true;
+        }
         next_keypress_index = i;
       }
       continue;
@@ -216,6 +225,14 @@ bool Qukeys::processQueue() {
         tap_repeat_.addr       = queue_head_addr;
         tap_repeat_.start_time = event_queue_.timestamp(0);
         flushEvent(event_key);
+        return true;
+      }
+      // If the opposite-hands rule is enabled, and the subsequent key is on
+      // the same hand as the qukey, skip the overlap delay and flush as
+      // primary immediately. Same-hand rollover is always treated as typing.
+      if (require_opposite_hands_ &&
+          !oppositeHands(queue_head_addr, event_queue_.addr(next_keypress_index))) {
+        flushEvent(queue_head_.primary_key);
         return true;
       }
       // Now we know the qukey has been released, but we need to check to see if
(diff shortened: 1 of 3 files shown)


plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.cpp
```diff
@@ -194,6 +194,15 @@ bool Qukeys::processQueue() {
...
       if (next_keypress_index == 0) {
+        // If the opposite-hands rule is enabled and this subsequent key is on
+        // the same hand as the qukey, immediately flush as primary. There's no
+        // need to wait for the overlap comparison because same-hand rollover
+        // is never treated as a modifier chord.
+        if (require_opposite_hands_ &&
+            !oppositeHands(queue_head_addr, event_queue_.addr(i))) {
+          flushEvent(queue_head_.primary_key);
+          return true;
+        }
```
EvyBongers, line 205: It appears to me that this clause should be outside of the parent condition, similar to the spacecadet check at L190

plugins/Kaleidoscope-Qukeys/src/kaleidoscope/plugin/Qukeys.h
```diff
@@ -210,13 +229,21 @@ class Qukeys : public kaleidoscope::Plugin {
     Key alternate_key{Key_Transparent};
   } queue_head_;
 
+  // When true, qukeys will only resolve to their alternate value when the
+  // subsequent keypress is on the opposite hand.
+  bool require_opposite_hands_{false};
+
+  // The column index that divides left and right hands.
+  uint8_t split_column_{8};
```
EvyBongers, line 237: I wonder if it would be possible (and maybe better) to use the driver data instead of hard coding a number that just so happens to work for two of Keyboardio's keyboards. I _think_ this should work:

```suggestion
  uint8_t split_column_ = kaleidoscope::Device::Props::KeyScannerProps::matrix_columns / 2;
```

eirvandelden: I'll try it out and let you know!

eirvandelden: I'll try it out and let you know!

eirvandelden:
@EvyBongers one of your 2 comments is not doing what we thought it would, I tried them and could activate alt modes again on the  same cluster. I will experiment with each comment separately

PR #1540, opened 2026-04-17
Fix ble reconnection after sleep

zachgersh:
So far, so good after running this on my keyboard for the last few days here. All three of these changes work together and I've got descriptive commit messages on each describing why / what.


zachgersh:
Hey @obra can you have a look at this? I've been running this on my keyboard now for a week and its been way better bluetooth connection wise. I think this would probably solve a lot of complaints.

zachgersh:
@obra bump again. I really think this should be merged. Works way better than the base firmware for me as far as reconnection is concerned.

karptonite:
I just want to chime in: I installed this on my Preonic, and it has totally fixed the issue for me. No problem with battery life that I can see.

